PR #9, opened 2020-03-01
Publish photoview as a single docker image.

subdavis:
While I recognize that distributing docker images that contain both compiled client code and a server is somewhat antithetical to docker, it's become more or less the standard, particularly within the selfhosted community.  Plex  Photoprism, Minio, Pihole, Nextcloud, the list goes on.  They all distribute a single docker image.

I think Photoview, for the sake of good user experience AND flexibility, should do the same.  The requirement that users use your docker compose config or else "just figure it out" with their own container orchestration will be, IMO, a barrier to entry and a nuisance to a great deal of prospective users.

I've opened this as a draft PR because I the docs haven't been fixed and there are some rough spots.   In particular, the catchall router that returns the index html page is bullshit, and I'd like to change that router to be, lexically, "attempt serve from static assets anything that doesn't start with /api, and fallback to index.html".

I'm also not terribly familiar with service workers, but I didn't see any issues with how this is being done.  Can always force some cache headers into the response.

I think this is an awesome project, and I totally respect what you tried to do with separating the images, but a go server and 2 nginx servers is very much overkill for the sort of application being developed here.

If you don't like the catchall, we could also serve the UI from `/ui` and do a simple redirect for `/`, though I think React's HTML5 history router isn't quite as robust as would make that a good experience for developers.  I actually tried to do that first and, being a Vue.js developer, became frustrated and gave up.  Parcel is also not my jam, but I digress.

Hope we can have a conversation about this change, because I think it'd add a lot of value to the project.


subdavis:
Also note that, yes, I did just blindly copied gorilla/mux's SPA router :)

viktorstrate:
First of all, thank you for taking the time to make the pull request and writing the detailed description.

I think it's a good idea to combine the services into one docker container.
I agree that making it as easy as possible to set up the project is very important.
It would also make it easier to publish the image to Docker Hub.

> In particular, the catchall router that returns the index html page is bullshit, and I'd like to change that router to be, lexically, "attempt serve from static assets anything that doesn't start with /api, and fallback to index.html".

Doesn't your code check whether a file exists and serves that and falls back to the index.html file like it should? - What am I missing

I think the "catchall method" it the way to go, and properly also what is the closest to how it currently works. That is, if the route starts with `/api` it goes to the api router, and everything else is handled by the spa router.

I think that the spa router maybe should be moved to its own file under `/api/routes/ui.go`, to keep `server.go` as simple as possible.

The environment variables `API_ENDPOINT` and `PUBLIC_ENDPOINT` could probably also be combined into one.

I will have to try it out on my desktop computer, with docker installed, later.

subdavis:
> I think that the spa router maybe should be moved to its own file under /api/routes/ui.go , to keep server.go as simple as possible.

Agreed.  I'll do this.

> Doesn't your code check whether a file exists and serves that and falls back to the index.html file like it should? - What am I missing

The thing this code does that I think is unreasonable is instead of a nice 404 on endpoints that aren't real, like /api/foo, this will just return index.html

That needs to be fixed, IMO.  I'll make another attempt.

viktorstrate:
I see, maybe the `/api` endpoint could maybe be turned into a sub-router, and if nothing matches after `/api` it would send a 404.
Something like:

```go
apiRouter := endpointRouter.PathPrefix("/api").Subrouter()
apiRouter.Handle("/graphql", ...)
photoRouter := apiRouter.PathPrefix("/photo").Subrouter()

// If none of the previous handlers match, return 404
apiRouter.Handle("/", NotFoundHandler)
```

I think that "not found" pages on the ui, should be handled by the front-end react-router.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# Build API
+FROM golang:alpine AS api
+
+WORKDIR /app
+COPY api /app
+
+RUN go get -d -v ./...
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o photoview .
+
+# Copy built app to nginx environment
```
viktorstrate: This comment should probably be changed to `production environment` or somthing like that

docker-compose.example.yml
```diff
@@ -25,7 +25,7 @@ services:
       # Change This: The publicly exposed url for the api
       # For example if the server is available from the domain example.com,
       # change this value to http://example.com/api
-      - API_ENDPOINT=http://localhost:8080/api
+      - API_ENDPOINT=http://localhost:8080/
       - PUBLIC_ENDPOINT=http://localhost:8080/
```
viktorstrate: Theese should be combined into one called maybe `ENDPOINT_URL`

docker-compose.example.yml
```diff
@@ -25,7 +25,7 @@ services:
...
       - API_LISTEN_PORT=80
```
viktorstrate: This could be removed as the standalone API has been removed

docker-compose.example.yml
```diff
@@ -37,32 +37,6 @@ services:
...
       - api_cache:/app/cache
 
-  ui:
-    build:
-      context: ./ui
-      args:
-        # Change This: The publicly exposed url for the graphql api
-        # For example if the server is available from the domain example.com,
-        # change this value to http://example.com/api/graphql
-        GRAPHQL_ENDPOINT: http://localhost:8080/api/graphql
```
viktorstrate: This variable is used when building the ui to know where to connect to the backend see `apolloClient.js` below.

https://github.com/viktorstrate/photoview/blob/e0ada51b6ba0a787ced22d309a100e6848c3641b/ui/src/apolloClient.js#L16-L21

This could probably be easily fixed, as the front-end and back-end are served on the same domain

docker-compose.example.yml
subdavis: I don't think it has.  These changes affect the production docker image, but the "standalone api" will still be used during development of the frontend when using the parcel dev server.

viktorstrate: I see, we can't remove that. We could make it default to port 80 if no environment variable is given, but that should be a pull request for itself.

subdavis:
I'm still planning to do this.  At a conference, so response times may be delayed.  I'll rebase if necessary

viktorstrate:
@subdavis are you still planning on finnishing this, or should I take a crack at it?
I think this feature would really help the project.

docker-compose.example.yml
viktorstrate: This has been resolved in 48520a7

viktorstrate:
I have just installed this branch on the demo server and everything worked as expected.

PR #17, opened 2020-03-13
Sql 4 bytes utf

viktorstrate:
Migrate SQL to extended unicode, as a follow-up to a65a7bc2890f43eb8640ca183b2269fcfa02971e

@inDream, would this solve it?


inDream:
Yes, thank you.

PR #27, opened 2020-04-14
fix default port number

SlavikCA:
docker-compose has port 8000, but README had 8080. Setting README to 8000, so it match docker-compose


viktorstrate:
Thank you, I missed that after rewriting the Docker file.

PR #28, opened 2020-04-14
Database connection not closed if issue with migration

BkSouX:
Defer is not executed with a fatal, the app crashes directly.
It is better to use a panic, defer is called before exiting the app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ yarn-error.log*
 
 # vscode
 __debug_bin
+.idea
--- a/api/database/mysql.go
+++ b/api/database/mysql.go
@@ -1,11 +1,14 @@
 package database
 
 import (
+	"context"
 	"database/sql"
 	"log"
 	"net/url"
 	"os"
 	"time"
+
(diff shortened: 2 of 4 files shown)


viktorstrate:
Would it be worse to just manually close the database connection before calling log.Fatal instead?
I think that would be simpler and easier to understand.

Maybe we should handle the case where the [http.ListenAndServe](https://github.com/viktorstrate/photoview/blob/11ea1c13bd0b74953fac55b79ac5a3ea403322a4/api/server.go#L99) fails as well?

viktorstrate:
There is potentially one more place where the database would not be closed properly, and that is if a connection was successful, but the ping failed.

https://github.com/viktorstrate/photoview/blob/master/api/database/mysql.go#L69

viktorstrate:
Awesome I will take a look later 👍

viktorstrate:
The code looks good. I've added some small cleanup changes i found while reading the code.
If it's ok with you, I will merge.

BkSouX:
OK :+1:

PR #53, opened 2020-06-24
Fixes viktorstrate/photoview#23

stz184:
InnoDB has a maximum index length of 767 bytes for tables that use COMPACT or REDUNDANT row format, so for utf8mb3 or utf8mb4 columns, you can index a maximum of 255 or 191 characters, respectively.

That's why I created a new column storing a MD5 hash of the path and made it unique. The MD5 hash has only 32 characters and can be generated natively in MySQL and MariaDB. It helps us to avoid maximum key length and in the same time enforce unique photo and album paths.

- Added path_hash column to photo and album tables
- Added down migration file for 005_utf8_migration (just for consistency)
- Added PathHash field to Album and Photo struct
- album_scanner.go and photo_scanner.goo perform checks by MD5 hash now


api/database/migrations/002_photo.up.sql
```diff
@@ -20,7 +20,8 @@ CREATE TABLE IF NOT EXISTS album (
   title varchar(256) NOT NULL,
   parent_album int,
   owner_id int NOT NULL,
-  path varchar(512) NOT NULL UNIQUE,
```
viktorstrate, line 23: New changes to the database should be moved to a new migration file to make updating possible without deleting the database.
Somthing like `007_path_hashes.up.sql`.

I don't know if there is a way to alter the table to add the `path_hash` attribute and automatically generate all the hashes on migration? Maybe somthing like this, though I haven't tested it:

```sql
-- Remove unique
ALTER TABLE photo MODIFY path varchar(512) NOT NULL;

-- Add path_hash and set it to the md5 hash of the path attribute
ALTER TABLE photo ADD path_hash varchar(32) UNIQUE;
UPDATE photo p SET path_hash = md5(p.path);
ALTER TABLE photo MODIFY path_hash varchar(32) NOT NULL UNIQUE;
```

Maybe you know of a better way to do this?

stz184: > Thank you so much @stz184 for this contribution. I probably wouldn't have figured this out otherwise.
> 
> Apart from the merge conflict, it is only the database migration that needs to be solved - let me know if you want me to take a look at this.

I just fixed the conflicts. I was working on the old codebase, it was my bed. 
Anyway.
I don't really understand why do you think that only database migration is needed. 
Something needs to update the hashes every time when a photo or album is added and use it for comparison when check for existence is made. What's the point to have a hash column if you continue to do your checks for existing photos/albums using the path column which is not indexed instead of path_hash which has a unique key?

About the migration itself - I also thought that it should be created as new file and applied without deleting the entire database. Sadly, this is not possible. Just imagine the issue that me @BkSouX has experienced - we failed to set up the database because the migration failed on v5 - 005_utf8_migration.up.sql with an error

    details: Error 1071: Specified key was too long; max key length is 3072 bytes

So, in this case the migration will never reach and apply version 006 and 007. 
If you have any ideas on how I can alter the tables album and photos even if 005_utf8_migration.up.sql failed and reapply the UTF8 to UTF8MB4 migration again I will refactor the PR.

stz184: P.S. I thought I can edit the 002_photo.up.sql which will be applied before 005_utf8_migration.up.sql when a new project is set up and 007 migration file that will alter the photo and album table to add "path_hash" column if does not exists for the existing projects that somehow succeded to migrate the tables to UTF8MB4 and have version 5 of the migration applied.
Sadly, there is no such as thing as "Add a column if it does not exists" in MySQL.

viktorstrate: I see, we have to make the changes in `002_photo.up.sql`.
Maybe a mysql IF statement could be used to check if the column exists?

> I don't really understand why do you think that only database migration is needed.
Something needs to update the hashes every time when a photo or album is added and use it for comparison when check for existence is made. What's the point to have a hash column if you continue to do your checks for existing photos/albums using the path column which is not indexed instead of path_hash which has a unique key?

I do think every that every comparison should use the indexed `path_hash` instead, I don't know where i hinted otherwise?
I really like your changes to the go code regarding inserting and querying the `path_hash`.
Please elaborate if I missed somthing :)

stz184: Sorry, I misunderstood you. I misunderstood "it is only the database migration that needs to be solved" as a hint that there is no need of go code change :)
About the IF statement - It seems to be possible to check for column existence only inside a procedure or trigger (check this out https://stackoverflow.com/questions/972922/add-column-to-mysql-table-if-it-does-not-exist/33755258)

Anyway, I find it very ugly solution to create a temporary procedure, call it and then drop it, but if you think it worths trying - I will apply it :)

viktorstrate: INFORMATION_SCHEMA could be used to determine if the column exists, it seems to be avaliable in all modern sql databases. (see this https://tableplus.com/blog/2018/08/ms-sql-server-how-to-check-and-add-new-column-if-it-does-not-exist.html)

Somthing like this:
```sql
IF NOT EXISTS (
  SELECT * FROM INFORMATION_SCHEMA.COLUMNS
  WHERE TABLE_NAME = 'photo' AND COLUMN_NAME = 'path_hash'
) BEGIN
  -- Do migration here
END;
```

stz184: The article you cited is written with Microsoft SQL server in mind. The example that the author has given will produce 

> Error in query (1064): Syntax error near 'IF NOT EXISTS

error in MySQL/MariaDB

viktorstrate: @stz184 Oh I thought that `IF NOT EXISTS` worked for mysql aswell, sorry for not testing it first.
I think that a temporary stored procedure might be the best solution then. I came up with this, which works quite well as it also generates the hashes automatically.

This has been tested with `0.4.11-MariaDB`

```sql
DELIMITER //
CREATE PROCEDURE MigratePathHashIfNeeded()
BEGIN

IF NOT EXISTS( SELECT *
            FROM INFORMATION_SCHEMA.COLUMNS
           WHERE table_name = 'photo'
			 AND table_schema = DATABASE()
             AND column_name = 'path_hash')  THEN
             
    -- Remove unique index from photo.path
    ALTER TABLE photo DROP INDEX path;

    -- Add path_hash and set it to the md5 hash based of the path attribute
    ALTER TABLE photo ADD path_hash varchar(32) UNIQUE;
    UPDATE photo p SET path_hash = md5(p.path);
    ALTER TABLE photo MODIFY path_hash varchar(32) NOT NULL UNIQUE;
	
END IF;

END //
DELIMITER ;

CALL MigratePathHashIfNeeded();
DROP PROCEDURE MigratePathHashIfNeeded;
```

stz184: Cool! I will add apply it and ping you back :)

stz184: @viktorstrate , it seems that I can't use DELIMITER statement outside cli.
MySQL keep saying Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DELIMITER // when I run the migration. 
Please, check this issue https://github.com/golang-migrate/migrate/issues/399

viktorstrate: @stz184 This issue is quite a bit more complicated that I first anticipated 😊 
As mentioned in the issue you linked, the delimiter is not needed if `multi-statement` is turned on (which it is)
> Just stick to a simple ; delimiter and add multiStatements=true to your DSN.

I've pushed some migration changes that works without custom delimiters.
I have tested it both by migrating from the current master branch to this branch and from an empty database, and it seems to work as expected. I might test a bit more before the final merge.

If you have any comments or suggestions to my edits please share.

stz184: Oh my god, I completely miss that part about multiStatements=true parameter in the DSN.
Thanks for pointing me this out.
I guess we are not ready to merge this PR?

viktorstrate: I will migrate the demo server, and merge if everything goes as expected.
Thank you so much for the help both the code changes and troubleshooting.

You shall be more than welcome to contribute again another time.

PR #76, opened 2020-09-07
Fixes viktorstrate/photoview#6

stz184:
I have added a "Show only the favorites" checkbox on both /pages and /album/:id/ pages

- On Photos page, when the toggle is checked only albums containing favorited media are displayed and only favorited media is lisred. The URL changes to /photos/favorites and it can be used to directly access the favorited media.
-  On the individual album's page, the toggle shows only favorited media when it is checked. The URL changes to /album/ID/favorites and it can be used to directly access the favorited media from this particular album.
    forceVisible() is called when the loading ends in PhotoGallery component because otherwise if you switch to Favorites and some of the favorited images have been outside the viewport before that, they will not be lazy-loaded once they are "moved" to the visible area because non-favorited images are cleared.

**Changes:**
- Added favorites toggle on both Album and Photos page
- When the favorites are viewed, the URL changes
- Removed unused variable activeImage
- Added match.params propTypes to AlbumPage component


viktorstrate:
I have only skimmed the code on github, but it looks good.
I have just tested it and I found some minor issues:

- If I favorite an image and activate the filter without reloading, the newly favorited image doesn't show up.
- Maybe use `replaceState` instead of `pushState` when updating the url. I think it is more natural to navigate back to previous page instead of changing a parameter on the current page when you click the back button in the browser.
- There's some problems on smaller screens. I think the best solution would be to move the checkbox down below the album title and left align it. And just have at the top, also left aligned, on the `/photos` page.

Otherwise it looks great. Thank you so much for your contributions. I don't have a lot of time to work on Photoview atm. because of semester start on uni, but I hope to do more work soon when everything setteles down a little 😊

stz184:
Thanks for the notes :)
I will apply the fixes as soon as I can. 
Thanks for reviewing it!

stz184:
Thanks for the notes :)
I will apply the fixes as soon as I can. 
Thanks for reviewing it!

stz184:
Apologies for the huge delay!
I think I applied all suggested changes + fixed some issues I found. I tried to simplify the logic as much as possible but I am open to hearing advice.

ui/src/Pages/PhotosPage/PhotosPage.js
```diff
@@ -43,13 +55,59 @@ class PhotosPage extends Component {
...
+    window.addEventListener('popstate', this.onPopState)
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('popstate', this.onPopState)
+  }
+
+  favoritesCheckboxClick(refetch) {
+    const onlyWithFavorites = !this.state.onlyWithFavorites
+    history.pushState(
```
viktorstrate, line 86: Should use `history.replaceState`

ui/src/Pages/AlbumPage/AlbumPage.js
```diff
@@ -38,22 +42,70 @@ const albumQuery = gql`
...
+        }
+        setOnlyFavorites(newState)
+      })
+    } else {
+      setOnlyFavorites(newState)
+    }
+    history.replaceState(
+      {},
+      '',
+      '/album/' + albumId + (onlyFavorites ? '/favorites' : '')
```
viktorstrate, line 74: When navigating to a `/album/:id` page, and click show only favorites, everything works as expected, except that the url doesn't update to `/album/:id/favorite`, but when you click it again to show all images again the url updates to `/album/:id/favorite`.

I could imagine it's because `onlyFavorites` on this line should be changed to `newState`

viktorstrate (changes requested):
Awesome! Would you mind merging with the latest changes from master?
It's much better, but I still found some small issues, I've made some comments in the code related to the issues I found when testing it.

ui/src/Pages/AlbumPage/AlbumPage.js
stz184: Yeah, I forgot this. I did a refactoring and somehow miss that.

ui/src/Pages/PhotosPage/PhotosPage.js
stz184: Oh, I forgot this. You told me to change it to replaceState and I did it on Album's page but forgot this one.

stz184:
Thanks for the notes, everything is fixed now.

viktorstrate (approved):

viktorstrate:
Looks great thank you!

PR #99, opened 2020-10-15
Sorting options

stz184:
This PR adds a dropdown for media sorting.
- FormatSQL() method in the backend API is refactored because the old approach of validing the input using regex was not perfect:
   - completely ignored album.title sorting (because of the dot sign) 
   - it was not possible to pass complex sorting like `media.media_type, SUBSTRING_INDEX(media.path, '.', -1)`
   - it revealed information about the database structure which may be security problem.
- Added new component `GalleryGroups` in order to prevent some unecessary rerenderings of the whole layout
- Refactored the AlbumTitle component and extracted the favorites toggle logic and the sorting dropdown in a new component called `AlbumFilter`
- Refactored `AlbumPage` in order to prevent unnecessary rerenderings of the `AlbumFilter` and `Layout` components


stz184:
I just updated the PR with few minor changes:
- Added URL parameters for order column and order direction
- forceCheck() is called when the loading is complete in order to display photos that may have been outside the viewport but are visible now because of changed ordering
- Moved the label of Favorites toggle to the left for consistency

PR #112, opened 2020-10-30
Migrate all class based react components to function based

stz184:
Migrating components to functional ones.


viktorstrate:
Thank you, looks good!

The tests in `Layout.test.js` just needs to be updated, you can take a look at my changes to the way the `.authToken()` function is being mocked here for reference.
https://github.com/viktorstrate/photoview/commit/51556437f63dfae51db1defc45b47dc7105f5f20#diff-38d8262343cc2dc682ee8d2418a875a9fc049961588933cd1afb121d6a5e3234

Also the static analyzer found a waning in [AlbumPage.js:99](https://github.com/viktorstrate/photoview/pull/112/files#diff-4c41e8d4e646aed1ff67d032a577f4c722b0ac8d78ea735dc61e838ea76e7bcdR99),
that the variable refetch is used before its declaration, unless there is a good reason for this, it should be moved below the declaration of refetch.

I wouldn't mind making the last small changes if you want that.

stz184:
I will make the changes, I just as too busy these days.
If it is blocker for you, just let me know.

viktorstrate:
Sure thing, take your time :)

stz184:
I just fixed the test and the issue with the `refetech` variable.
I saw that you have refactored the tests so you don't overwrite the signature of the functions you want to mock but you are using jest to mock it (as it should be). That's really nice, thanks!
But this does not prevent us from accidental use of mocked instead of original functions in subsequent tests. Pease, consider using `afterEach` hook and `clearAllMocks/resetAllMocks` methods. 
Good article about that is https://codewithhugo.com/jest-stub-mock-spy-set-clear/

viktorstrate:
Looks good. Yes I will update the Layout tests to clear the mocks for each test.
Thank you for linking the article.

ui/src/components/albumGallery/AlbumGallery.js
```diff
@@ -56,7 +56,9 @@ const AlbumGallery = ({
 
   useEffect(() => {
     const updateImageState = event => {
-      setImageState(event.state.imageState)
+      if (event.state.imageState) {
```
viktorstrate, line 59: Is there a reason you put this check here?
It seems to break fullscreen history functionallity.

When I visit an album, click full screen on an image and the press the back button, it doesn't go out of full screen, it does though if I comment this change out.

viktorstrate (changes requested):
@stz184 I took a proper look and I've collected a list of files that still contain class based react components.

- [x] ui/src/Pages/Settings/UserRow.js (Mutation in ChangePasswordModal)
- [ ] ui/src/Pages/LoginPage/InitialSetupPage.js
- [x] ui/src/App.js
- [x] ui/src/components/photoGallery/MediaThumbnail.js (LazyPhoto)
- [ ] ui/src/components/routes/Routes.js
- [ ] ui/src/components/sidebar/Sidebar.js

I don't have time to work on Photoview myself atm. so please don't feel rushed.

ui/src/components/albumGallery/AlbumGallery.js
viktorstrate: Solved in later commit

viktorstrate:
Looks great! I'll take a proper look in the coming days.

viktorstrate:
Thank you! I found no problems related to to your changes when testing it it.
I did, however find a new bug #133 when setting a password for a share, but it isn't related to these changes.

PR #135, opened 2020-11-20
Update JPEGS on sidecar change

seam345:
# Changes
HighRes and thumbnail JPEGs now update when the RAW sidecar (.xmp) file changes

# How
Darktable CLI already uses the sidecar file if it can find it, however photoview has no idea if the sidecar file has changed so I hashed the file and stored it in the DB to detect changes, if a change is detected I delete the old highres and thumbnail and re-encode them. Due to Darktable already using the xmp file if it found it I have left the call to Darktable unchanged as explicitly passing Darktable the .xmp file causes Darktable to fail if there is any problem with the .xmp file compared to now where it falls back to just encoding the picture as it is)

# Bug
After making the changes I realised getting the size for my raw files (NEF) is incorrect, however they would be incorrect anyway if I had a crop in the .xmp file as the previous method encoded the high res JPEG and then read the dimensions from the new JPEG. I changed it to read the dimensions directly from the raw file but I think it is reading the dimensions of the embedded JPEG thumbnail (as the sizes reported are 160x120 instead of (4282x2844). I plan to resolve this but wanted to get a sense if the JPEG update feature was something we wanted in the first place.

I have very little experience in Go if you would like anything changed please let me know I am happy to make many changes.
We most likely want to store the hash/path in a separate table or the hashSideCarFile function may want to live somewhere else, happy to move them to wherever is recommended.

### DIFF
--- a/api/database/migrations/003_sar_side_car_files.down.sql
+++ b/api/database/migrations/003_sar_side_car_files.down.sql
@@ -1,0 +1,4 @@
+
+ALTER TABLE media
+  DROP COLUMN IF EXISTS side_car_path,
+  DROP COLUMN IF EXISTS side_car_hash;
--- a/api/database/migrations/003_sar_side_car_files.up.sql
+++ b/api/database/migrations/003_sar_side_car_files.up.sql
@@ -1,0 +1,4 @@
+
+ALTER TABLE media
+  ADD COLUMN IF NOT EXISTS side_car_path varchar(1024) DEFAULT NULL,
+  ADD COLUMN IF NOT EXISTS side_car_hash varchar(32) DEFAULT NULL;
--- a/api/scanner/process_photo.go
+++ b/api/scanner/process_photo.go
@@ -319,7 +319,7 @@ func processRawSideCar(tx *sql.Tx, imageData *EncodeMediaData, highResURL *model
 	currentSideCarPath := scanForSideCarFile(photo.Path)
 
 	if currentSideCarPath != nil {
-		currentFileHash := hashSideCarFile(currentSideCarPath)
+		currentFileHash = hashSideCarFile(currentSideCarPath)
 		if photo.SideCarHash == nil || *photo.SideCarHash != *currentFileHash {
 			sideCarFileHasChanged = true
 		}
@@ -354,7 +354,7 @@ func processRawSideCar(tx *sql.Tx, imageData *EncodeMediaData, highResURL *model
 		// save new side car hash
 		_, err = tx.Exec("UPDATE media SET side_car_hash = ?, side_car_path = ? WHERE media_id = ?", currentFileHash, currentSideCarPath, photo.MediaID)
 		if err != nil {
-			return errors.Wrapf(err, "could not update side car hash (%d, %s)", photo.MediaID, *currentFileHash)
+			return errors.Wrapf(err, "could not update side car hash for media: %s", photo.Path)
 		}
 	}
 	return nil
--- a/api/scanner/process_photo.go
+++ b/api/scanner/process_photo.go
@@ -96,10 +96,22 @@ func processPhoto(tx *sql.Tx, imageData *EncodeMediaData, photoCachePath *string
 		return false, errors.Wrap(err, "error processing photo highres")
 	}
 
-	// Generate high res jpeg
 	var photoDimensions *PhotoDimensions
 	var baseImagePath string = photo.Path
 
+	mediaType, err := getMediaType(photo.Path)
+	if err != nil {
+		return false, errors.Wrap(err, "could determine if media was photo or video")
+	}
+
+	if mediaType.isRaw() {
+		err = processRawSideCar(tx, imageData, highResURL, thumbURL, photoCachePath)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	// Generate high res jpeg
 	if highResURL == nil {
 
 		contentType, err := imageData.ContentType()
@@ -116,25 +128,9 @@ func processPhoto(tx *sql.Tx, imageData *EncodeMediaData, photoCachePath *string
 
 			baseImagePath = path.Join(*photoCachePath, highres_name)
 
-			err = imageData.EncodeHighRes(tx, baseImagePath)
-			if err != nil {
-				return false, errors.Wrap(err, "creating high-res cached image")
-			}
-
-			photoDimensions, err = GetPhotoDimensions(baseImagePath)
+			err = generateSaveHighResJPEG(tx, photo.MediaID, imageData, highres_name, baseImagePath, -1)
 			if err != nil {
 				return false, err
-			}
-
-			fileStats, err := os.Stat(baseImagePath)
-			if err != nil {
-				return false, errors.Wrap(err, "reading file stats of highres photo")
-			}
-
-			_, err = tx.Exec("INSERT INTO media_url (media_id, media_name, width, height, purpose, content_type, file_size) VALUES (?, ?, ?, ?, ?, ?, ?)",
-				photo.MediaID, highres_name, photoDimensions.Width, photoDimensions.Height, models.PhotoHighRes, "image/jpeg", fileStats.Size())
-			if err != nil {
-				return false, errors.Wrapf(err, "could not insert highres media url (%d, %s)", photo.MediaID, photo.Title)
 			}
 		}
 	} else {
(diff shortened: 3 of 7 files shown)


api/database/migrations/003_sar_side_car_files.down.sql
```diff
@@ -0,0 +1,4 @@
+
+ALTER TABLE media
+  DROP COLUMN IF EXISTS side_car_path,
+  DrOP COLUMN IF EXISTS side_car_hash;
```
viktorstrate, line 4: Small typo in DROP

api/database/migrations/003_sar_side_car_files.up.sql
```diff
@@ -0,0 +1,4 @@
+
+ALTER TABLE media
+  ADD COLUMN IF NOT EXISTS side_car_path varchar(1024) NOT NULL DEFAULT '',
+  ADD COLUMN IF NOT EXISTS side_car_hash varchar(32) NOT NULL DEFAULT '';
```
viktorstrate, line 4: Maybe we should use NULL here to indicate the sidecar file doesn't exist?

api/scanner/scanner_media.go
```diff
@@ -10,6 +13,30 @@ import (
...
 )
 
+func scanForSideCarFile(path string) (string, error) {
+	testPath := path + ".xmp"
+	_, err := os.Stat(testPath)
+	if os.IsNotExist(err) {
+		return "", err
+	} else {
+		return testPath, nil
+	}
```
viktorstrate, line 23: Again maybe return a `*string` and set it to nil instead of an empty string.

And then only return an error if the error is somthing other than the file doesn't exist, as then it might be worth logging the error.
See here as an example:
https://github.com/viktorstrate/photoview/blob/940e48fee3f320e60eed284c4c5f49688b011273/api/routes/spa.go#L43-L53

api/scanner/process_photo.go
```diff
@@ -96,10 +96,72 @@ func processPhoto(tx *sql.Tx, imageData *EncodeMediaData, photoCachePath *string
...
+
+
+
+				// save new side car hash
+				_, err = tx.Exec("UPDATE media SET side_car_hash = ? WHERE media_id = ?", currentFileHash, photo.MediaID)
+				if err != nil {
+					return false, errors.Wrapf(err, "could not update side car hash (%d, %s)", photo.MediaID, currentFileHash)
+				}
+			}
+		}
```
viktorstrate, line 161: Maybe move all this (everything inside of `if mediaType.isRaw() { ... }` to a new function.

viktorstrate (changes requested):
Thank you for the pull request. This is a nice, unobtrusive addition.

I haven't tested it yet but it looks good overall. I've added some comments, mostly related to code structure.

I'll do a proper test later though.

viktorstrate:
>After making the changes I realised getting the size for my raw files (NEF) is incorrect, however they would be incorrect anyway if I had a crop in the .xmp file as the previous method encoded the high res JPEG and then read the dimensions from the new JPEG. I changed it to read the dimensions directly from the raw file but I think it is reading the dimensions of the embedded JPEG thumbnail (as the sizes reported are 160x120 instead of (4282x2844). I plan to resolve this but wanted to get a sense if the JPEG update feature was something we wanted in the first place.

Regarding jpeg cropping, can't we just read the dimension of the generated jpeg file (after is has been cropped by darktable)?
The dimensions saved to the database, should not reflect the dimensions of the original photo but the generated photo.

I might be missing your point, in that case, please explain it again.

seam345:
> > After making the changes I realised getting the size for my raw files (NEF) is incorrect, however they would be incorrect anyway if I had a crop in the .xmp file as the previous method encoded the high res JPEG and then read the dimensions from the new JPEG. I changed it to read the dimensions directly from the raw file but I think it is reading the dimensions of the embedded JPEG thumbnail (as the sizes reported are 160x120 instead of (4282x2844). I plan to resolve this but wanted to get a sense if the JPEG update feature was something we wanted in the first place.
> 
> Regarding jpeg cropping, can't we just read the dimension of the generated jpeg file (after is has been cropped by darktable)?
> The dimensions saved to the database, should not reflect the dimensions of the original photo but the generated photo.
> 
> I might be missing your point, in that case, please explain it again.

Yes sorry little tricky to explain I made a little table of it but need to zoom into the image, the important information is in the markdown table.

Screenshot_20201120_194623

 | NEF image 4288x2844 | master branch | seam345:sean/update-preview-on-sidecar-change |
 | ------------------- | ------------- | --------------------------------------------- |
 | With side car file  | original (1311x1949)   large (1311x1949)  small (688x1024) | original (160x120)  large (1311x1949)  small (688x1024) |
 | Without side car file  | original (4288x2844)   large (4288x2844)  small (1024x680) | original (160x120)  large (4288x2844)  small (1024x680) |

To attempt to explain the table, the NEF image has a size of 4288x2844 NEF also embeds a JPEG for fast display (I think most raw formats do). 

When on the master branch, the image is first encoded with the high res function generating the Large file, then the dimensions of this generated image is read and inserted into the DB for both the original and the large image. Resulting in the resolutions reading the same even when darktable cropped the JPEG version because it spotted and used the sidecar file, showing an incorrect dimensions for the original image.

On my branch I attempted to read the dimensions directly from the NEF file (to get around the cropping problem), however I think as the function is written for JPEGs it has found the dimensions for the embedded JPEG in the NEF file resulting in 160x120 regardless of weather the image is cropped.


Hope that all makes sense :)

api/database/migrations/003_sar_side_car_files.up.sql
seam345: Shall do, I didn't originally because I wasn't sure how to handle having a nil or string but looks like a *string will work :)

viktorstrate:
Okay got it.

I tested around a little and found that we might be able to read the dimensions from the exif data.
I managed to build this small example that works:

```go
exifTags, err := exif.Decode(photoFile)
if err != nil {
	return nil, errors.Wrap(err, "Could not decode EXIF")
}

imageWidthRaw, _ := exifTags.Get(exif.ImageWidth)
imageHeightRaw, _ := exifTags.Get(exif.ImageLength)

imageWidth, _ := imageWidthRaw.Int(0)
imageHeight, _ := imageHeightRaw.Int(0)

fmt.Printf("%d x %d\n", imageWidth, imageHeight)
// Prints "5472 x 3648" for an example cr2 raw image
```

seam345:
> ```go
> exifTags, err := exif.Decode(photoFile)
> if err != nil {
> 	return nil, errors.Wrap(err, "Could not decode EXIF")
> }
> 
> imageWidthRaw, _ := exifTags.Get(exif.ImageWidth)
> imageHeightRaw, _ := exifTags.Get(exif.ImageLength)
> 
> imageWidth, _ := imageWidthRaw.Int(0)
> imageHeight, _ := imageHeightRaw.Int(0)
> 
> fmt.Printf("%d x %d\n", imageWidth, imageHeight)
> // Prints "5472 x 3648" for an example cr2 raw image
> ```


I gave it a try and it seems to still outputs 160x120 
This is the file I am trying to read the size of: 
https://drive.google.com/file/d/1o4K8eAsatOHdqMNux5gz7RgCbY06AnOP/view?usp=sharing

I was hoping Darktable would have an option to spit out the metadata but sadly not :(

api/scanner/process_photo.go
seam345: moved :) this function grew slowly so it's size snuck up on me haha

api/scanner/scanner_media.go
seam345: I've amended it to return *string. Bit unsure about the logging, hopefully its formatted correctly.

api/database/migrations/003_sar_side_car_files.down.sql
seam345: Corrected and I also corrected it in 002_site_info_more_settings.down.sql :)

viktorstrate:
I tested your photo and the exif parser did return the right dimensions `5472 x 3648` 🤔

seam345:
Well that was unexpected... I'm quite happy to rename my files :)

viktorstrate:
> Well that was unexpected... I'm quite happy to rename my files :)

I found out this didn't really solve the problem, it was very unstable and got it wrong even though somtimes.

viktorstrate:
It might be because it wrongly finds the dimension of an embedded jpeg file instead.

We could maybe replace the exif library with somthing like this [barasher/go-exiftool](https://github.com/barasher/go-exiftool) - a wrapper around exiftool which certainly gets it right, but this would require users to have that installed. This wouldn't really be a problem in Docker as it could just be bundled with.

seam345:
As I'm using docker, I have no problems with adding exiftool :)

If you want to add it, would you like it in this PR or I can submit another?

viktorstrate:
Cool, lets do it in another pull request. You can just use the current implementation for now, that is use the dimensions of the generated high-res photo.

seam345:
okidoki I'll drop the commit :)

viktorstrate:
I've opened a new issue for exif, see #136.

Is this pr finished, so I can properly test and merge it?

seam345:
yes it is finished :)

I started looking at creating unit tests for the scanner but would possibly require some mocking to reduce the need for many test image files, but shall be something I'll look at doing in the future :)

viktorstrate:
>I started looking at creating unit tests for the scanner but would possibly require some mocking to reduce the need for many test image files, but shall be something I'll look at doing in the future :)

Awesome, I've thought of good ways to test the API, but haven't really found any yet, though I have not a lot of experience with unit testing.

Photoprism has an assets directory with some example photos, these are used in their unit tests. See this file as a very small example https://github.com/photoprism/photoprism/blob/develop/internal/photoprism/timezone_test.go

seam345:
Admittedly my testing experience is limited, but slowly growing. I'll have a think and see what I can do after adding the exiftool :)

viktorstrate (approved):

viktorstrate:
I've pushed some small bug fixes. Everything works as it should, except that the dimensions of RAW images are displayed with the crop.

PR #148, opened 2020-12-06
Fixes viktorstrate/photoview#8

stz184:
- Added new property CounterpartPath to Media struct to hold the path to the counterpart JPEG file (if any)
- Added new MediaType method isBasicSupportedisBasicTypeSupported()
- Added new function isFileExists() to minimize the code duplication


api/scanner/scanner_media.go
```diff
@@ -8,24 +8,92 @@ import (
...
+	fileExtType, found := fileExtensions[strings.ToLower(ext)]
+
+	if found {
+		if fileExtType.isBasicTypeSupported() {
+			return nil
+		}
+	}
+
+	pathWithoutExt := strings.TrimSuffix(imagePath, path.Ext(imagePath))
+	for ext, mediaType := range fileExtensions {
```
viktorstrate, line 82: Maybe add a helper method for `MediaType`s to get the corresponding extensions, something like this

```go
func (mediaType *MediaType) FileExtensions() []string { ... }
```

Then we could just loop over the extension like so

```go
for extension := range TypeJpeg.FileExtensions() { ... }
```

viktorstrate (changes requested):
Looks good, I've added some comments, it's all just small refactorings.
I haven't tested it yet, only looked through the code.

stz184:
Thanks for the comments. I will apply the changes today :)

api/scanner/scanner_media.go
```diff
@@ -8,24 +8,92 @@ import (
...
-	testPath := path + ".xmp"
+func isFileExists(testPath string) bool {
 	_, err := os.Stat(testPath)
 
 	if os.IsNotExist(err) {
-		return nil
+		return false
 	} else if err != nil {
 		// unexpected error logging
 		log.Printf("ERROR: %s", err)
```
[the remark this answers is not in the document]
stz184: It is not an actual error. The function is called `isFileExists` so if a file is missing this is not exactly an error - the intention of the check may be to ensure that the file is missing. If you still want to have a debug log, I will add such.

api/scanner/scanner_media.go
stz184: Thanks for pointing me this out, it's really neat idea!

api/scanner/scanner_media.go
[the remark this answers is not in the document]
viktorstrate: I understand. It's more of a warning. I think it would be fine to add some context to the error log, just so if it appears it will be easier to find out where it came frome.

api/scanner/scanner_media.go
stz184: I used `mediaType MediaType` instead of `mediaType *MediaType` in order to use it directly with constants such `TypeJpeg` otherwise I am getting an error TypeRange is not a pointer.

stz184:
Test instructions:
Variant A) Use the debugger to ensure that the counterpart JPEG file next to the raw image with the same basename is used.
Variant B) Place in the same folder a raw image and JPEG with the same base name but totally different view. The thumbnails should be generated from the JPEG version :)

viktorstrate:
Looks great, I think we talked past each other regarding the logging, I didn't mean you to add another log. I've fixed it in the newest commit.

Thank you! Will merge now

stz184:
> Looks great, I think we talked past each other regarding the logging, I didn't mean you to add another log. I've fixed it in the newest commit.
> 
> Thank you! Will merge now

Sorry about the logging, I misunderstood you.... I thought you want logging anyway so I did such. Once again, excuse me for the misunderstanding.

viktorstrate:
It's no problem at all, I really appreciate your work, thank you 😊

PR #162, opened 2020-12-30
Add navigation control to map

Kjeldgaard:
In places tab, navigation controls are added to the mapbox map, see example.

Navigation


viktorstrate:
Looks good, thank you!

PR #164, opened 2021-01-12
Added db connection retry logic

stz184:
Added db connection retry logic

### DIFF
--- a/api/database/database.go
+++ b/api/database/database.go
@@ -50,23 +50,20 @@ func SetupDatabase() (*gorm.DB, error) {
 	sqlDB.SetMaxOpenConns(80)
 
 	if err != nil {
-		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-		defer cancel()
-
 		for retryCount := 1; retryCount <= 5; retryCount++ {
-			select {
-			case <-ctx.Done():
-				log.Println(ctx.Err())
-				return nil, err
-			default:
-				if err := sqlDB.PingContext(ctx); err != nil {
-					log.Printf("WARN: Could not ping database: %s, Will retry after 1 second", err)
-					time.Sleep(time.Second)
-				} else {
-					return db, nil
-				}
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+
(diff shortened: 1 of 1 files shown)


api/database/database.go
```diff
@@ -39,11 +41,34 @@ func SetupDatabase() (*gorm.DB, error) {
...
+				if err := sqlDB.PingContext(ctx); err != nil {
+					log.Printf("WARN: Could not ping database: %s, Will retry after 1 second", err)
+					time.Sleep(time.Second)
+				} else {
+					return db, nil
+				}
+			}
+		}
+		return nil, err
+	}
```
viktorstrate, line 71: I think that we should use a new context for each attempt, instead of reusing the same for all attempts. I.e. move the creation of ctx inside of the loop, and replace the whole switch statement with the contents of just the default branch, (the .PingContext function will automatically stop and return an error if the timeout is reached).

Also a nice improvement would be to double the wait time for each attempt. For example the first time docker-compose is started it takes some time to setup the database (more than 5 seconds) this would allow it to wait for longer without spamming the pings each second.

I've added a suggestion to implement it below, what do you think?

```suggestion
	if err != nil {
		for retryCount := 1; retryCount <= 5; retryCount++ {
			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
			defer cancel()

			waitSeconds := int(math.Pow(2, float64(retryCount-1)))

			if err := sqlDB.PingContext(ctx); err == nil {
				return db, nil
			}

			log.Printf("WARN: Could not ping database: %s. Will retry after %d second\n", err, waitSeconds)
			time.Sleep(time.Duration(waitSeconds) * time.Second)
		}
		return nil, err
	}
```

viktorstrate (commented):
You mentioned that you got some unexpected errors/warnings when starting the server whilst it's pinging. I tried to do the same, but I couldn't replicate it, it worked as expected for me.

I've added some suggestions in a comment, could you try to see if moving the timeout inside the loop solves it?

stz184:
I've tested and applied the suggested code :) Thanks for the guidelines, @viktorstrate

PR #137, opened 2020-11-23
Release v2.0

viktorstrate:
See [milestone for release 2.0](https://github.com/viktorstrate/photoview/milestone/2).

### Database migration cleanup

- [x] Media sharing
- [x] Filters (sort by)
- [x] Periodic scanner
- [x] Search (#155)
- [x] Reimplement database connection retries (assigned @stz184)

### New features

- [x] Prefix environment variables with `PHOTOVIEW_` to ensure they are unique. (assigned @stz184)
`API_LISTEN_IP` -> `PHOTOVIEW_LISTEN_IP`
`API_LISTEN_PORT` -> `PHOTOVIEW_LISTEN_PORT`

- [x] Add support for database drivers: Sqlite and Postgres
This introduces new environment variables `PHOTOVIEW_DATABASE_DRIVER` and `PHOTOVIEW_SQLITE_PATH`.
Rename `MYSQL_URL` to `PHOTOVIEW_DATABASE_URL`

- [x] Restructure User and Album relation https://github.com/photoview/photoview/issues/67#issuecomment-724978779 #157

Issue #67, opened 2020-08-15
Additional users cannot see photos (single photo path)


kpine:
I probably have a misunderstanding about how multiple users are supposed to work. Can multiple users share the same photo path?

I have a single directory of photos mapped to `/photos` in the Docker container. For the admin user I set `/photos` as the photo path. All the photos appear for that user. I then added a second user (another family member) and also configured that user's photo path to `/photos`, but no photos are visible. Are the photo paths exclusive to each user?

viktorstrate:
Currently each users root path should be not be overlapping with other users, ie. each photo is only owned by one user.

The solution would either be to have each users directory inside of the `/photos` directory, like `/photos/user1` and `/photos/user2`, otherwise multiple mounts could be set up from the `docker-compose.yml` config file.

The problem is that if a photo is shared between two users, it would be a waste to generate the thumbnails twice, once for each user, so it would require some changes to implement this.
Specifically albums/directories should not be bound to a specific user, and photo favourites and public shares should be bound to both a user and a photo instead of only the photo.

numen31337:
I would like to suggest a feature that allows users to have multiple folders. In this way, it would be possible to have a shared folder across multiple accounts and individual folder for those photos that should be bound to the user. This would be very useful when setting up a family library.

viktorstrate:
@numen31337 I think this sounds like a really good approach.
This does require some significant internal restructuring though, but I think it would be worth it for the flexibility it would give.

## Proposed solution

If the database is changed such that the `root_path` of the users and the `owner_id` of albums are removed.
And instead a new many-to-many relation between users and albums is made, then multiple users can share the same album.

As all albums already have the `path` attribute, which describes where on the filesystem the album is located, the `root_path` is not neccessary, as it would be infered from the album it self.

Also public shares could still be infered, as the `share_token` table already has the `owner_id` attribute.

```diff
CREATE TABLE user (
-  root_path varchar(512),
)

CREATE TABLE album (
-  owner_id int NOT NULL,
)

+CREATE TABLE user_album_relation (
+  user_id int NOT NULL,
+  album_id int NOT NULL,
+
+  PRIMARY KEY (user_id, album_id)
+)
```

But this would mean that favorites would be shared across users, which is probably unwanted.
To fix this, the favorites attribute could be moved to a new table like so.

```diff
+CREATE TABLE user_media_data (
+  user_id int NOT NULL,
+  media_id int NOT NULL,
+  favorite boolean DEFAULT FALSE,
[...]

PR #165, opened 2021-01-16
Support Photoview ignore file in user root

Kjeldgaard:
Read .photoviewignore file in user root directory, and excludes
directories which match a .photoviewignore entry.

### DIFF
--- a/api/scanner/scanner_user.go
+++ b/api/scanner/scanner_user.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"fmt"
 	"io/ioutil"
+	"bufio"
 	"log"
 	"os"
 	"path"
@@ -13,7 +14,30 @@ import (
 	"github.com/photoview/photoview/api/graphql/models"
 	"github.com/photoview/photoview/api/graphql/notification"
 	"github.com/photoview/photoview/api/utils"
+	"github.com/zabawaba99/go-gitignore"
 )
+
+func getPhotoviewIgnore(ignorePath string) ([]string , error){
+	var photoviewIgnore []string
+
+	// Open .photoviewignore file, if exists
+	photoviewIgnoreFile, err := os.Open(path.Join(ignorePath, ".photoviewignore"))
+	if err != nil {
+		return photoviewIgnore, err
+	}
+
+	// Close file on exit
+	defer photoviewIgnoreFile.Close()
+
+	// Read and save .photoviewignore data
+   	scanner := bufio.NewScanner(photoviewIgnoreFile)
+   	for scanner.Scan() {
+		photoviewIgnore = append(photoviewIgnore, scanner.Text())
+		log.Printf("Ignore found: %s", scanner.Text())
+	}
+
+   	return photoviewIgnore, scanner.Err()
+}
 
 func findAlbumsForUser(db *sql.DB, user *models.User, album_cache *AlbumScannerCache) ([]*models.Album, []error) {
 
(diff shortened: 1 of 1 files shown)


Kjeldgaard:
I did a quick go at the ignore/exclude directories, similar to .gitignore. It is my first Go code, so if it is hopeless, feel free to reject the pull request. However, it seems to work :)

I have 1 issue though, and that is my test directory contains a single album with 34 images, and a few sub directories which should be ignored. The sub directories are ignored as expected. And when I browse "Albums" I get a single album containing all images as expected. But if i browse "Photos", only 12 images are shown... Having said that, I don't think I caused this (famous last works). But have a look, at let me know what you think.

viktorstrate:
Thank you for the pull request. 

It overall looks good, though I haven’t tested it yet - will do later.

How does it work in regards to nested children, for example if I add an ignore file to one folder will it also apply for subfolders?

And also what happens if one folder contains an ignore file, but also contains a child folder with another ignore folder? In that case I imagine the best solution would be to merge the ignore rules.

Kjeldgaard:
> Thank you for the pull request.
> 
> It overall looks good, though I haven’t tested it yet - will do later.
> 
> How does it work in regards to nested children, for example if I add an ignore file to one folder will it also apply for subfolders?
> 
> And also what happens if one folder contains an ignore file, but also contains a child folder with another ignore folder? In that case I imagine the best solution would be to merge the ignore rules.

I'm afraid your expectations are too high :) This is just a basic implementation, though I would expect it covers the majority of the use cases. It looks for a .photoviewignore at user root only and it is dirs which are excluded, not individual files. The reason for this first approach is two-fold, I expect it covers a large majority of the use cases, and I wouldn't spend too much time on it, if my approach was not what you have envisioned.

Nevertheless, I see two obvious extensions for this.
1. It should work on indiviual files as well. This should be straightforward. I'll start with this as soon as time allows.
2. Support .photoviewignore file in subfolders and not only in user root. Here are several routes to take. I'll have to think more about this.

viktorstrate:
> It looks for a .photoviewignore at user root only and it is dirs which are excluded, not individual files.

Ah I see. That might be just fine for v1, but the next major release (see pull-request #137), the way user paths work has been changed to allow each user to have multiple paths, and also users can overlap paths.

This means though, that checking for a single .photoviewignore file at the top level root path might not be a viable solution anymore.
See this https://github.com/photoview/photoview/issues/67#issuecomment-724978779 for a more detailed explaination of the new structure.

> It should work on indiviual files as well.

This would be a fine addition, but I don't necessarily think that it would be critical as I would imagine by far the largest use case for this would be to ignore whole directories.

----

As the v2 branch is soon to be merged into master I think the best solution would be to find a way to integrate this with the v2 branch.

I don't think it would be that hard though, the only real difference important for this is that now in `findAlbumsForUser` the user has an array of root paths instead of just a single one, nearly everything else is the same.

See this snippet from the v2 branch:
https://github.com/photoview/photoview/blob/a10f8ad1ef5a0b4a5dbfd3aa6d9b19d51b42d0d7/api/scanner/scanner_user.go#L29-L32

Maybe a simple solution would be to add a new attribute to the `scanInfo` struct, that holds the ignore rules for that path.
Then each time a new `scanInfo` object is pushed to the `scanQueue` (see below) it should inherit the ignore rules from its parent.
https://github.com/photoview/photoview/blob/a10f8ad1ef5a0b4a5dbfd3aa6d9b19d51b42d0d7/api/scanner/scanner_user.go#L129-L144

If you think that it would be too much to take on or you don't want to, that's fine I can finish it then.

Kjeldgaard:
Sorry for the late reply. I've been busy with family life. I haven't really looked in to V2 yet. I'll be happy to look at a V2 approach for ignoring folders. However, I can't promise any time frame. When I get some time again, I'll see if you have progressed with this issue for V2.

viktorstrate:
No worries. Let me know if you start on the issue, so I can assign it to you.

Issue #168, opened 2021-01-17
Final work before releasing v2


viktorstrate:
### Additions
- [x] Migrate the maps page with the new database integration. (assigned @stz184 #175)

### Bugfixes
- [x] Scan All button doesn't work (assigned @stz184)
- [x] Add new user from settings page doesn't work (assigned @stz184 #171)
- [x] Delete users from settings page doesn't work (assigned @stz184 #172)
- [x] Videos doesn't seem to work
- [x] EXIF data doesn't show up
- [x] Fix search
- [x] Fix favorites

#### Compatibility error with Gorm and MariaDB when running in Docker
- [x] Fix this problem

This seems to be a periodic error, sometimes when inserting to the database the errors below is thrown.

<details><summary>See error logs regarding this problem</summary>
<p>

```
photoview_1  | 2021/01/17 21:24:58 /app/scanner/scanner_album.go:42 Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'CONFLICT DO NOTHING' at line 1
photoview_1  | [0.507ms] [rows:0] INSERT INTO `users` (`created_at`,`updated_at`,`username`,`password`,`admin`,`id`) VALUES ('2021-01-17 21:24:58.164','2021-01-17 21:24:58.164','viktorstrate','$2a$12$6ketlL58FgVIjtITQpaBh.CAFF5dciorwa/5td8UdKqGua8QwrxDm',true,1) ON CONFLICT DO NOTHING
photoview_1  |
photoview_1  | 2021/01/17 21:24:58 /app/scanner/scanner_album.go:42 Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'CONFLICT DO NOTHING' at line 1
photoview_1  | [0.734ms] [rows:0] INSERT INTO `user_albums` (`album_id`,`user_id`) VALUES (1,1) ON CONFLICT DO NOTHING
photoview_1  |
photoview_1  | 2021/01/17 21:24:58 /app/scanner/scanner_album.go:42 Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'CONFLICT DO NOTHING' at line 1; Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'CONFLICT DO NOTHING' at line 1
[...]

PR #171, opened 2021-01-21
* Fixes photoview/photoview#168

stz184:
Fixed issue adding new users.
I wanted to use @exports graphql directive as described [here](https://www.apollographql.com/docs/link/links/rest/#export-directive), but the backend server keeps saying `Unknown directive "export".`. I checked the corresponding github page on the used go graphql server and I didn't find any information about the support status of this directive so I guess it just not implemented yet.

So I chained the mutations for creating user and adding user path using the onComplete directive.
In case of error adding the path (for example invalid path or typo) the user is still added but an error message is shown. 
The user has to click the Edit button and retry adding the album path(s). 
I found that this much easier than keeping the adding user form open but changing the "Add user" action to just adding the path because the user has been successfully added and just adding the path failed. 
Of course, I am open to any suggestions that may improve the user experience.

### DIFF
--- a/ui/src/Pages/SettingsPage/Users/AddUserRow.js
+++ b/ui/src/Pages/SettingsPage/Users/AddUserRow.js
@@ -4,16 +4,20 @@ import React, { useState } from 'react'
 import { Button, Checkbox, Input, Table } from 'semantic-ui-react'
 
 const createUserMutation = gql`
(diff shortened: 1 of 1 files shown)


viktorstrate:
Looks good, thank you

PR #172, opened 2021-01-23
* Fixes photoview/photoview#168

stz184:
Fixed issue deleting existing users.
It handles the associated albums and deletes them only if they are not referenced by any other users.
To keep the data consistent, the whole operation is wrapped in a transcation.
If everything goes well, the corresponding cache directories are also deleted.


stz184 (commented):
Sorry for the dumb PR

viktorstrate:
Not a problem, just some stupid mistakes.

PR #175, opened 2021-01-28
Fixes photoview/issues#168

stz184:
- Migrated mediaGeoJson resolver to GORM
- Fixed issue loading media list in MediaList() function
- Fixed typo in the column "gps_lonitude" (renamed to "gps_longitude")

### DIFF
--- a/api/database/database.go
+++ b/api/database/database.go
@@ -130,8 +130,6 @@ func SetupDatabase() (*gorm.DB, error) {
 }
 
 func MigrateDatabase(db *gorm.DB) error {
-
-	db.Migrator().RenameColumn(&models.MediaEXIF{}, "gps_lonitude", "gps_longitude")
 	db.AutoMigrate(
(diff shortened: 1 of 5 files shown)


viktorstrate:
Awesome, looks great!

I noticed that the database rename migration prints an error if it didn't find the misspelled column. I think that we should either check that the column needs to be renamed first. Or maybe just don't do the migration at all, as we haven't released v2 yet, we don't need to support it, this might be the cleanest. What do you think?

stz184:
> Awesome, looks great!
> 
> I noticed that the database rename migration prints an error if it didn't find the misspelled column. I think that we should either check that the column needs to be renamed first. Or maybe just don't do the migration at all, as we haven't released v2 yet, we don't need to support it, this might be the cleanest. What do you think?

I can remove the migration, that's the easiest approach indeed. I am doing it now.
No need to make the code more complex than needed.

Issue #149, opened 2020-12-07
Album sharing: scroll down issue


vrenaville:
Hello,

I have an issue, with the release 1.1.3 ARM docker image.

When I share an album, I can only see the first part of the album.
If I scroll down image is not visible, seems to be an JS issue .

![Selection_085](https://user-images.githubusercontent.com/355215/101328225-08cbd200-3870-11eb-8593-8c3953d7abc4.png)

viktorstrate:
This is related to the react-lazyload library. This could maybe be solved by calling [forceCheck](https://github.com/twobin/react-lazyload#forcecheck) after the page has fully loaded

vrenaville:
@viktorstrate 
Hello,

Sorry for last reply, I have try to check how to fix that, but I'm a newbie on JS , do you have any pointer where I need change it ?

Thanks for your help and work.

viktorstrate:
I can see that somthing like this is already done here

https://github.com/photoview/photoview/blob/f2212988b74617d56df0f0f9dca3482142f3ab1e/ui/src/components/photoGallery/PhotoGallery.js#L89-L91

You can try to experiment with this, maybe try to call it every time it re-renders (move it out of the useEffect) and see if it helps.

I apreciate your help :)

vrenaville:
@viktorstrate 

Hello,

I have try without luck, to correct this issue :/ .

Am I the only one to have this issue ?

viktorstrate:
Solved in fa35ea1

PR #178, opened 2021-01-31
Check lazy load on scroll in PhotoGallery view

andrasmaroy:
My attempt at fixing #149.

I'm no frontend expert, but this seems to fix the issue for me.

### DIFF
--- a/ui/src/components/photoGallery/PhotoGallery.js
+++ b/ui/src/components/photoGallery/PhotoGallery.js
@@ -88,11 +88,12 @@ const PhotoGallery = ({
 
   useEffect(() => {
     !loading && forceCheck()
+    /*eslint no-unused-vars: ["error", { "argsIgnorePattern": "^_" }]*/
     const onScroll = _ => {
-      !loading && forceCheck();
+      !loading && forceCheck()
     }
 
-    window.addEventListener("scroll", onScroll)
+    window.addEventListener('scroll', onScroll)
     return () => {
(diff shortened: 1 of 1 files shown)


viktorstrate:
Thank you for the pull request. Correct me if i'm wrong, but doesn't [react-lazyload](https://github.com/twobin/react-lazyload) already do this?

With my experience this problem doesn't happen when scrolling, only sometimes when resizing the window, or when there are too few photos on the page to being able to scroll.

Can you think of a way where I can test this reliably? In what case(s) did you experience the bug where this change fixed it?

andrasmaroy:
I encountered this issue when opening a shared album in a window that didn't fit all the photos. The photos that fit in the beginning loaded correctly, the rest stayed grey as reported in the issue.

Tested in safari and firefox, both reproduced the issue.

viktorstrate:
Okay thank you. I will investigate it further.
I would like if we could fix this without adding another scroll listener, as I think it could hurt performance, but this info is a good starting point.

viktorstrate:
I found a layout issue with album shares, and it also solved the lazy load problem.
See my fix in commit fa35ea1.

I will close the original issue aswell, but if it still isn't fixed please write a comment in the issue and I will open it again.
Thank you for providing the steps to reproduce, and for taking the time to attempt a fix.

andrasmaroy:
Looking good, thanks for fixing it properly :)

Np, I know how hard things are to fix without knowing how to repro.

PR #180, opened 2021-01-31
Implement postgres support

andrasmaroy:
Adding support for posgres, clicked through the major functions on UI with all 3 drivers and everything seems to be working fine.


viktorstrate:
Awesome thank you for the pull-request!

Looks good, I will test it properly soon. 👍

PR #193, opened 2021-02-12
Infinite scroll pagination

viktorstrate:
### Todo

- [x] Timeline page
- [x] Albums page
- [x] Loading indicator


eyenx:
Has pagination been added to public shared albums too? On a shared album I just see 200 photos and no more.

viktorstrate:
@eyenx you're right the limit is just hard-coded to 200. This should be an easy fix though. I'll open a new issue for it to better remember it. Thank you!

PR #210, opened 2021-02-23
V2 ignore directories

Kjeldgaard:
Hi,

I've updated the ignore directory feature to work with V2. It looks for .photoviewignore files in all folders. If a .photoviewignore file is found, the rules are appended to the current set of rules. All subdirs inherit the ignore rules from the parent. Only dirs are ignore, not files.

I can see that I messed up rebasing, so it looks like I modified 48 files. The only file I modified is 'scanner_user.go'.


viktorstrate:
Awesome, looks great!
Do you think it would be possible to fix the rebase problem, I'm not sure if merging this could cause problems with git.
Maybe the simplest way would be to just manually copy the changes to a fresh branch?

Kjeldgaard:
Sure, I'll create a new branch and pick that single commit i did. I expect to do it tonight.

PR #213, opened 2021-02-23
V2 Ignore Directories

Kjeldgaard:
Looks for .photoviewignore file in all root dirs and potential subdirs.
Subdirs inherit the ignore rules from parent dir. If a .photoviewignore
file is found in a subdir, the new ignore rules are appended to the
existing ignore rules.

Ignore match engine is from 'github.com/sabhiram/go-gitignore'.

### DIFF
--- a/api/scanner/scanner_user.go
+++ b/api/scanner/scanner_user.go
@@ -101,7 +101,7 @@ func findAlbumsForUser(db *gorm.DB, user *models.User, album_cache *AlbumScanner
 
 		// Skip this dir if in ignore list
 		ignorePaths := ignore.CompileIgnoreLines(albumIgnore...)
-		if (ignorePaths.MatchesPath(albumPath)) {
+		if (ignorePaths.MatchesPath(albumPath + "/")) {
 			log.Printf("Skip, directroy %s is in ignore file", albumPath)
 			continue
 		}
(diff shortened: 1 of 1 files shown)


Kjeldgaard:
Now a clean PR.

viktorstrate:
I just tested it and it overall looks great!

### Trailing slashes
One small issue that I found while testing is that it won't ignore patterns if they end in a slash.
As albums are directories I think that should be allowed.

```
test_album/   # will not work
test_album    # does work
```

I think it would be as easy as replacing
```go
if (ignorePaths.MatchesPath(albumPath))
``` 
with
```go
if (ignorePaths.MatchesPath(albumPath + "/")) {
```


### Support files as well?

I thought it would be a nice addition if files/media could be ignored as well, I imagine it could be nice if you eg. wanted to completely ignore all media of a certain file extension.

I think this could easily be added, but if you don't want to I will gladly merge it anyways.
For reference, the logic would likely have to be added to the `findMediaForAlbum()` function under `scanner_album.go`

Kjeldgaard:
Great! I've also wondered about the trailing slash behaviour. Anyway, I'll change it, such that 'test_album/' also match.
Regarding matching files/media, I think I'll do that as a separate issue if that is okay with you.

viktorstrate:
No issue, I will merge this then. 👍

emielmolenaar:
This is great stuff, thanks!

PR #219, opened 2021-02-26
Add environment variables section

throwaway876:
I had a bit of trouble finding the right format/name for environment variables between the website and github (and in the docker-compose sample.) 
I believe this is correct based off my kubernetes manifests and how it has worked for me. Will also serve as a reference point to add new variables as needed throughout development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ The photos will have to be scanned before they show up, you can start a scan man
 > This section describes all available environment variables that can be configured.
 
 > database related
+
 Required |Variable | Default | Notes 
 ---------|---------|---------|------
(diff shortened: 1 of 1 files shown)


viktorstrate:
@throwaway876 thank you.
I just added the same info to the documentation here
https://photoview.github.io/docs/installation-environment-variables/

I would rather have a single place for this information there isn't multple places to update when new environment variables are added.

I do though like the table structure you've used, it makes it very readable. Maybe we could use that format on the documentation page?

throwaway876:
Submitted a PR on the webpage repo. Might need to check the required checkmarks (gets sticky depending on if one is set then it requires a follow on one- haven't thought how to make it clearer than it is.)

PR #221, opened 2021-02-26
Ignore media files

Kjeldgaard:
Added feature to ignore specific media files. E.g. to ignore .png media
files, add '*.png' to .photoviewignore file or add 'img_1234.jpg' to
ignore a specific file. This applies to the current dir and any
potential subdir. Matching is case sensitive. File ignore pattern must
include a '.', otherwise the pattern is treated as a dir ignore pattern.

### DIFF
--- a/api/graphql/models/album.go
+++ b/api/graphql/models/album.go
@@ -17,7 +17,6 @@ type Album struct {
 	Owners   []User `gorm:"many2many:user_albums"`
 	Path     string `gorm:"not null"`
 	PathHash string `gorm:"unique"`
-	IgnoreFiles string
 }
 
 func (a *Album) FilePath() string {
--- a/api/graphql/models/album.go
+++ b/api/graphql/models/album.go
@@ -17,6 +17,7 @@ type Album struct {
 	Owners   []User `gorm:"many2many:user_albums"`
 	Path     string `gorm:"not null"`
 	PathHash string `gorm:"unique"`
+	IgnoreFiles string
 }
 
 func (a *Album) FilePath() string {
--- a/api/scanner/scanner_album.go
+++ b/api/scanner/scanner_album.go
@@ -3,7 +3,10 @@ package scanner
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"path"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/photoview/photoview/api/graphql/models"
@@ -136,8 +139,28 @@ func findMediaForAlbum(album *models.Album, cache *AlbumScannerCache, db *gorm.D
 		return nil, err
 	}
 
+	// Get file ignores
+	var ignores []string
+	if (len(album.IgnoreFiles) > 0) {
+		ignores = strings.Split(album.IgnoreFiles, ",")
+	}
+
 	for _, item := range dirContent {
 		photoPath := path.Join(album.Path, item.Name())
+
+		skipFile := false
+		for _, ignore := range ignores {
+			match, _ := filepath.Match(ignore, item.Name())
+			if (match) {
+				// Skip file if it matches an ignore entry
+				log.Printf("Ignore match, ignore = %s, file = %s\n", ignore, item.Name())
+				skipFile = true
+				break
+			}
+		}
+		if (skipFile) {
+			continue
+		}
 
 		if !item.IsDir() && isPathMedia(photoPath, cache) {
 			// Skip the JPEGs that are compressed version of raw files
--- a/api/scanner/scanner_album.go
+++ b/api/scanner/scanner_album.go
@@ -5,8 +5,6 @@ import (
 	"io/ioutil"
 	"log"
 	"path"
-	"path/filepath"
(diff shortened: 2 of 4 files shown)


Kjeldgaard:
First attempt at a media file ignore feature. Comments are welcome.

api/scanner/scanner_user.go
```diff
@@ -39,6 +55,20 @@ func getPhotoviewIgnore(ignorePath string) ([]string , error){
    	return photoviewIgnore, scanner.Err()
 }
 
+func getIgnoreFiles(ignoreAll []string) (string) {
```
viktorstrate, line 58: I'm not quite sure why this returns a single string rather than an array?
Nevertheless the function could maybe also be deleted if `sabhiram/go-gitignore` was used for files as well.

api/scanner/scanner_user.go
```diff
@@ -169,7 +200,7 @@ func findAlbumsForUser(db *gorm.DB, user *models.User, album_cache *AlbumScanner
 						return err
 					}
 				}
-
+				album.IgnoreFiles = getIgnoreFiles(albumIgnore)
```
viktorstrate, line 203: This could be saved in the `cache` struct, as also mentioned earlier.

viktorstrate (changes requested):
Hi @Kjeldgaard thank you for the pull-request. I've added some comments. I'd like to hear what you think as well. Please feel free to reply to them if you think I've missed something or it's unclear.

api/graphql/models/album.go
```diff
@@ -17,6 +17,7 @@ type Album struct {
 	Owners   []User `gorm:"many2many:user_albums"`
 	Path     string `gorm:"not null"`
 	PathHash string `gorm:"unique"`
+	IgnoreFiles string
```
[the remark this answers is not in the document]
Kjeldgaard: Cool! It was totally the intention. But I agree, AlbumScannerCache is preferable. I was, wrongly, under the impression that AlbumScannerCache was only "valid" when an album/dir was scanned the first time. I'll change this.

api/scanner/scanner_album.go
```diff
@@ -136,9 +139,29 @@ func findMediaForAlbum(album *models.Album, cache *AlbumScannerCache, db *gorm.D
...
+	if (len(album.IgnoreFiles) > 0) {
+		ignores = strings.Split(album.IgnoreFiles, ",")
+	}
+
 	for _, item := range dirContent {
 		photoPath := path.Join(album.Path, item.Name())
 
+		skipFile := false
+		for _, ignore := range ignores {
+			match, _ := filepath.Match(ignore, item.Name())
```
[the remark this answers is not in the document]
Kjeldgaard: I was actually considering to use path/filepath for both dirs and files. At one point I had a problem with the gogitignore implementation regarding file ignore. I'll do some testing in Go Playground, and use sabhiram/go-gitignore if possible.

api/scanner/scanner_user.go
```diff
@@ -17,6 +18,21 @@ import (
...
+func isFileIgnore(ignoreEntry string) (bool){
+	if (strings.Contains(ignoreEntry, "/")){
+		// Folder ignore if entry contains '/'
+		return false
+	}
+
+	if (strings.Contains(ignoreEntry, ".")){
+		// File ignore if entry contains '.' and not '/'
+		return true
+	}
```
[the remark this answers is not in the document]
Kjeldgaard: See above.

api/scanner/scanner_user.go
Kjeldgaard: Will do.

api/scanner/scanner_user.go
Kjeldgaard: Because I encountered a problem I didn't knew how to solve otherwise :) (https://stackoverflow.com/questions/52244001/go-gorm-mysql-unsupported-type-string-a-slice-of-string). But this will not be necessary if using either AlbumScannerCache or sabhiram/go-gitignore.
Disclaimer: This is the first time programming in Go. For a while I wanted to learn a new language, I strongly considered Rust, but then I stumbled on your project here, and decided to give Go a shot. So I "apologies" for the unusual Go code I might write.

viktorstrate: Okay it makes sense if you'd save it to the database.

No issue, your code looks clean :)

api/scanner/scanner_album.go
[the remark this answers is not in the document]
viktorstrate: Okay sure, let me know if you find any problems when using `sabhiram/go-gitignore`.

Kjeldgaard:
Ok, so I'm now using cache for storing ignore data. And I'm not using path/filepath for any file matching anymore. I had an issue with albums showing up, despite the album only contained files which should be ignored. It required me to change directoryContainsPhotos(), but now it passes my own tests.

viktorstrate:
Looks great, I'll test it my self in one of the following days 👍

Issue #243, opened 2021-03-07
Localization in different languages


Zazou89:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Add a localization system. For my example, some members of my family doesn't speak English. I can help to translate in French.

**Describe alternatives you've considered**
Stay in English

**Additional context**
No

viktorstrate:
This would probably be the best library for the job:
https://react.i18next.com/

Issue #94, opened 2020-10-10
Advanced Search (search by exif)


viktorstrate:
Add a new page for more advanced searches.
Which allows the user to filter by:

- Date range
- Filter media type (video or photo or both)
- General exif fields (camera, lens, flash settings etc.)

Maybe inspired by [Github's advance search](https://github.com/search/advanced)

Also add search tags in header search bar. like `type:video date:>10-02-2020 landscape` to search for `videos` after `10-02-2020` including the word `landscape`

Surrealcrow:
Hi! 
I'm the one who propose this feature,

I use this feature alot in dark table, it has an option named "collect images" where you can specify the filter, i usually filter by camera but when looking for an specific image I use the date taken filter as i remember aproximately in which date i did take the photo. (adding some screenshoots of the filters)
![alloptions](https://user-images.githubusercontent.com/70404884/95690521-1fdaa480-0bde-11eb-87d9-d952b32b9895.png)
![cameraview](https://user-images.githubusercontent.com/70404884/95690522-20733b00-0bde-11eb-867c-1d25de2b3f94.png)
![Foldersview](https://user-images.githubusercontent.com/70404884/95690523-210bd180-0bde-11eb-83ae-5b9458190a40.png)

viktorstrate:
Thank you @Surrealcrow, I will take that into account. A solution could be to add a filter dropdown instead of the `Show only favorites` toggle. That lets the user filter by some of the same things as darktable.

acc557:
It would be great if Adobe's XMP Face region metadata can also be made searchable. 
https://github.com/bpatrik/pigallery2 has this implemented.

gdepre:
place/town search would be useful (I use it a lot at work for searching photos taken in one specific place at different dates.

ericerkz:
Searching by date/person would be absolutely helpful. (being able to type someone's name in the search feature, or the date ex: 
"January 30th") and having results query in that way.

MinchinWeb:
I think searching by camera and/or lens would be a cool way to explore my library. I hope this can eventually be added :)

rt87:
+1

kkovaletp:
As was asked in the #564, this page should also have a media type filter, so that users can search for particular types (videos/images), and even exact source formats

PR #258, opened 2021-03-19
Use exiftool for parsing exif data

Kjeldgaard:
Use exiftool to parse exif metadata.


Kjeldgaard:
Tested with images from Canon camera and Samsung phone. Probably not working for all makes as I would expect different exif metadata formatting.

api/graphql/models/media_exif.go
```diff
@@ -12,11 +12,11 @@ type MediaEXIF struct {
...
 	Exposure        *string
 	Aperture        *float64
-	Iso             *int
+	Iso             *int64
 	FocalLength     *float64
 	Flash           *string
-	Orientation     *int
-	ExposureProgram *int
+	Orientation     *int64
+	ExposureProgram *string
```
viktorstrate, line 19: Changing the type of `ExposureProgram` to a string causes two problems, first it would require some migration to make updating seemless. But more importantly, keeping it as an integer would make localization much easier in the future (see #243).

Do you think it's possible to get the raw integer value from `go-exif`?

api/scanner/exif/exif.go
```diff
@@ -31,7 +31,7 @@ func SaveEXIF(tx *gorm.DB, media *models.Media) (*models.MediaEXIF, error) {
 		}
 	}
 
-	var parser exifParser = &internalExifParser{}
+	var parser exifParser = &externalExifParser{}
```
viktorstrate, line 34: I see that you've deleted the internalExifParser, I think it would be best to keep that one and on startup check if `exiftools` is installed and set `parser` to the external or internal parser based on that.

viktorstrate (changes requested):
Thank you for the PR, it generally looks good. I have added some comments to the code  :)

Also I can see that the code scanner found some unused `err` variables, regarding `regexp.Compile`, I don't think it's necessary to handle these as the regex is hard coded, just replace the `err` with `_` to suppress these warnings.

api/scanner/exif/exif.go
Kjeldgaard: Good point, didn't consider that :) I'll keep the internal parser.

api/graphql/models/media_exif.go
Kjeldgaard: Yes, that should be doable.

Kjeldgaard: Okay, so to make Exiftool return ExposureProgram as an integer, I need to pass an additional argument "-n" to Exiftool. According to https://github.com/barasher/go-exiftool/blob/master/exiftool.go it is possible with extraInitArgs, but I can't figure out the correct syntax. Can you help me figure out the correct NewExiftool() call with "-n" as additional argument?

viktorstrate: Okay, I haven't tested it but from looking at the code you linked quickly I am guessing that you should do something like this:

```go
et, err := exiftool.NewExiftool(func(opts *Exiftool) error {
  opts.extraInitArgs = []string{"-n"}
  return nil
})
```

I might be wrong though :)

api/scanner/exif/exif.go
viktorstrate: Perfect :)

api/graphql/models/media_exif.go
Kjeldgaard: That is similar to what I found out. Unfortunately, .extraInitArgs is not exported, thus it is not possible to parse additional startup arguments to exiftool.

So there are different ways forward:
1) Extract metadata as strings (this PR)
2) Fork go-exiftool, create support for parsing additional arguments to exiftool, and extract all metadata as integers. Note, this means that e.g. flash information is also returned as an integer. See more here: https://exiftool.org/TagNames/EXIF.html and https://exiftool.org/TagNames/EXIF.html#Flash So that would affect the internal exif parser as well.
3) ...

What do you think? In my opinion, I think this PR should be modified to make it an intermediate step towards 2) and accept that the extraction of metadata with exiftool is a work in progress.
I haven't looked much at the internal exif parser, but is it possible to extract e.g. the flash information as an integer?

viktorstrate: I think the best solution would be to open a pull request to the `go-exiftool` repo, making `extraInitArgs` public.
If they don't merge it then we'll maintain our own copy.

I think it's rather simple to extract the flash as an integer by using [internalExifParser.readIntegerTag](https://github.com/photoview/photoview/blob/a14b12b8d49abc20e4308ceb0ba92815db671701/api/scanner/exif/exif_parser_internal.go#L171) instead:

```go
flash, err := p.readIntegerTag(exifTags, exif.Flash, media)
```

As database migrations can add a lot of complexity, I think it would be best if we modified the database structure as little as possible. I think the approach with the least amount of extra work would be to first merge this when `go-exiftool` has been updated so flash and exposure is parsed properly.

Would you mind making the pull request to `go-exiftool`, or would you rather I did it?

Kjeldgaard: I forked go-exiftool yesterday and created a function for parsing additional arguments. I changed the import to my forked version. But then it didnt build... Should I manually edit the go.sum and go.mod files? I just deleted all barasher/go-exiftool references, not sure if that was my build problem.

viktorstrate: Hmm, maybe you have to run `go get github.com/Kjeldgaard/go-exiftool` to download it to the go cache first, then remove the original `go-exiftool` from Photoview's `go.mod` file to avoid conflicts.

Kjeldgaard:
So I have it working again. Currently, using my own fork of go-exiftool. I've created a PR (https://github.com/barasher/go-exiftool/pull/24) to implement the required functionality. We can either wait for the PR to be accepted, or initially use my fork.

Please have a look at MediaSidebar.js, if flash metadata is 0, the flash string is not displayed, and I can't figure out why...

viktorstrate:
Looks good at a quick glance, I'll test it properly later and also take a closer look at MediaSidebar.js.

viktorstrate (approved):

viktorstrate:
I fixed the problem in MediaSidebar.js see the latest commit (da664c7).

Everything looks good except that the exposure time is formatted in different ways for the internal and external parser.
Internal saves it as a fraction string eg. `1/50` whereas the external saves it as a decimal eg. `0.02`.
The easiest solution would probably be to format the external parser before saving it to the database, and this is probably fine for now.

But I think the best solution in the long run would be to save it as a number in the database rather than a string, but this would require some database migration. This would be nice as it would make it easier to implement exif filtering later on (see #94).

Kjeldgaard:
Barasher released a new version of go-exiftool with support for "no print conversion", so I am using that now. I've also changed exposure time to a float.
Finally, flash string is still not shown if value is 0, or I am doing something wrong... Anyway, it is probably not important, since 0 means the flash is off.

viktorstrate:
- There might still be some problems with the side bar. It might be a good idea to add UI tests for this.
- We also need some manual database migration as the automatic one cannot convert the `exposure` column if it has entries with the old fractional values eg. the string `1/60`.

I will gladly merge this pr and do the last part before next release if you'd like.

Kjeldgaard:
> next

That sounds good to me if you'll take care of the UI and database tasks.

viktorstrate:
Sure thing, thank you for the effort.

PR #275, opened 2021-03-30
Replace react-lazyload with custom IntersectionObserver

stz184:
- If it is possible, HTMLImageElement.loading is used a lightweight native browser lazy loading solution
- Custom lazyLoad library is created depending on IntersectionObserver. If IntersectionObserver is not available, the images are eagerly loaded
- removed react-lazyload from the dependencies

The implementation was tested on desktop Chome and Firefox, as well on mobile Android Firefox and Chrome browsers.


ui/src/components/photoGallery/ProtectedMedia.js
```diff
@@ -1,6 +1,9 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
+const isNativeLazyLoadSupported = false //'loading' in HTMLImageElement.prototype
```
viktorstrate, line 4: I don't understand the purpose of this variable. Does it ever change if it's const?

ui/src/components/photoGallery/ProtectedMedia.js
```diff
@@ -18,17 +21,24 @@ const getProtectedUrl = url => {
...
  */
-export const ProtectedImage = ({ src, ...props }) => (
-  <img
-    key={src}
-    {...props}
-    src={getProtectedUrl(src)}
-    crossOrigin="use-credentials"
-  />
-)
+export const ProtectedImage = ({ src, loading, ...props }) => {
```
viktorstrate, line 24: I like the use of a new prop to specify `eager` and `lazy` loading, but as `loading` is used a lot as a boolean prop I think it would be better to called it something else like `loadingMode`.

Also I think a quick comment explaining that it should be either `eager` or `loading` and that you need to manually update it with `lazyLoad` if set to `loading` would be nice, as it's no obvious directly from the code.

viktorstrate (changes requested):
Awesome great work!
I can see that it generated a version 1 `package-lock.json` and modified the file substantially, are you using an old version of npm?
Can you change it back to v2, you should be able to just run `npm install` with the newest version of npm to update it.

I've also added a few comments in the code :)

ui/src/components/photoGallery/ProtectedMedia.js
stz184: This variable is used to check if native lazy load support via  attribute supported.
For testing purposes, it was hardcoded to false in order to check the fallback implementation of IntersectionObserver.
Fixed it.

ui/src/components/photoGallery/ProtectedMedia.js
stz184: I understand your idea and reasoning to suggest renaming of `loading` to `lazyLoad` but... I like the idea to keep component props names as close as possible to the underlying HTML elements. For example, ProtectedImage `src` prop is used to set the `src` attribute of the underlying `` HTML element and `loading` prop is used to set `loading` attribute of `` . 
In my opinion, it is much easier to debug code if the variable names do not change when passed as an argument to the descending functions. 

Anyway, if you really think that setting `lazyLoad` prop to `<ProtectedImage />` it will be moe intuitive approach for setting `loading` attribute of underlying `` element - I am fine with refactoring it, just confirm iit.

stz184:
I wasn't satisfied with my initial implementation so I changed it:
- LazyLoad helper is now a singleton class reusing the same IntersectionObserver instance for performance reasons
- IntersectionObserver is now disconnected when the component is unmounted for performance reasons
- `<StyledPhoto />` is now using `useCallback` to avoid unnecessary rerenderings
- Fixed issue setting loading attribute to the `` rendered by `<ProtectedImage/>` component
- Retested with desktop (Linux) and mobile (Android 10) Firefox and Chrome browsers

Enjoy :)

ui/src/Pages/AlbumPage/AlbumPage.js
```diff
@@ -124,6 +125,17 @@ function AlbumPage({ match }) {
     [setOnlyFavorites, refetch]
   )
 
+  useEffect(() => {
+    LazyLoad.loadImages(document.querySelectorAll('img[data-src]'))
+    return () => LazyLoad.disconnect()
+  })
```
viktorstrate, line 131: Add no dependencies to useEffect to only call disconnect when unloaded.
disconnect was called on each page load without this

```suggestion
  }, [])
```

ui/src/components/photoGallery/ProtectedMedia.js
viktorstrate: That makes sense, I see pros and cons for each approach, so feel free to keep it as is if you think that's the best option. If we just make sure to document it then it should be easy to understand nevertheless.

viktorstrate (changes requested):

ui/src/Pages/AlbumPage/AlbumPage.js
stz184: Yeah, I know this, but I forget it accidentally.
Thank you for noticing it!

ui/src/components/photoGallery/ProtectedMedia.js
stz184: Fixed it, check the latest commit :)

stz184:
- Added missing dependencies array on useEffect hook in ProtectedMedia.js
- The new property of ProtectedImage component is renamed from loading to lazyLoading and its type is changed to boolean. Since  attribute is pointless when it is not supported by the browser or when it is used with default value "eager", there is no much sense in setting it at all.
- Added description how does the new property works as part of the docblock.

I hope it is "almost" perfect now :))

viktorstrate:
Looks close to perfect ;) - thank you

PR #316, opened 2021-04-15
Resumption of translation 

Zazou89:
First French translation was probably made with Google Translate. I need to start translation from scratch. I just urgently updated few strings.


viktorstrate:
Okay thank you for improving it. Let me know when you are finished and I'll merge it.

PR #323, opened 2021-04-19
Spanish translation

danielsb:
First pass for the Spanish translation. I tried to stick to the length of the original strings, but we might want to double check once this is implemented. 

Please, let me know if there's a problem!


viktorstrate:
Looks good, thank you!

PR #330, opened 2021-04-20
Add german translation

robin-moser:
Hi,

first of all, a huge Thanks for this project, it fits perfectly for my personal media collection, I absolutely love it!

After seeing the localization options, I added a German translation and generated the needed files. It would be great, if the PR gets accepted.

After all, I have a question about the code generation:
I tried automatically generating the ui code with `npm run genSchemaTypes`, but this changed all occurrences of single quotes to double quotes in all generated files. I needed to revert everything and make the needed change manually. Did I make something wrong or is there a setting to generate the correct code in favor of the existing code?

### DIFF
--- a/api/graphql/generated.go
+++ b/api/graphql/generated.go
@@ -1849,6 +1849,7 @@ enum LanguageTranslation {
   Danish,
   Spanish,
   Polish,
+  German,
 }
 
 type UserPreferences {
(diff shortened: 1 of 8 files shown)


viktorstrate:
> I tried automatically generating the ui code with npm run genSchemaTypes, but this changed all occurrences of single quotes to double quotes in all generated files.

Hmm, I've not noticed that before. It should be automatically fixed by prettier in the pre-commit hook though.

Looks good, thank you

PR #346, opened 2021-04-27
Replace debian with alpine in Dockerfile

viktorstrate:


viktorstrate:
https://github.com/richfelker/musl-cross-make/issues/79

deric:
Beware that performance might go down with musl, especially on memory intensive operations (like indexing). It might be better solution to reduce current image size to absolute minimum.

viktorstrate:
@deric thank you for the insight. I think the optimal solution would be to statically compile the API into a single binary, that could be shipped with the `scratch` or `busybox` as the base image.

I tried that a couple of month ago when working on it, but `libheif` doesn't come pre-compiled as a static library anywhere, and compiling it manually seemed like a big task, especially if it had to be done everytime the Dockerfile was built. Alternativeley it could be pre-compiled and simply copied over each time the Dockerfile is built.

I'd love to hear if you have any suggestions for reducing filesize.

deric:
@viktorstrate Converting everything to Alpine/busybox might be a bit premature optimization. Yes, you'll save a few MB on disk, but using a larger ecosystem (like Debian) that supports multiple architectures out of a box, is a huge benefit. Also when you're using "standard" images, Docker is capable of reusing lower layers, so you don't have to care so much about base image size. 

I guess many people would like to run a "photo server" on Raspberry PI / NAS server, where the storage is at least couple o GB, but you have to support ARM. At this point I'd say it's more important to spent deveploper's time on functionality rather than solving strange compiling issues for various platforms.

I think Go/JavaScript is a good choice for such software, it should be possible to develop fairly quickly and support many platforms. 

Thanks for all your work. Well done! :clap:

viktorstrate:
I've thought of replacing it with [RawTherapee](https://rawtherapee.com) as an replacement for Darktable, I think that's what Darktable builds on top of.

Thank you for helping out ☺️

Issue #361, opened 2021-05-04
Problem with cifs share - Could not read directory


muCruiser:
I have a cifs share on synology mounted to the docker container on a raspberry pi
After scanning it get the error message:

Scanner error
Could not read directory (/share/***directory***/***sub-directory****/): readdirent: no such file or directory

basically photoview has access and can read the directory, but then has a problem

if found this issue desribed for go
[https://github.com/golang/go/issues/39237](https://github.com/golang/go/issues/39237)

can you implement the described action

`export GODEBUG=asyncpreemptoff=1`

or is there another solution?

viktorstrate:
Thank you for the detailed report, I've heard others having the same issue as well.

Although adding the `GODEBUG` variable might solve the issue, I don't know if it introduces other problems, and from the thread you've linked it doesn't seem to be a perfect solution.

Does it work to add the environment variable to the docker-compose file?
If so, it's probably the easiest to let users opt in to that fix. Also I don't know if that problem will be solved in a future release of Go or CIFS.

```diff
  photoview:
    image: viktorstrate/photoview:2
    restart: always
    ports:
      - "8000:80"
    depends_on:
      - db

    environment:
      - PHOTOVIEW_DATABASE_DRIVER=mysql
      - PHOTOVIEW_MYSQL_URL=photoview:photosecret@tcp(db)/photoview
      - PHOTOVIEW_LISTEN_IP=photoview
      - PHOTOVIEW_LISTEN_PORT=80
      - PHOTOVIEW_MEDIA_CACHE=/app/cache
+     - GODEBUG="asyncpreemptoff=1"
```

muCruiser:
Thanks for this solution. It works now, but without quotation marks!

`     - GODEBUG=asyncpreemptoff=1`

viktorstrate:
Nice, I think this is the best solution.
It might be worth mentioning it on the documentation website though.

klejejs:
Had the same problem. This helped! Thanks!

PR #365, opened 2021-05-08
To use photoview with samba and cifs share

muCruiser:
Due to a problem with access to cifs / samba share the godebug has to be changed


viktorstrate:
Nice addition thank you, for future reference see also #361

PR #373, opened 2021-05-18
Sanitize float values in EXIF fields

Dhole:
# Issue

After letting photoview scan my photo collection, I found some problematic jpg which were not added to the database.

Here's the log when running with PostgreSQL:
```
photoview_1  | [0.057ms] [rows:0] INSERT INTO "media_exif" ("created_at","updated_at","camera","maker","lens","date_shot","exposure","aperture","iso","focal_length","flash","orientation","exposure_program","gps_la
titude","gps_longitude") VALUES ('2021-05-18 16:22:50.792','2021-05-18 16:22:50.792','GT-S5830','SAMSUNG',NULL,'2013-11-03 15:31:06',+Inf,2.600000,100,3.550000,0,1,3,NULL,NULL) ON CONFLICT DO NOTHING RETURNING "id
"
photoview_1  | 2021/05/18 16:22:50 /app/scanner/exif/exif.go:66 +Inf is not a number
```

I get the same result with MariaDB, and a similar error log.

The affected photo is not added to the database (it's not indexed).

By examining the file with `exiftool` I see that there are two fields with `Inf` value:
```
Exposure Time                   : Inf
Shutter Speed                   : Inf
```

According to MariaDB, the `double` type doesn't support the Infinity value: https://mariadb.com/kb/en/double/
PostgreSQL does support Infinity, -Infinity and Nan: https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-FLOAT Nevertheless this error is present in photoview with PostgreSQL.  Maybe it's related to gorm?

# Proposal

The proposed fix in this PR sanitizes all float EXIF fields so that they are never +Inf, -Inf or Nan, following these rules:
- `+Inf` -> `MaxFloat64`
- `-Inf` -> `-MaxFloat64`
- `NaN` -> `0`

# Notes
I haven't seen any test that performs SQL queries, so I haven't written any test for this issue.

The image that I encountered that triggered this issue was taken by an old phone and was not edited.  I'm attaching an edited example image that shows the issue, for testing purposes.


viktorstrate:
Hi @Dhole thank you for the pull-request and the detailed description.
Do you know what it means for the EXIF values to be infinity? In my mind it doesn't make sense for the `exposure time` or `shutter speed` to be infinity. Would it be better to set them to `nil` to mark them as undefined instead?

I think it would make more sense to sanitize the float values right after they are extracted by exiftool.
This would also make it easier to test as the `ParseExif` function is already covered by tests.
https://github.com/photoview/photoview/blob/1791ef1b5c7c5e16ba86c1260605441730b72e10/api/scanner/exif/exif_parser_external.go#L83-L88

Do you have a sample image that could be added to the exif [`test_data/`](https://github.com/photoview/photoview/tree/master/api/scanner/exif/test_data) directory?

The testing procedure is not documented properly yet. To run the exif related tests run the following command:

```bash
$ go test ./scanner/exif -filesystem -v
```

Dhole:
> Hi @Dhole thank you for the pull-request and the detailed description.
> Do you know what it means for the EXIF values to be infinity? In my mind it doesn't make sense for the `exposure time` or `shutter speed` to be infinity. Would it be better to set them to `nil` to mark them as undefined instead?

I investigated the meaning of an Inf in these two fields and I think you're right, they don't make sense.  I checked the [EXIF spec version 2.2](https://www.exif.org/Exif2-2.PDF), which is the reported version of the offending images and I see that `Exposure Time` is a RATIONAL, which is calculated from two unsigned LONGs (numerator and denominator).  `Shutter Speed` is an SRATIONAL (signed RATIONAL). There's no mention of infinity in this field (infinity is only mentioned as a possible value for `Subject Distance`).  By setting a denominator of 0 in a RATIONAL/SRATIONAL you get infinity, so the value is possible but it's meaning not specified.

For reference, these are the two phones that I have found that contain Inf values in `Exposure  Time` and `Shutter Speed`:
- SAMSUNG GT-I9100 (Software: I9100XWLSW)
- SAMSUNG GT-S5830 (Software: S5830BGKP4)

I think your suggestion of setting the values to `nil` is a better approach.  I'll update the PR.

> I think it would make more sense to sanitize the float values right after they are extracted by exiftool.
> This would also make it easier to test as the `ParseExif` function is already covered by tests.
> 
> https://github.com/photoview/photoview/blob/1791ef1b5c7c5e16ba86c1260605441730b72e10/api/scanner/exif/exif_parser_external.go#L83-L88
> 
> Do you have a sample image that could be added to the exif [`test_data/`](https://github.com/photoview/photoview/tree/master/api/scanner/exif/test_data) directory?
> 
> The testing procedure is not documented properly yet. To run the exif related tests run the following command:
> 
> ```shell
> $ go test ./scanner/exif -filesystem -v
> ```

What you say makes sense!  I'll move the sanitation to `ParseExif` and add a test for it.

Dhole:
@viktorstrate I've updated the PR following your suggestions.  I've also fixed an issue in the EXIF internal parser: the ExposureTime was being parsed as a Float but it's a Rational (I've updated the exif test to validate the Exposure with `./test_data/bird.jpg`)

viktorstrate:
Thank you!

PR #389, opened 2021-06-04
update Kagami/go-face target commit and new deps for it

feinedsquirrel:
go-face only linked blas, which caused issues on systems that don't bundle cblas with blas in the distribution repositories (e.g. Arch Linux). It was updated to explicitly link cblas. This requires Ubuntu systems (and similar distributions) to explicitly install cblas so the new linking doesn't fail. `libatlas-base-dev` is one way to solve this (and I tested it). OpenBLAS may also satisfy it, but I have not tested that.

Referencing the newer go-face commit allows systems such as Arch Linux to install photoview.


viktorstrate:
Thank you for the update.

_A link to the original patch Kagami/go-face#72 for future reference._

PR #412, opened 2021-06-26
Adding translation for Russian language

andreimk:
Taken the English version as a base and then translated it in to Russian.


viktorstrate:
Awesome, thank you!

Issue #431, opened 2021-07-15
Symlinks to directories should be handled like directories, not files


hupfdule:
**Describe the bug**
A symlink to a directory leads to a scanner error as photoview tries to handle it like a file and parse it.
Instead it should be handled like a directory and therefore generate a new subalbum.

**To Reproduce**
Prepare a directory structure with real photos (The amount of albums and photos is not necessary, but demonstrates the purpose of such symlinks) :

```
/photos/album1/photo1.jpg
/photos/album1/photo2.jpg
/photos/album2/photo3.jpg
/photos/album2/photo4.jpg
/photos/album3/photo5.jpg
/photos/album3/photo6.jpg
```

and then add another directory with symlinks to some of those album directories:

```
/photos/hiking/Austria → ../album1/
/photos/hiking/Norway  → ../album3/
```

Then scan the photos.

**Expected behavior**
The symlinks to other directories should be treated like directories and therefore additional albums should be generated.

What actually happens is that they are treated like files and the scanning failes for them.

 - Version [2.3.4]

**Additional context**
Symlinks to files are already handled correctly. They are treated like regular files and scanning succeeds (if they are photos or videos).

PR #432, opened 2021-07-15
Treat symlinks to directories like directories

hupfdule:
This allows symlinking to create additional subalbums.

Closes: #431

One drawback is that the files in the symlinked directories are scanned twice as they have different paths. 
It would be nice if the applications identifies them as identical (e.g. based on an md5sum) and use the already scanned information for both subalbums (the real one and the symlink). But this is not part of this PR.

### DIFF
--- a/api/scanner/scanner_album.go
+++ b/api/scanner/scanner_album.go
@@ -181,7 +181,13 @@ func findMediaForAlbum(album *models.Album, cache *scanner_cache.AlbumScannerCac
 	for _, item := range dirContent {
 		photoPath := path.Join(album.Path, item.Name())
 
-		if !item.IsDir() && cache.IsPathMedia(photoPath) {
+		isDirSymlink, err := utils.IsDirSymlink(photoPath)
+		if err != nil {
+			log.Printf("Cannot detect whether %s is symlink to a directory. Pretending it is not", photoPath)
+			isDirSymlink = false
+		}
+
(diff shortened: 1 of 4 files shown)


viktorstrate:
Looks good, thank you. Would you be up for writing some tests for the `IsDirSymlink` function?
If you don't want to, I'll still gladly merge and add it myself later.

**Test setup:**
- Make a `api/utils/utils_test.go` and add the following
```go
func TestMain(m *testing.M) {
	os.Exit(test_utils.IntegrationTestRun(m))
}
```
- Then for each test within the `api/ui` directory that uses the file system add the line `test_utils.FilesystemTest(t)`. This marks it as an integration test.

Take a look at [scanner_test.go](https://github.com/photoview/photoview/blob/master/api/scanner/scanner_test.go) for inspiration.

hupfdule:
OK, I tried writing a test case for that method.

I am not very used to Go, therefore the test may not be optimal. Feel free to adapt it if necessary.

viktorstrate:
The tests look great, exactly what I imagined!
Although I think you forgot to clean up the `TODO` on line 20?

hupfdule:
> Although I think you forgot to clean up the `TODO` on line 20?

You're absolutely right. Fixed it now.

viktorstrate:
Looks good, thank you!

PR #444, opened 2021-07-21
Patch fix for face images

### LINKED ISSUE #438: Offset in face icons on People page
**Describe the bug**
Some of the preview icons for face groups appear to be offset from the actual detected faces in the new redesign.

**To Reproduce**
Unclear - about 17/20 of the most common faces in my library seem to suffer from this (see screenshot below). Current working theory is that it seems to be related to the orientation of the original photo, primarily affecting landscape photos.
image

**Expected behavior**
As before the redesign.

 - Are you using Docker? **yes**
 - Version: **Docker (edge)**

PJ-Watson:
Patches up the x offset for face circle images on the people page, where the Tailwind base style overrides the existing layout. 

Should fix #438.

Reverting max-width should have no effect on other pages (and doesn't from my testing), since the width is explicitly defined in other cases.

### DIFF
--- a/ui/src/index.css
+++ b/ui/src/index.css
@@ -21,3 +21,9 @@ html {
     transform: scale(1, -1);
   }
 }
+
+@layer base {
+  img {
+    max-width: revert;
+  }
+}


viktorstrate:
Thank you, I think this default is better for this project

PR #450, opened 2021-07-22
chore: Use excplit docker network in docker-compose example.

Sieboldianus:
Minor improvement, as it will prevent any interference with other docker services running on the same host (as happened to me).


viktorstrate:
Doesn't docker-compose automatically isolate each project into its own Docker network?

> https://docs.docker.com/compose/networking/: By default Compose sets up a single network for your app. Each container for a service joins the default network and is both reachable by other containers on that network, and discoverable by them at a hostname identical to the container name.

Sieboldianus:
Interesting - I was not aware of this. Apologies, it appears that I have solved my issue while adding the network syntax, which is obviously wrong given the default-network behavior you linked.

Closing this.

PR #459, opened 2021-07-25
Update docker-compose.example.yml

hakanu:
Add a little bit more details about what to update in docker-compose.yml file.


viktorstrate:
Thank you for the contribution, but I'd rather not have too many comments in the docker-compose file to not make it harder to read. I'd much rather have in depth documentation on the docs website.

The docker section is still not written, but I'd like this kind of information to go there instead.
https://photoview.github.io/docs/installation-docker/

PR #471, opened 2021-08-10
Added Simplified Chinese translation

seanzhang98:


viktorstrate:
There is already a Traditional Chinese translation `zh-HK` which is labeled `繁體中文` in the UI.
Should this be a separate one from that, and if so, what should the label be in the drop down?

I know very little about the different dialects of Chinese :)

seanzhang98:
@viktorstrate 
Hi, just saw your comment. 
Zh-CN should be fine.  "简体中文" can be used in the dropdown menu.

Issue #418, opened 2021-06-29
Folder with both raw and jpegs only displays raw files


ScuttleSE:
**Describe the bug**
I have multiple folders with scanned negatives where I have both a DNG and a JPG of the scanned image. Same filename except the file extension. Photoview only displays the DNG file.

**To Reproduce**
1. Scan an album with both DNG and JPG-files
2. Observe that just the DNG file is displayed

ScuttleSE:
Did some testing and it only seems to affect raw and JPG-files.

I created a folder with two files, a JPG and a PNG. Both shoed up. I added a DNG, it doesn't show up, so I guess it has to do with thumbnail creation for jpeg/raw files? If you have two with the same filename but different extensions, whichever is scanned first gets in.

viktorstrate:
This is an intentional behaviour. This is because some photographers shoot in RAW + JPEG, with this feature they don't see all their pictures duplicated.

ScuttleSE:
I see!

I would like to elaborate a bit on this feature then.

In my case, a lot of the folders contains scanned negatives where the raw file is just that, the raw file, while the jpeg is a copy with the colors inverted.

Right now photoview seems to favor the raw image and creates a thumbnail from that, instead of using the jpeg-file. An option to select which version of an image to select for thumbnails would be appreciated.

Also, there is no way to access the jpeg-copy of the file right now. If I pull up the info on one of the raw images, I get the option of downloading the original raw file, or two versions of that raw file converted into jpeg-images. The existing jpeg is inaccessible.

viktorstrate:
That is definitely a bug. It's supposed to use the jpeg version for the thumbnail and all variants should be downloadable from the sidebar. I guess this feature was introduced before proper unit-testing was in place, and that that bug was introduced without anyone noticing it.

PR #477, opened 2021-08-22
Fix: Correctly display counterpart JPGs for RAWs

tgrondier:
high res thumbnails were generated before counterparts for raw files were searched for. 

This fixes https://github.com/photoview/photoview/issues/418


viktorstrate:
Thank you!

Issue #480, opened 2021-08-29
Graphql dependency seems to be missing in ui/package.json


tmerten:
I just installed photoview manually and found that the frontend did not built until I installed graphql

```
photoview# yarn run build
yarn run v1.22.10
$ NODE_ENV=production node --experimental-modules build.mjs
 > error: [plugin: babel] [BABEL]: Cannot find module 'graphql'
...
```
(the same using npm by the way)

```
photoview# yarn add graphql
...

photoview# yarn run build
yarn run v1.22.10
$ NODE_ENV=production node --experimental-modules build.mjs
esbuild done
workbox done
build complete
Done in 5.05s.
```

As I am not sure if it is intentional and it is a one line fix here is an issue instead of a PR :)

viktorstrate:
Are you running the latestest commit on the master branch?
`NODE_ENV=production node --experimental-modules build.mjs` is an old command for building the project

tmerten:
Hi @viktorstrate . Thanks for the feedback. I did not notice how rapidly you guys pushed something new. I will pull once more and check out the new build script ;).

tmerten:
ups. While closing this and re-building on current master:

```
photoview# npm run build

> photoview-ui@0.0.1 build /usr/local/photoview/ui
> craco build

Creating an optimized production build...

warn - You have enabled the JIT engine which is currently in preview.
warn - Preview features are not covered by semver, may introduce breaking changes, and can change at any time.
Failed to compile.

./node_modules/@apollo/client/core/LocalState.js
Cannot find module: 'graphql'. Make sure this package is installed.

You can install this package by running: yarn add graphql.
```

with this diff the build works for me (tm):

```
diff --git a/ui/package.json b/ui/package.json
index c3f3e58..79a0c9d 100644
--- a/ui/package.json
+++ b/ui/package.json
@@ -32,6 +32,7 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-jest-dom": "^3.9.0",
     "fs-extra": "^10.0.0",
+    "graphql": "^15.5.1",
     "i18next": "^20.3.2",
     "mapbox-gl": "^2.3.1",
     "postcss": "^7.0.36",
```

viktorstrate:
Okay, a PR is welcome :)

PR #505, opened 2021-09-05
Install npm dependencies

benpro:
It is required to run `npm install`.


viktorstrate:
It should not be necessary to run `npm install graphql`, please see #480.

viktorstrate (changes requested):
It should not be necessary to run `npm install graphql`, please see #480.

benpro:
Right, updated.

viktorstrate:
Thank you, looks good :)

PR #514, opened 2021-09-11
add supported systems and systemd files

feinedsquirrel:
Let me know what you think of these, along with the additions to the documentation. If any alterations need to be made, they might need to be made in tandem.


viktorstrate:
Looks good 👍

Issue #89, opened 2020-09-25
Selectable Album thumbnail


FelixUster:
Hey,

thank you very much for this project. It is basically what I am looking for. 
It would be great, if it would be possible to select the album thumbnail, instead of using the earliest Picture in a folder. 
of course, I am willing to help, maybe you can hint me to the respective file, as I have never used JS or Go.

viktorstrate:
Hi @FelixUster 

I think this is would be a nice feature.
It could be implemented in a similar way as how Google Photos does it. Clicking a button labeled "Use as album cover".

If you have very little experience with Javascript and Go, then I think it would be quite ambitious to implement this.
But if you want to give it a shot nevertheless, here are the basic steps:

- Add a [database migration](https://github.com/viktorstrate/photoview/tree/master/api/database/migrations) where a new column is added for a potential cover image, referencing the media_id of the selected cover image.
- Add and implement a graphql mutation to change the cover of an album.
- Update the [album thumbnail resolver](https://github.com/viktorstrate/photoview/blob/bbff25973e0017061ab2b5f43562a9e3825e4d9e/api/graphql/resolvers/album.go#L102-L127) to use the selected album cover if set, otherwise use the current implementation to dynamically select the first one.
- Integrate the graphql mutation into the web interface.

viktorstrate:
Implemented in #526

jiangzhenjerry:
Hi! Sorry for bumping a closed ticket. I do enjoy the "Set as album cover photo" function, so thanks for implementing it. On the other hand, maybe one day this feature could be extended a bit? For example, for the following folder structure:

```
Folder A
├── Folder B
│   ├── 1.jpg
│   └── 2.jpg
└── Folder C
    ├── 3.jpg
    └── 4.jpg
```
We can now easily change the cover photos of Folder **B** and Folder **C**, but not that of Folder **A** because Folder A doesn't have any photos directly under it.

Furthermore, if I change the cover photo of Folder **B** to 2.jpg (from 1.jpg by default), Folder **A** would still use 1.jpg (first photo in the first subfolder) as the cover photo. It would be nice if Folder **A** at least follows my choice of cover photo for Folder B.

EgonHeuson:
This would be an amazing feature!! ❤️ But Photoview is already a very nice job! Thanks!!

hcooper:
+1 controlling the album cover of a parent album would be great.

Lingxi-Li:
[...]

PR #526, opened 2021-09-18
Selectable Album Covers

PJ-Watson:
A solution to #89 , allowing album covers to be selected. 

Back end - as I understand, gorm should handle the auto-migration for a new column, which is just an integer referring to media.id (null value of 0 means no cover has been set). The SQL query for setting a new cover photo is definitely not the most efficient way of writing that, so I'll update that once I'm more familiar with how this part works.

Front end - the layout has been copied from the sharing section, for consistency. Rather than creating an entirely new section, as in the first draft, the cover photo section piggybacks off the sharing section, so that users without sharing permissions cannot change the album cover (and fewer queries need to be sent to the backend).

Known issues - the cover photo selection appears on the sidebar regardless of which page the user is on. Evidently, a slightly different menu should appear for the people pages, and I'll look at implementing that next.


api/graphql/models/album.go
```diff
@@ -17,6 +17,7 @@ type Album struct {
 	Owners   []User `gorm:"many2many:user_albums;constraint:OnDelete:CASCADE;"`
 	Path     string `gorm:"not null"`
 	PathHash string `gorm:"unique"`
+	CoverID  int    `gorm:"index;"`
```
viktorstrate, line 20: Is there a reason why this is indexed?

Also it might be better to have the type be `int*` to indicate that it can be null, and then update all null checks to be `album.CoverID == nil` rather than `album.CoverID == 0`, I think tha's more clear.

api/graphql/resolvers/album.go
```diff
@@ -148,24 +150,44 @@ func (r *albumResolver) Thumbnail(ctx context.Context, obj *models.Album) (*mode
...
+	// 		SELECT child.* FROM albums AS child JOIN sub_albums ON child.parent_album_id = sub_albums.id
+	// 	)
+	//
+	// 	SELECT * FROM media WHERE media.album_id IN (
+	// 		SELECT id FROM sub_albums
+	// 	) AND media.id IN (
+	// 		SELECT media_id FROM media_urls WHERE media_urls.media_id = media.id
+	// 	) LIMIT 1
+	// `, obj.ID).Find(&media).Error
+
```
viktorstrate, line 190: This should probably be deleted

api/graphql/schema.graphql
```diff
@@ -273,6 +278,8 @@ type Album {
   path: [Album!]!
 
   shares: [ShareToken!]!
+
+  coverID: Int!
```
viktorstrate, line 282: Isn't this redundant, as it can be derived from the `thumbnail: Media` attribute?

viktorstrate (changes requested):
First of all nice work and thank you for the contributions!

Most of it looks solid, but I've left a few comments on things that maybe should be changed.

One think that I'd really like is some unit tests, especially on the API code, if you are unsure on how to go about that, please ask and I'll point you in the right direction.

---

> The SQL query for setting a new cover photo is definitely not the most efficient way of writing that, so I'll update that once I'm more familiar with how this part works.

I'm not sure why that's not efficient? Can you please help me understand :blush:

> Front end - the layout has been copied from the sharing section, for consistency. Rather than creating an entirely new section, as in the first draft, the cover photo section piggybacks off the sharing section, so that users without sharing permissions cannot change the album cover (and fewer queries need to be sent to the backend).

I think that's perfect, I imagine that the UI will be modified a bit at some point, maybe a options dropdown will be added.

For now I think it would be nice to get an indication that it worked, when the button is pressed now it doesn't look like anything happens. Maybe just disable the button after it has been pressed and make the text a little transparent.

> Known issues - the cover photo selection appears on the sidebar regardless of which page the user is on. Evidently, a slightly different menu should appear for the people pages, and I'll look at implementing that next.

I think that's ok, especially if the button is moved into an options dropdown in the future.

api/graphql/schema.graphql
PJ-Watson: I did initially think the same, but graphql throws up several errors if it can't find cover_id in the schema - left it in for now, will try and figure out why later.

api/graphql/resolvers/album.go
PJ-Watson: Good catch - I thought I'd removed all the commented out code, but apparently not!

api/graphql/models/album.go
PJ-Watson: No particular reason, I just copied some of the previous code without thinking! Explicit null checks are a good idea, agreed on that.

PJ-Watson:
> First of all nice work and thank you for the contributions!

Thanks for writing the rest of it - much easier to tinker around the edges than starting from scratch!

> One think that I'd really like is some unit tests, especially on the API code, if you are unsure on how to go about that, please ask and I'll point you in the right direction.

I've had a look around, I think I can figure out what's going on - I'll try and get those sorted out tomorrow.

> > The SQL query for setting a new cover photo is definitely not the most efficient way of writing that, so I'll update that once I'm more familiar with how this part works.
> 
> I'm not sure why that's not efficient? Can you please help me understand blush

So there are two separate resolvers at the moment: one with the album.ID, which resets the cover_id field to null, and one with the media.ID, which sets the cover_id field of the related album with itself. It's the second one I had problems with - currently, it finds the media from the ID in the mutation, then fetches the album, then updates cover_id. There should be a more efficient way of doing this than three SQL queries (possibly in a single line with gorm?), but I couldn't figure it out when I was writing the code.

> I think that's perfect, I imagine that the UI will be modified a bit at some point, maybe a options dropdown will be added.
> 
> For now I think it would be nice to get an indication that it worked, when the button is pressed now it doesn't look like anything happens. Maybe just disable the button after it has been pressed and make the text a little transparent.

Good idea - all sorted now. :)

> > Known issues - the cover photo selection appears on the sidebar regardless of which page the user is on. Evidently, a slightly different menu should appear for the people pages, and I'll look at implementing that next.
> 
> I think that's ok, especially if the button is moved into an options dropdown in the future.

I wonder if it might make more sense to have a slightly different setup for the facegroup cover photos, but I'll focus on finishing off this bit first!

viktorstrate:
To test the API, I suggest you only test the actual database code and not the graphql resolver it self.
To do this take a look at [api/graphql/models/actions](https://github.com/photoview/photoview/tree/master/api/graphql/models/actions) and the corresponding tests.

Regarding environments, then when testing the API, it looks for database environment variables from the `testing.env` file, rather than the normal `.env` file. This is to make it easier to separate the test and dev database.

PJ-Watson:
Bizarre - the thumbnail resolver fails on the github actions, but not when I run the tests locally. I'll have a look at why that might be happening later, but at least the tests for setting/resetting the cover photo seem to be working now.

viktorstrate:
It's beginning to look good, thank you for putting in the effort.
The reason why the project coverage is so low is simply because you added `resolvers_test.go` which means a lot of new files a now included in the unit tests, you can ignore that.

Let me know if/when you are done, and I'll take a closer look :)

PJ-Watson:
> It's beginning to look good, thank you for putting in the effort.

Thank you! I thought it was worth putting in a bit more time to understand how it all works - will make it a lot easier to write tests for other resolvers in future (e.g. a potential facegroup.cover_photo).

> The reason why the project coverage is so low is simply because you added `resolvers_test.go` which means a lot of new files a now included in the unit tests, you can ignore that.

Ah, that would make sense - I was wondering why it dropped so much. Unfortunately I think `resolvers_test.go' is necessary to stop cyclic imports.

> Let me know if/when you are done, and I'll take a closer look :)

I think it's pretty much done for now (at least on the API side). The tests check if the thumbnails are resolving correctly in the null case, or when it's already been set, then it'll test whether Album.CoverID can be set/reset correctly. 

The only known problem at the moment is refreshing the album thumbnails on update. On the album page, I can set the fetch policy of the query to `fetchPolicy:"cache-and-network"`, but this seems to results in an infinite loop of network requests. I've set it to `"network-only"` instead, which means the query will be re-run rather than cached on page re-render, but I think the performance hit should be minimal if the thumbnails are already cached separately. There's quite a few Github issues about infinite loops on useQuery, so hopefully one of those will have a solution.

viktorstrate (approved):
I have made some restructuring of the album resolvers as well as the tests you wrote. I haven't changed what they test, only the structure.

I also fixed the problem where the thumbnails wasn't updating properly, the problem was that the response from the mutations didn't include the id of the new thumbnail so Apollo didn't know to update it's internal cache.

I also removed `coverID` from the graphql schema, now it's only an internal value.

I now approve this PR and will happily merge it if you also think it's fine. Thank you for the rather big contribution, and I hope this wasn't your last :blush:

PJ-Watson:
> I have made some restructuring of the album resolvers as well as the tests you wrote. I haven't changed what they test, only the structure.
> 
Having had a quick look through the changes, thanks for doing this! I haven't used Go before (coming from a Python background), so it's very helpful to see how things can be laid out better.

> I also fixed the problem where the thumbnails wasn't updating properly, the problem was that the response from the mutations didn't include the id of the new thumbnail so Apollo didn't know to update it's internal cache.
> 
> I also removed `coverID` from the graphql schema, now it's only an internal value.
> 
That would make a lot of sense; I'm glad you managed to fix it.

> I now approve this PR and will happily merge it if you also think it's fine. Thank you for the rather big contribution, and I hope this wasn't your last 😊
>
I'm definitely happy with that! This issue looked like low-hanging fruit compared to some of the others (given you'd already laid out the steps needed to solve it), so it seemed a good place to start. I'll definitely be looking at adding a similar system for faces next, since the underlying logic should be pretty similar.

viktorstrate:
Awesome 🎉

> This issue looked like low-hanging fruit compared to some of the others (given you'd already laid out the steps needed to solve it), so it seemed a good place to start.

Maybe I should start doing that more then :)

PR #534, opened 2021-09-26
Fix bug in new install.

PJ-Watson:
In a new setup with no photos in the database, the default timeline will throw a TypeError - `TimelineFilters.tsx` doesn't check if the `myMedia` array is empty (but it does still need the null check, otherwise that object will be undefined).

(Only one line added, I didn't think it was worth opening a new issue!)


viktorstrate:
Thank you. When you thought you had tested everything :)

PR #541, opened 2021-09-28
add a note that the mapbox token must be a public one

DjSni:


viktorstrate:
If this has to be more specific, I think `access token` is more precise and less confusing than `public token`.
And technically I think a `temporary token` can still be used, even though it would probably not make sense.

It might be better to add a section to the documentation that describes how to generate a token and what scopes that should be enabled.

PR #552, opened 2021-09-30
Update Dockerfile

elvece:
Persist `curl` as a dependency, mainly for upstream service availability health checks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -98,6 +98,5 @@ ENV PHOTOVIEW_UI_PATH /ui
 
 EXPOSE 80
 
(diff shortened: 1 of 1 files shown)


viktorstrate:
Thank you, I don't see why we shouldn't integrate the healthcheck directly into the docker image as well.
This could be done with the [`HEALTHCHECK` instruction](https://docs.docker.com/engine/reference/builder/#healthcheck) inside of the Dockerfile.


I think this could be a good candidate for the health check curl command, obviously the url should be changed:

```bash
curl --fail 'https://photos.qpqp.dk/api/graphql' -X POST -H 'Content-Type: application/json' --data-raw '{"operationName":"CheckInitialSetup","variables":{},"query":"query CheckInitialSetup { siteInfo { initialSetup }}"}'
```

What do you think?

elvece:
Sure, that seems like a good thing to have in general. Will add.

Issue #255, opened 2021-03-17
Server Error: internal system error at (recognizeUnlabeledFaces)


ced455:
**Describe the bug**
Server Error: internal system error at (recognizeUnlabeledFaces)

**To Reproduce**
When Hitting the button "Recognize unlabeled faces"

**Expected behavior**
unlabeled faces faces to get recognized ?

**Screenshots**
![image](https://user-images.githubusercontent.com/26870023/111444741-9b4aff80-8702-11eb-832d-d5bb39887fff.png)

**Your environment:**
Docker Latest 

Thank you again for this app !

Manfred73:
Yesterday I installed latest docker version: 2.3.13 ([446ef9a4].
I'm also experiencing this error when click the button.

diyoyo:
Hi, it seems like pull request 577 has never been merged with main, right?
How to fix this issue then?
PS : I'm using the latest docker container.

PR #566, opened 2021-10-10
Face testing

### LINKED ISSUE #551: JS Error on action "Move image faces"
**Describe the bug**

I want to move some faces from one group to another. For this I open a face from the "people" page. Then click on "Moves faces" and after selecting an image and hitting "Next" and "Move", the browser prints this error:

```
Uncaught (in promise) Error: Paginate argument is missing for query: imageFaces
```

The UI doesn't change and it's only possible to cancel.

**To Reproduce**
1. Go to *people page* and select a face 
2. Click on 'Moves faces' action button
3. Select one or more images, click on 'Next'
4. Open the browser console
4. Choose a destination face group and click "Move image faces"
5. See the error in the console

**Expected behavior**

I would expect the action to work so that the selected images are moved from the current group to the target group.

**Your environment:**

 - Are you using Docker? no
 - Database: PostgreSQL 13
 - Browser: Firefox 91.0.2 
 - Version: master, `e05c4bef822a68cc1d2ac9b6902a919e39b9242b`

PJ-Watson:
This closes #551, along with #255. Definitely a work in progress - no tests yet, and not all functionality added. 

#255 - clusters unrecognised faces, then attempt to match them. Although this is a significant departure from the current setup of scanning for matches on a face-by-face basis, it avoids incorrect matches accumulating as part of recognised faces. 

Will explain the changes once I've finished ironing out all the bugs!

### DIFF
--- a/ui/src/Pages/PeoplePage/SingleFaceGroup/DetachImageFacesModal.tsx
+++ b/ui/src/Pages/PeoplePage/SingleFaceGroup/DetachImageFacesModal.tsx
@@ -53,6 +53,10 @@ const DetachImageFacesModal = ({
     refetchQueries: [
       {
         query: MY_FACES_QUERY,
+        variables: {
+          limit: 50,
+          offset: 0,
+        },
       },
     ],
   })
--- a/ui/src/Pages/PeoplePage/SingleFaceGroup/MergeFaceGroupsModal.tsx
+++ b/ui/src/Pages/PeoplePage/SingleFaceGroup/MergeFaceGroupsModal.tsx
@@ -41,11 +41,17 @@ const MergeFaceGroupsModal = ({
 }: MergeFaceGroupsModalProps) => {
   const { t } = useTranslation()
 
-  const [selectedFaceGroup, setSelectedFaceGroup] =
-    useState<myFaces_myFaceGroups | singleFaceGroup_faceGroup | null>(null)
+  const [selectedFaceGroup, setSelectedFaceGroup] = useState<
+    myFaces_myFaceGroups | singleFaceGroup_faceGroup | null
+  >(null)
 
   const history = useHistory()
-  const { data } = useQuery<myFaces, myFacesVariables>(MY_FACES_QUERY)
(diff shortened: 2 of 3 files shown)


PJ-Watson:
> Awesome looks great after a quick skim of the changes. I really like the idea of using DBScan to cluster the faces first.
> 
Thank you! I'm not sure how easy it would be to implement a similar system for when faces are first scanned in (e.g. per album), but I think that's something to consider at a later date.
> I noticed that you have forked two Go projects ([`goface`](https://github.com/PJ-Watson/go-face) and [`clusters`](https://github.com/PJ-Watson/clusters)) and used your version. Are you planning to have your changes merged into the original projects, or is this just for the testing/development phase?
> 
For `go-face`, I'll see if the change can be merged into the main repository - otherwise it'll be a fairly simple solution without the change. 

`clusters` doesn't appear to be in active development, so I think I'll stick with my fork - there's at least one bug in the main branch which I hope I've fixed!

> Do you think you'll be changing the the way face groups are managed (move, merge, detach, etc.). Because I am working on adding faces to the sidebar, and if you want to make changes, I'd rather wait until this is merged to implement that.

Generally no - although I might like to change the way this currently works, I don't think I'll have time to get round to it at the moment. The main thing I'm trying to add (inspired by favourites in photos) is the ability to hide faces - although it would be good to see those faces directly in the sidebar anyway, so I don't think there will be any conflicts!

PJ-Watson:
Just popping back in to say I haven't completely abandoned this - unfortunately work and other commitments have put this on the back seat for now (and most likely until the end of the month unfortunately :( ). I'll see if I can tidy up the code I have already, a smaller commit to solve the two linked issues might be more manageable than trying to fix everything in one go!

viktorstrate:
Thank you for letting me know. That's completeley fine, take the time you need.
I agree, let's keep this PR as short and managable as possible.

As the face recognition part of the API has barely any testing, I'd really like some more tests on that moving forward. I think that it will be very natural to add simultaneously with fixing #255 - to verify that the fix(es) do indeed work.

PR #656, opened 2022-03-05
Fix slow query in getting media of a face by using another join

Phlogi:
The subquery returns all potential media id's. Typically, we have a couple of faces and thousands of media files. 
A join is much faster. 

With about 50k images and a face that was present in 2'000 images the query was unusable slow, it took about 60s. 
This is a very important performance fix, as I think many users will run into it. 
Maybe there are other areas where the same improvement is possible, I didn't check.


viktorstrate:
Thank you!

PR #664, opened 2022-03-23
fix switching page causes setting to crash

Goppai:
Setting crashes when switching between people and setting pages multiple time
WeChat8c36c2c4c389dff4fd32cd159327e95a
s

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -59,7 +59,6 @@
   "scripts": {
     "start": "BROWSER=none PORT=1234 craco start",
     "build": "craco build",
-    "eject": "react-scripts eject",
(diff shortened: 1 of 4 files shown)


Goppai:
The variables are created when their containing Lexical Environment is instantiated but may not be accessed inany way until the variable’s LexicalBinding is evaluated. 
[https://262.ecma-international.org/6.0/#sec-let-and-const-declarations](url)
may be caused by this

viktorstrate:
Thank you, looks great. I'm sorry that I haven't been able to look at this yet. Hope to get to it soon.

PR #730, opened 2022-08-11
systemd run with opt deps

feinedsquirrel:
`systemd` needs to have a `PATH` variable defined to find and run other executables as dependencies. Since we already have an environment file, and that environment file is read in the unit file (`photoview.service`), the simplest solution is to add `PATH` in the `.env` file.

I don't believe it will get in the way of the rest of the program that reads other env vars from that file.

I've tested on Arch. Docker, Ubuntu, etc. do still need to be tested.

### DIFF
--- a/api/example.env
+++ b/api/example.env
@@ -30,3 +30,6 @@ PHOTOVIEW_SERVE_UI=0
 # Set to 1 to set server in development mode, this enables graphql playground
 # Remove this if running in production
 PHOTOVIEW_DEVELOPMENT_MODE=1
+
+# For the systemd-unit to be able to find optional dependencies
+PATH=/usr/bin:/usr/bin/vendor_perl


viktorstrate:
The `example.env` file(s) are mostly intended as a template for developers, this addition might therefore be a little confusing for developers. Also it it quite system-specific. For example this would be different on mac.

Since `systemd` setups are meant for production, I think it would be better to have a separate `.env` file dedicated to that in the `systemd` directiory. I'm not very knowledgeable in this area, but would this be possible and a good solution for the problem?

feinedsquirrel:
related to #787
I have been smashed a fair bit over the past 7 months. I think adding a line to the systemd unit file would be a plausible solution, as mentioned in a comment by a user on the AUR package page:
https://aur.archlinux.org/packages/photoview#comment-903924
Though we need to discuss what upstream wants here because that user has `sbin` on his PATH, but Arch Linux doesn't typically use `<>/sbin/<>` anything, everything should be installed to e.g. `/usr/bin` and `<>/sbin/<>` symlinks to `<>/bin/<>` for backwards compatibility.

Since there has been more interaction on this, I'll try to bump it up on my personal queue. I'll try to test moving path to the systemd unit file myself, and see what I can put together.

I also will open a new pull request once I have an idea for a solution for the other problem raised in the AUR package comments.

Issue #236, opened 2021-03-03
Adjustable large picture dimensions or smth like 50% resized format


ivkiwi:
**Is your feature request related to a problem? Please describe.**
As of now, large jpegs from raw are generated same size as the original raw files.
In my personal case 42mpix ( and i'm not even talking about 61mpix raws of a7rIV as they're unprocessable until darktable is updated to a more recent version) are a little bit too much for just a fullsize preview.
<img width="446" alt="Screenshot 2021-03-03 at 14 48 48" src="https://user-images.githubusercontent.com/69342745/109801493-9892d980-7c2f-11eb-8cc5-584093975f0a.png">

**Describe the solution you'd like**
Add a 50% resize option or adjustable setting in admin area.

viktorstrate:
Thank you for the request, I totally agree except that if a zoom feature (see #217) was implemented, it might be desirable to be able to zoom in to 100% to check for sharpness?

I would like to hear your opinion on that, I haven't worked with a DSLR with that high of a resolution. Is it desirable to being able to zoom right in to 100% without having to download the image first?

ivkiwi:
Wow, thanks for such a quick response!
From my point of view the best way is to have an adjustable setting like "Large preview size: 100%/50%/etc" with simple input field with percentage setting.
Even with zoom feature fullsize unedited raw previews are not the best option imho, but with this setting anyone will be able to choose what he needs to obtain in result of preview generation.

brandonkal:
Just offering another UX opinion:
The proposed variable percentage option does not make sense. 50% on your Sony may be desired but for your current Sony but would you want that for your 12 MP phone? People will be endlessly tweaking and it won't work out as most shoot with multiple cameras.

It would make more sense to offer a ladder. Define a size for Large (say for example 4000px long). Not as a user-configured knob.
Then offer:
Small
Medium
Large (also preview)
Original Size

Collapse the last two if the original is smaller than the large size.

Issue #217, opened 2021-02-26
Zoom Feature


el-joseppe:
Mainly due to sharpness check I really miss a simple feature to switch between at least 2 zoom states: "fit to screen" and "100%" or a free zooming tool plus drag-ability. 

In the existing fullscreen-mode scrolling and dragging with the mouse has no function and could therefore be used to zoom in and out and drag the zoomed-in section around. 

Alternatively i also would like or even prefer a keyboard-shortcut to switch between "fit to screen" and "100%". 

Thank you!

viktorstrate:
I think this would be a nice addition, I often use the 100% zoom function in Lightroom for my DSLR photos, to check for sharpness as well.

metbosch:
Maybe using the https://github.com/prc5/react-zoom-pan-pinch/ module?

hgpuke:
This is currently my most wanted feature.

resuser:
Adding support for this feature. I have two main reasons for wanting this:

1. Bandwidth: It would be ideal for the image loaded in full screen match or only slightly exceed the resolution of the screen. When zooming, JS would dynamically load the full size (or a specified maximum size). Photoview already does this when flipping through images, with all the thumbnails already loaded and the full size image loaded after a slight delay. I'm proposing loading a screen resolution size first, then full size on zoom.
2.

PR #731, opened 2022-08-12
Extend swipe functionality

PJ-Watson:
On its own, this is a very simple addition to the swipe functionality on touch screen compatible devices, allowing for a swipe down to close the main image view. However, it did leave me wondering whether it'd be possible to extend the functionality of the image viewer itself.

There's a couple of open issues regarding zooming in on the single photo viewer (#217 and #236 ) - if https://github.com/FormidableLabs/react-swipeable/issues/244 is to be believed, the swipe overlay on the media can easily handle a double-click/tap to zoom event. The main problem then is figuring out a simple way to actually show the media at 100% size, and allow the user to navigate around whilst zoomed in, without the existing overlay (buttons and swipe gestures) getting in the way. 

[On a separate note, I do intend to implement a test for the react-swipeable component - identifying the element isn't a problem, but testing-library/react doesn't natively support a swipe gesture. I'll figure it out before adding any more swipe functionality though]

### DIFF
--- a/ui/src/components/photoGallery/presentView/PresentNavigationOverlay.tsx
+++ b/ui/src/components/photoGallery/presentView/PresentNavigationOverlay.tsx
@@ -96,10 +96,17 @@ const PresentNavigationOverlay = ({
   }, [])
 
   const handlers = useSwipeable({
+    onSwipedDown: () => {
+      if (disableSaveCloseInHistory === true) {
+        dispatchMedia({ type: 'closePresentMode' })
+      } else {
+        closePresentModeAction({ dispatchMedia })
+      }
+    },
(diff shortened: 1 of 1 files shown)


viktorstrate:
> The main problem then is figuring out a simple way to actually show the media at 100% size, and allow the user to navigate around whilst zoomed in, without the existing overlay (buttons and swipe gestures) getting in the way.

For 100% zoom I think the solution would be to make the `` tag positioned `absolute` with its width/height set to the actual pixel size of the image.

I am very open for larger changes to the full screen view if needed for zoom or other new features.
#217 is a feature that would be really nice to add, it would really fit the project goal of being aimed at photographers.

PR #752, opened 2022-10-06
Restyle titles and spacing in timeline gallery

orellazri:
Restyle the titles and album names to look a bit better, and decrease the horizontal spacing between photos in the same album.

Before:
Before

After:
After


stephenp:
🎉

PR #773, opened 2022-11-13
add "copy direct link" btn 

### LINKED ISSUE #745: Sharing a direct link to a picture
Sometimes I want to use direct links to picture. For example when I am publishing a blog post, I want to be able to share the picture and point to it directly. Right now the sharable link point to the picture inside the photoview UI.

We could add, on the public link list, a second "copy link" button that copies the direct link instead of the normal link.

It is possible to get the direct link by opening the shared picture, right-clicking on it and doing a "Copy image link", but that's not really an ideal flow for that.

**Additional context**
I don't think it's too big of a change, if you're interested in that feature I could open a PR

noel123007:
Closes #745

Added a "copy direct link" button which copies the direct link to the image instead of photoview link

### DIFF
--- a/api/scanner/scanner_album.go
+++ b/api/scanner/scanner_album.go
@@ -103,7 +103,7 @@ func ScanAlbum(ctx scanner_task.TaskContext) error {
 		mediaData := media_encoding.NewEncodeMediaData(media)
 
 		if err := scanMedia(ctx, media, &mediaData, i, len(albumMedia)); err != nil {
-			return errors.Wrap(err, "album scan")
+			scanner_utils.ScannerError("Error scanning media for album (%d) file (%s): %s\n", ctx.GetAlbum().ID, media.Path, err)
 		}
 	}
 


viktorstrate:
Sorry for the very long delay.

I just looked at it and I'm wondering if the "Copy direct link" and "Copy link" are too similar.
Maybe it would be better to have the "Copy direct link" as a button on the "shared media page", the page you visit when you copy the link?

Like shown on the screenshot here:

Screenshot 2022-11-27 at 21 24 52

kkovaletp:
@jordy2254, I've tested this and see that the new button copies only the unique part of the URL, missing the base URL. 
Also, there is a question from @viktorstrate about placing the button somewhere else, however, I think that it should be in the current location, as the copied direct link works only if the shared link is created and not password-protected, so having this button somewhere else will create much more confusion. It is better to rename the "Copy Link" to something like "Copy Link to Page".
BTW, isn't the code looking like a hack here: the new button definition looks different than the existing buttons in that block.

What do you think about this PR?

PR #790, opened 2023-01-17
Fix #789: Set the map language to UI language

### LINKED ISSUE #789: Please supports Non-English map in Place
Now the map in Place is always in English, and it isn't so good for the user who doesn't speak English. For example, me :-).
Since supporting Non-English website language, I think Non-English map should be provided too.

I think manual or automatic configuring the map language is a good idea.

- Manual configuring the map language
Configuring the map language in Setting, just like configuring the website language. 

- Automatic configuring the map language
Set the map language to the same one of the website language in Setting.

I'm not so familiar to the Mapbox API. But according to some pages found on Internet, Mapbox provides the API to set the language of map. 

None.

**Additional context**
None.

WindLi001:
This is a solution of  #789. The language of map in Place will be set to the same one of the UI. If the UI language is NOT supported by Mapbox, then the English map will be provided.

### DIFF
--- a/api/scanner/media_encoding/media_utils/photo_dimensions.go
+++ b/api/scanner/media_encoding/media_utils/photo_dimensions.go
@@ -48,6 +48,11 @@ func (dimensions *PhotoDimensions) ThumbnailScale() PhotoDimensions {
 		height = 1024
 	}
 
+	if width > dimensions.Width {
+		width = dimensions.Width
+		height = dimensions.Height
+	}
+
 	return PhotoDimensions{
 		Width:  width,
 		Height: height,
--- a/ui/src/components/mapbox/MapboxMap.tsx
+++ b/ui/src/components/mapbox/MapboxMap.tsx
@@ -6,6 +6,7 @@ import styled from 'styled-components'
 import 'mapbox-gl/dist/mapbox-gl.css'
 import { mapboxToken } from './__generated__/mapboxToken'
 import { isDarkMode } from '../../theme'
+import { SetMapLanguages } from '../../localization'
 
 const MAPBOX_TOKEN_QUERY = gql`
   query mapboxToken {
@@ -59,12 +60,16 @@ const useMapboxMap = ({
       mapboxLibrary.accessToken = mapboxData.mapboxToken
 
     map.current = new mapboxLibrary.Map({
+      zoom: 6, // zoom must be initialized to 6 or more, or the language of map cannot be set.
+      center: [-77.0259, 38.901], // Initial to the land.
       container: mapContainer.current,
       style: isDarkMode()
         ? 'mapbox://styles/mapbox/dark-v10'
         : 'mapbox://styles/mapbox/streets-v11',
       ...mapboxOptions,
     })
+
+    SetMapLanguages(map.current)
 
     configureMapbox(map.current, mapboxLibrary)
     map.current?.resize()
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -12,6 +12,7 @@
         "@apollo/client": "^3.6.9",
         "@babel/preset-typescript": "^7.18.6",
         "@headlessui/react": "^1.6.6",
+        "@mapbox/mapbox-gl-language": "^1.0.0",
         "@types/geojson": "^7946.0.8",
         "@types/jest": "^28.1.4",
         "@types/mapbox-gl": "^2.7.3",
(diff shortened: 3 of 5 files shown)


WindLi001:
Adding another bug fix. Now photoview generating 1024px thumbnail for each photo, including the photo smaller than 1024px. 
I add a limit. If the original photo is smaller than 1024x, then the thumbnail will be the same size of the original photo.

ui/src/components/mapbox/MapboxMap.tsx
```diff
@@ -59,13 +60,17 @@ const useMapboxMap = ({
       mapboxLibrary.accessToken = mapboxData.mapboxToken
 
     map.current = new mapboxLibrary.Map({
+      zoom: 6, // zoom must be initialized to 6 or more, or the language of map cannot be set.
+      center: [-77.0259, 38.901], // Initial to the land.
```
viktorstrate, line 64: Why did you hard-code these coordinates in particular?

api/scanner/media_encoding/media_utils/photo_dimensions.go
```diff
@@ -48,6 +48,11 @@ func (dimensions *PhotoDimensions) ThumbnailScale() PhotoDimensions {
 		height = 1024
 	}
 
+	if width > dimensions.Width {
+		width = dimensions.Width
+		height = dimensions.Height
+	}
```
viktorstrate, line 54: I'd appreciate if you send another PR for this change.
Should it also consider if the height is smaller?

viktorstrate (changes requested):
Thank you for your pull request, it would be a nice addition.
I've left som small comments with some change requests.

I'm very sorry that I didn't had the time to look at this before now

ui/src/components/mapbox/MapboxMap.tsx
WindLi001: I set the map center, because if it isn't set, the center will be in the ocean by default. Then at zoom 6, the clients will see nothing on the map, and maybe they will think the map is broken. The best way is choosing the coordinate of the client's nation by default, but I don't find the way to get this information. So I have to hard-code a random coordinate in the land.

api/scanner/media_encoding/media_utils/photo_dimensions.go
WindLi001: Because the aspect of thumbnail is the same as photo, comparing width or height is enough. I choose comparing width.

ui/src/components/mapbox/MapboxMap.tsx
viktorstrate: Is it maybe appropriate to use the default [0, 0] coordinate and then have the default zoom be zoomed out to the max value so the entire world is visible and the user can zoom to their desired place?

WindLi001: I set zoom to 6 because of a bug of mapbox-gl-language. [https://github.com/mapbox/mapbox-gl-language/issues/60](url)
In fact, at beginning I don't set either the zoom or the center, but just only call MapboxLanguage({defaultLanguage: xxx}). Then I found the map language is still English. I check the codes several times but have no idea. Fortunately, at last I found the BUG of mapbox-gl-language and raise the issue above. But until now, this bug is still here.
So now we have 2 choices:
1. Waiting mapbox-gl-language fix this bug, and don't set either the zoom or the center. But before they fix it, the map language will always be English.
2. Setting the zoom and center to provide multi-language map, waiting them fix this bug, and then delete the setting of zoom and center.
I want to fix this bug by myself, but I'm still learning JavaScript now.

WindLi001: With the beginner’s luck, I found another way can change the map language without setting zoom to more than 6. If delete the setting of map projection in ui\src\Pages\PlacesPage\PlacesPage.tsx like this,
```
  const { mapContainer, mapboxMap, mapboxToken } = useMapboxMap({
    configureMapbox: configureMapbox({ mapboxData, dispatchMarkerMedia }),
    mapboxOptions: {
      //projection: {
      //  name: 'globe',
      //},
    },
  })
```
the map language can be changed too.
I still don't know why the map language is related to the zoom or projection. And I have a look to the code of mapbox-gl-language but found nothing about zoom or projection. 
I think may the map projection is important to the plane and ship, but not for a photo gallery :-). So, if you agree, I'll revert this commit, delete the setting of zoom and center, delete the setting of projection, and commit again.

api/scanner/media_encoding/media_utils/photo_dimensions.go
WindLi001: I move this commit to a new pull request #798, and this one will be closed.

ui/src/components/mapbox/MapboxMap.tsx
WindLi001: I move this commit to a new pull request #797, and this one will be closed.

WindLi001:
This pull request is created directly on my master branch, which is absolutely wrong. So, I create new pull requests #797 and #798 on other branch, and this one will be closed.

PR #795, opened 2023-02-05
fix transaction already commited error

Ajnasz:
I encountered with the following error:

> 2023/02/05 07:33:00 /app/scanner/face_detection/face_detector.go:92 sql: transaction has already been committed or rolled back
> [0.042ms] [rows:0] SELECT * FROM `media` WHERE `media`.`id` = 823 ORDER BY `media`.`id` LIMIT 1
> 2023/02/05 07:33:00 ERROR: Error detecting faces in image (/photos/Borzsony2017/DSC_0028.NEF): sql: transaction has already been committed or rolled back

It turned out it comes from the api/routes/photos.go

I found a very similar code in album_scanner.go.
The difference I saw was that while in the single photo request the
transaction passed to the `scanner_tasks.Tasks.BeforeProcessMedia` call,
in the album_scann.go the transaction created after this call and
created from the context which returned by `BeforeProcessMedia`.

Another difference was that in the `ProcessSingleMedia` call the
`AfterProcessMedia` call was called with the same - db transaction -
context, in the album_scanner it was called outside of the transaction.

I changed the logic by merging the two behavior:
Create the transaction from the context of `BeforeProcessMedia` and also
use the transaction context in the `AfterProcessMedia`.
After the change the error disappeared.

So to have it in a common place I extracted that logic into a function
and use for both the single photo request and in the album scanner.

I did not go more deeper to find out what's going on with the context
under the hood.

### DIFF
--- a/api/routes/photos.go
+++ b/api/routes/photos.go
@@ -45,21 +45,16 @@ func RegisterPhotoRoutes(db *gorm.DB, router *mux.Router) {
 		}
 
 		if _, err := os.Stat(cachedPath); os.IsNotExist((err)) {
-			err := db.Transaction(func(tx *gorm.DB) error {
-				if err = scanner.ProcessSingleMedia(tx, media); err != nil {
-					log.Printf("ERROR: processing image not found in cache (%s): %s\n", cachedPath, err)
-					return err
-				}
+			// err := db.Transaction(func(tx *gorm.DB) error {
+			if err = scanner.ProcessSingleMedia(db, media); err != nil {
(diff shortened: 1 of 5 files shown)


viktorstrate:
You can add an `upstream` remote and create a branch from the upstream master branch.
See [Working with Forks](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/configuring-a-remote-repository-for-a-fork).

```
$ git checkout upstream/master
$ git checkout -b new-feature
```

viktorstrate:
Thank you, looks good :)

Issue #672, opened 2022-04-09
Change Mapbox language to user selected locale


log-time:
Make Mapbox use that selected language rather than always English.

Mapbox has a guide on this here:
https://docs.mapbox.com/mapbox-gl-js/example/language-switch/

IamSAL:
I want to take on this issue!

kkovaletp:
@log-time, thanks for the proposal. 

I'll close this issue, as we have the #789, which has a bit better description and a PR linked.

PR #797, opened 2023-02-10
Set the map language to UI language in new branch

WindLi001:
This is a solution of https://github.com/photoview/photoview/issues/789. The language of map in Place will be set to the same one of the UI. If the UI language is NOT supported by Mapbox, then the English map will be provided.
This is a new pull request of Set the map language to UI language, and I commit it to a new branch other than master. The old one will be closed soon.
As the comment in the old pull request #790, I delete the setting of zoom and center, and delete the setting of projection, which make the map language cannot be changed. 
I have a look to the commit setting the projection, it said "chore: add mapbox globe view", so I guess the committer just want to show the mapbox with whole world map initially. But I think maybe he or she don't know what the projection really means, in fact global projection just make a spherical map. So, I add zoom: 1 to show the mapbox with whole world map initially.

### DIFF
--- a/ui/src/localization.ts
+++ b/ui/src/localization.ts
@@ -6,6 +6,8 @@ import { initReactI18next, TFunction } from 'react-i18next'
 import { LanguageTranslation } from './__generated__/globalTypes'
 import { authToken } from './helpers/authentication'
 import { exhaustiveCheck, isNil } from './helpers/utils'
+import type mapboxgl from 'mapbox-gl'
+import MapboxLanguage from '@mapbox/mapbox-gl-language'
 
 export type TranslationFn = TFunction<'translation'>
 
@@ -37,6 +39,7 @@ const SITE_TRANSLATION = gql`
     }
   }
 `
+let map_language: LanguageTranslation | null
 
 export const loadTranslations = () => {
   const [loadLang, { data }] = useLazyQuery<siteTranslation>(SITE_TRANSLATION)
@@ -53,6 +56,8 @@ export const loadTranslations = () => {
       i18n.changeLanguage('en')
       return
     }
+
+    map_language = language
 
     switch (language) {
       case LanguageTranslation.Danish:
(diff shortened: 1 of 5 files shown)


kkovaletp:
I've merged this PR into the #910 - a temporary alternative to the `master` while the project is on hold, so users still could get the latest Photoview improvements and fixes by building an image locally from this PR code

viktorstrate:
I really like this addition, and I think it would be a good idea to merge it. But I think we recently added Ukrainian which should probably be added before we can merge this.

kkovaletp:
Oh, didn't expect that there might be an ordering issue.
@jordy2254 , please review this PR and check the merged commit https://github.com/photoview/photoview/commit/a90f30391cc8a2a7c892031da5d9a2ab46ac31b4

Do you see an issue with merging this PR after that commit is done?
If yes, what would be the best way of solving this issue now? Reverting the branch to some earlier commit will also revert other merged PRs and we'll need to deal with that as well...

kkovaletp (dismissed):

ui/src/localization.ts
```diff
@@ -148,3 +153,55 @@ export const loadTranslations = () => {
...
 }
+
+export const SetMapLanguages = (map: mapboxgl.Map) => {
+  if (isNil(map_language)) {
+    map.addControl(new MapboxLanguage({ defaultLanguage: 'en' }))
+    return
+  }
+
+  switch (map_language) {
+    case LanguageTranslation.Danish:
```
jordy2254, line 164: We already have all of this information within the `Userpreferences.tsx` called `languagePreferences`. We should extract this and the elements in `languagePreferences` into a map within this file which is `LanguageTranslation -> languageCode` Although in this case it does appear that the Chinese traditional & simiplified do not use the same code as we do within `UserPreferences`.

Rewriting it in this way gives us somewhere more central to pick up on language codes and means things don't get missed in the future.

jordy2254:
@kkovaletp Merge conflicts unfortunatly don't work in that way. A merge conflict happens when 2 people edit the same area of code, in this circumstance neither have modified the same area.

@WindLi001 I have added a comment to make this more generic and add a tie for this to the language selection using a map which should make it so that people adding languages in the future are less likely to miss it. A test could be written for this to ensure the `userPreferences` list is the same length as the number of map keys.

ui/src/localization.ts
```diff
@@ -148,3 +153,55 @@ export const loadTranslations = () => {
...
   }, [data?.myUserPreferences.language])
 }
+
+export const SetMapLanguages = (map: mapboxgl.Map) => {
+  if (isNil(map_language)) {
+    map.addControl(new MapboxLanguage({ defaultLanguage: 'en' }))
+    return
+  }
+
+  switch (map_language) {
```
jordy2254, line 163: We already have all of this information within the `Userpreferences.tsx` called `languagePreferences`. We should extract this and the elements in languagePreferences into a map within this file which is `LanguageTranslation -> languageCode` Although in this case it does appear that the Chinese traditional & simplified do not use the same code as we do within `UserPreferences` (You might need an object rather than one field in the map).

Rewriting it in this way gives us somewhere more central to pick up on language codes and means things don't get missed in the future. A test can also be written to ensure the language selection list length is the same as the map to stop someone doing something strange in the future.

jordy2254 (changes requested):

kkovaletp:
@jordy2254 as the PR owner is not responding, but @viktorstrate and I think that this improvement is adding value to the Photoview, Can you please fork it to your repo (by a command, similar to `git fetch upstream pull/797/head:797-map-lang`), go required changes, and merge it to master instead of this PR?

BTW, this PR is 1 of a few still not merged from the #910, so we're close to our 1st milestone)

kkovaletp:
I just found the #672, where there is the link to the MapBox example, so copy it here just in case: https://docs.mapbox.com/mapbox-gl-js/example/language-switch/

kkovaletp:
@vl-ivanov, please review this PR.

vl-ivanov:
@kkovaletp there are multiple unresolved issues that most likely the original PR author will not implement, since there are no comments from @WindLi001 after creating the PR.
Issues:
- Missing Ukrainian translation
- Not resolved PR comments 
- Code duplication
- No tests at all

kkovaletp:
I'm closing this stale PR, as there are plans to migrate from MapBox to MapLibre: https://github.com/photoview/photoview/pull/1403

PR #799, opened 2023-02-10
Use Clean() but not Abs() to prevent directory traversal

WindLi001:
The old code use filepath.Abs() to prevent directory traversal, and that's not so good.
The code of filepath.Abs() is below,
```
// Abs returns an absolute representation of path.
// If the path is not absolute it will be joined with the current
// working directory to turn it into an absolute path. The absolute
// path name for a given file is not guaranteed to be unique.
// Abs calls Clean on the result.
func Abs(path string) (string, error) {
	return abs(path)
}

func unixAbs(path string) (string, error) {
	if IsAbs(path) {
		return Clean(path), nil     // When the input path is absolute, the Abs() just call Clean() to make it safe. 
	}
	wd, err := os.Getwd()
	if err != nil {
		return "", err
	}
	return Join(wd, path), nil        // When not, the Abs() add the work directory before the it.
}

```
I think adding the working directory of application photoview before the input path is no sense. And because the Clean() isn't called after adding the working directory before input path, the .. is still in the path, and directory traversal doesn't be prevented.
So, I think no matter the input path is absolute or not, Clean() should be calles, and that's enough. BTW, because the input path is URL path but not file path, I use path.Clean() but not filepath.Clean() to it.

### DIFF
--- a/api/routes/spa.go
+++ b/api/routes/spa.go
@@ -3,6 +3,7 @@ package routes
 import (
 	"net/http"
 	"os"
+	"path"
 	"path/filepath"
 )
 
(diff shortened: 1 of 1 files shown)


viktorstrate:
Looks good, thank you

PR #802, opened 2023-02-15
Add support to parse DateTimes with UTC+Offset

john-dev:
The former change added support for various Dates in a specific order but did not take the UTC+Offset into account.


john-dev:
Thanks for the merge @viktorstrate - any chance you can fix the build action so the docker image is available to anyone?

Issue #571, opened 2021-10-13
Permission system


viktorstrate:
**Is your feature request related to a problem? Please describe.**
As an admin it would be nice to have fine grained control over what each user can and can not do.

**Describe the solution you'd like**
An option in the Settings page to enable/disable various features on a per user basis.
Such features might be:

- Change their own password (#263)
- Allow user to share their media with public links (#555)
- Allow user to start scanner for their own media
- Possible more in the future

viktorstrate:
You are right, thank you for referencing it. I will update the original issue, but I also think it's fine to have a separate issue for discussion things specific to the permission system.

WindLi001:
According to #808, I think there should be a `Capability` named `Face management`. If checked, then the user can change the face reorganization results of the photos in the path.

WindLi001:
> For reference, taken from [#263 (comment)](https://github.com/photoview/photoview/issues/263#issuecomment-868957865)
> 
> I think the permission table could be integrated directly into the `capabilities` column. Just by adding more checkboxes besides the admin one. (this way the "permission table" part is already created)
> 
> <img alt="Screenshot 2021-06-26 at 08 41 15" width="872" src="https://user-images.githubusercontent.com/4233458/123504545-481a0e80-d65a-11eb-84fe-83fa5c0ffd7a.png">

I think the `Admin` capability don't need to be set at user-path level. It should be kept in user level. Usually `Admin` means user management operations, including adding user, deleting user, resetting another user's password, authorizing a user to access a path, and so on. All of these operations are user level.
Other capabilities, such as `Face management`, `Download`, `Sharing`, `Scan` maybe need to be set at user-path level. 

Maybe `Face management` and `Scan` can be merged into a single capability named `Write` or `Management`. Then all the other operations which can modify the photo data added in the future could be authorized based on this capability. 

BTW, I think it will be better that allow all the user to change their own password, even they don't have `Admin` capability.

jordy2254:
As discussed on the discord and from #263 I'm interested in picking this one up and working on it over the next few weeks or so.

PR #808, opened 2023-03-01
Reject face modification from non-admin user

WindLi001:
Now both admin and non-admin user can modify face reorganization results of the photos. For example, the non-admin user can change the face name, merge the face group, and so on. I think maybe it is better that only allow the admin user has the write permission, and keep the non-admin user in read-only mode.


viktorstrate:
A usual use-case is for non-admin to have access to their own set of images using root directories that do not overlap with other users in the system. In this case, it makes the most sense to allow the user to organize their own images but be able to modify other users as an admin can.

An ideal soltion would be to have a more fine-grained capability system as suggested in #571

WindLi001:
OK, obviously my use-case is a simple one. All the users have the same set of images. One of them (in fact, I) mange the gallery, and the others only need to browse the gallery. So, this commit only suit to me, but not to other one.
I will have a look to the #571. I think we can discuss it in there.

PR #814, opened 2023-03-07
Initial setup login page should hide password input

komali2:
The form field for password `type` property wasn't set correctly, so passwords would display in browser as user typed them in

### DIFF
--- a/ui/src/Pages/LoginPage/InitialSetupPage.tsx
+++ b/ui/src/Pages/LoginPage/InitialSetupPage.tsx
@@ -119,6 +119,7 @@ const InitialSetupPage = () => {
                 ? 'Please enter a password'
                 : undefined
             }
+            type="password"
           />
           <TextField
             wrapperClassName="my-4"


viktorstrate:
If we hide the password field, I think we should also add a "repeat password" field, as it would be a hazzle if users made a typo and had to reset Photoview fix it.

Zhen-Bo:
> If we hide the password field, I think we should also add a "repeat password" field, as it would be a hazzle if users made a typo and had to reset Photoview fix it.

or maybe add a preview button to the field, can toggle password hide/display when you click it

komali2:
I agree with the two comments, however the current state is a security vulnerability, and the current PR fixes it, so in my opinion the two other mentions should be implemented separately to allow this small, more vital, change to be merged more immediately upon approval (and be easier to review).

dasKeks28:
Another issue with the non-password type field is that it's not detected by password managers.
At least 1Password in my case didn't offer to generate or save a password, so I had to do it manually.

And even worse is the fact that the browser might store the password in clear text for input suggestions on the site.

In my opinion a single password field, without "repeat password" is still way better than using a plain text field.

kkovaletp:
> I agree with the two comments, however the current state is a security vulnerability, and the current PR fixes it, so in my opinion the two other mentions should be implemented separately to allow this small, more vital, change to be merged more immediately upon approval (and be easier to review).

I've created the #910 as a temporary alternative to the `master` while the project is on hold. I like your PR and would merge it to this branch, but together with the implementation of the mentioned enhancements: 2 fields for password and show\hide functionality. Please point me to the PR, implementing that and I'll merge it all together

jordy2254:
I shall make a password confirmation component in the UI under #918 I think it's good to get this one in in the meantime. @kkovaletp Are you happy for this to be merged or should we wait until the password confirmation piece is done?

jordy2254 (approved):

kkovaletp:
> Are you happy for this to be merged or should we wait until the password confirmation piece is done?

I'd like to wait until we have a solid solution

jordy2254:
@kkovaletp given my limited time to the project at the moment I think we should move this I'll keep an eye on it when I go back to the other password piece and tidy up any duplication. In the meantime the code for this is fine and fixes an issue.

kkovaletp:
@jordy2254, I agree that the issue, described in the 1st PR's message here, is valid and the PR fixes it. However, without implementing at least 1 of the mentioned later enhancements (password preview and/or repeat password) fields, the end-user experience would be worse than now.
Also, as the initial setup is a 1-time operation, I don't see a big impact (and so, high severity) of this issue, as admin can find a place and time, where nobody is watching the screen. Also, as part of the server setup, an admin has to fill the `.env` file with DB passwords and an optional map token, which are as sensitive, as the password on the init page, but as that is a plain-text file, it is not possible to hide those inputs and it's clear that this operation has to be done while nobody is watching the screen.

> the browser might store the password in clear text for input suggestions on the site.

This is a valid difference, however, (from my point of view) if a user didn't disable this browser's spying feature, stored password in the browser's suggestions pool wouldn't be the biggest sensitive data leak in most cases, so again, I don't see a reason for high priority and urgent merge.

PR #823, opened 2023-03-21
Fix invalid UTC offset during datetime parsing

yosev:
In my initial request a few month's ago, i passed an invalid UTC offset format.
+02:00 must be -07:00

See [here](https://go.dev/src/time/format.go) line 68

### DIFF
--- a/api/scanner/exif/exif_parser_external.go
+++ b/api/scanner/exif/exif_parser_external.go
@@ -122,12 +122,12 @@ func (p *externalExifParser) ParseExif(media_path string) (returnExif *models.Me
 				found_exif = true
 				newExif.DateShot = &dateTime
 			} else {
-				layoutWithOffset := "2006:01:02 15:04:05+02:00"
+				layoutWithOffset := "2006:01:02 15:04:05-07:00"
 				dateTime, err = time.Parse(layoutWithOffset, date)
 				if err == nil {
 					found_exif = true
 					newExif.DateShot = &dateTime
-				}	
+				}
 			}
 			break
 		}


subtlepseudonym:
It may be a good idea to add a timestamp to the tested exif data

john-dev:
> It may be a good idea to add a timestamp to the tested exif data

Could you do that as i don't know a good way to implement this test as your test-files do not expose such a field in the correct order.

subtlepseudonym:
I spent some time digging into this, but I'm by no means an expert on EXIF data. It looks like exiftool only returns UTC offset values of +00:00 because the fields holding that data are stored as unix timestamps. Given that, I don't believe we _can_ test this functionality because it's not actually used (beyond preventing the 'extra text: "+02:00"' error when we fall back to the `FileModifyDate` tag).

We _can_ get UTC offset data from EXIF tags via the `OffsetTime*` tags added in EXIF v2.31, but I'm unsure how widely used they are. Also, the EXIF parsing package we're using for the internal parser was archived in 2020, so updating that isn't worth it just to add time zone support.

So yeah. Probably fine without a test. Apologies if I sent you down a rabbit hole as deep as the one I just went down.

kkovaletp:
@john-dev, I've created #910 as a temporary alternative to `master` while the project is on hold.

I'm not a developer, so could you please explain why we have the date hardcoded there and why it needs to be exactly that particular one, so you propose a PR to change the time zone? How does that function work in general?

If there is a reason and this is a bug that needs to be fixed, I'll merge it into my branch, so others can use the fixed version.

subtlepseudonym:
That particular timestamp is used as a reference time to parse format strings. The standard library defines a few other formats using the same timestamp here:
https://pkg.go.dev/time#pkg-constants

kkovaletp:
Done, PR is merged in my branch

viktorstrate (approved):

kkovaletp (approved):

jordy2254 (approved):

Issue #659, opened 2022-03-10
Password can be changed to an empty password, but on login password field mustn't be empty


TwlyY29:
Hi there,

first of all: thank you for this great piece of software! I use it to organize and access my photo library and I really love it!

Since I have photoview running in my home network unaccessable from the outside, I've been thinking about how I could make it easier for myself to access the library. My take was to set an empty password. The UI allows to change the password to an empty string. However, on login, I'm required to enter something into the password field.

I guess this is not a bug report, but something you maybe want to consider. 

Ok, bye then. Figuring out how to get access again ;-) 
Mirco

kevin-king:
I just ran into this same issue. I fixed it by (1) deleting the user in the database and (2) by resetting the initial_setup flag. All existing settings and scans were unaffected, which was nice since scanning over 100GB of photos takes FOREVER.

```
$ docker ps
CONTAINER ID    IMAGE             ...
e088e2aca383   mariadb:10.5   ...

$ docker exec -it e088e2aca383 bash

root@e088e2aca383:/# mysql -u photoview -p photoview
<enter password>           (password is photosecret by default, check your docker-compose.yml)

MariaDB [photoview]> select * from site_info;
---------------------------------------------------------
initial_setup  periodic_scan_interval  concurrent_workers 
---------------------------------------------------------
0              0                       3              

MariaDB [photoview]> select * from users;
---------------------------------------------------------------------------------------
id  created_at                updated_at                 username   password      admin
---------------------------------------------------------------------------------------
10  2022-04-06 13:06:50.897   2022-04-06 17:24:35.167    kevin      <hidden>      1 

MariaDB [photoview]> update site_info set initial_setup=1,periodic_scan_interval=0,concurrent_workers=3;

MariaDB [photoview]> delete from users where id=10;
```

Database credentials for the above are found in docker-compose.yml
```
- MYSQL_DATABASE=photoview
- MYSQL_USER=photoview
- MYSQL_PASSWORD=photosecret
```

Now log out, back in again, and reconfigure your user with a non-empty password.

Issue #624, opened 2021-12-19
Advanced authentication options


mindfuucker:
**Is your feature request related to a problem? Please describe.**
My photoview instance is running behind Authentik. This causes two login screens for users to be traversed.

**Describe the solution you'd like**
From simple to advanced there are a couple of options:

1. Disable authentication
2. Disable authentication with IP/CIDR whitelist
3. Basic auth (prompt or http://username:password@example.com/)
4. Header based authentication (like [Navidrome](https://www.navidrome.org/docs/usage/security/#reverse-proxy-authentication))
5. oidc/oauth2 integration
6. LDAP integration

**Describe alternatives you've considered**
Editing the HTML and hardcode a password at the login screen

**Additional context**
All options up to option 4 are quite easy to implement without a lot of hassle. 
I personally prefer header based authentication as you will be able to pass through additional information in headers. It's also easy to use as each reverse-proxy is able to inject some headers. It's also safe as the Header is a variable and can be different for each instance.

viktorstrate:
The immediate problem I can think of with something like header based authentication is that loading images relies on cookies as there aren't any ways to send custom headers for `<img />` tags.

So the server would have to recognize that an authorized request is made based on the headers and then set the cookie for the response. I guess this might be ok, but I don't know if there might be any pitfalls doing it that way.
I guess the headers could easily become out of sync with the cookie?

rpatel3001:
Just want to +1 header auth as all my services are behind traefik with traefik-forward auth, which sets X-Forwarded-User after a google login flow. I currently am the only user so disabling auth altogether would also work for now. I'm not familiar with this codebase and how it uses cookies, but [grafana](https://grafana.com/docs/grafana/latest/auth/auth-proxy/) is another potential reference for implementing this.

viktorstrate:
@rpatel3001 Grafana might be worth a look.

Maybe having an environment variable to completely disable any form of authentication and just have a single admin user that is always logged in, might be a nice start.
I think that would be simple and useful for many users who either run the instance locally-only or behind other means of authentication.
That way cookie authentication could also simply be turned off completely.

Sieboldianus:
This is only somewhat related: Is there any way to increase the lifetime of the auth cookie? 

[...]

Issue #478, opened 2021-08-23
Photo browsing without logging in


ray73864:
I'm using Photoview in a school environment as our mapped network drive that stores all the photos is rather unweildy and difficult for students to navigate and find the photos they need.

For staff it is fine, as the directory layout works quite nicely for their needs.

What we would like however is to have the admin login like we currently do which periodically scans the mounted CIFS from the windows server, but have it so that students can browse all the photos without having to log in.

Security isn't much of a problem, since you can't even access the photoview webserver unless you are authenticated on the network anyway.

viktorstrate:
I think that would be a nice addition!

luusl:
Yeah, that would be nice.

Skayo:
I'd love to be able to disable auth completely so I can throw Authelia in front of it

kkovaletp:
#930 describes this request better, so I'll close this one as a duplicate, even though it was created earlier

Issue #174, opened 2021-01-26
Gallery view page for password-less shares.


tehniemer:
I'd like the option to turn the login (root) page to a gallery view of password-less shares, and another option at the share level to show password-less shares on the gallery page or not, thoughts? 

This would make sharing more than a single album with people much easier, and could be easily communicated verbally.

kkovaletp:
While this request looks logical, it breaks the workflow of the shares: even passwordless shares have their unique links, and only those who know the exact link can view the shared content. Publishing them on the root page would break that part.

I think that #930 issue covers this use case better, so I'm closing this issue as not planned

PR #828, opened 2023-04-07
Remove login page non-empty password requirement

subtlepseudonym:
Passwords are allowed to be empty in the database, but the login page fails form validation if an empty password is entered.
This change _only_ removes that validation. This isn't a great solution, but will allow password-less users in the short term.

This "bug" / feature request has been mentioned a few times:
- [#478](https://github.com/photoview/photoview/issues/478)
- [#659](https://github.com/photoview/photoview/issues/659)
- [#174](https://github.com/photoview/photoview/issues/174) (semi-related)

More extensive changes to the authentication system have been discussed in [#624](https://github.com/photoview/photoview/issues/624)

### DIFF
--- a/ui/src/Pages/LoginPage/LoginPage.tsx
+++ b/ui/src/Pages/LoginPage/LoginPage.tsx
@@ -96,12 +96,7 @@ const LoginForm = () => {
         className="w-full"
         type="password"
         label={t('login_page.field.password', 'Password')}
-        {...register('password', { required: true })}
-        error={
-          formErrors.password?.type == 'required'
-            ? 'Please enter a password'
-            : undefined
-        }
+        {...register('password', { required: false })}
       />
       <input
         type="submit"
--- a/ui/src/Pages/LoginPage/LoginPage.tsx
+++ b/ui/src/Pages/LoginPage/LoginPage.tsx
@@ -96,7 +96,7 @@ const LoginForm = () => {
         className="w-full"
         type="password"
         label={t('login_page.field.password', 'Password')}
-        {...register('password', { required: false })}
+        {...register('password')}
       />
       <input
         type="submit"


kkovaletp:
@jordy2254, as you're working on user management epic, I've assigned this PR to you. Feel free to decide whether we need it, can include it in the final solution, or decline it.

From the product perspective, I'd say that in general, we might support passwordless user scenario, but:
- it should be disabled by default
- can be enabled by Admin on the Settings page
- a big warning should be shown to Admin when he\she performs the enable operation.

ui/src/Pages/LoginPage/LoginPage.tsx
```diff
@@ -96,12 +96,7 @@ const LoginForm = () => {
         className="w-full"
         type="password"
         label={t('login_page.field.password', 'Password')}
-        {...register('password', { required: true })}
-        error={
-          formErrors.password?.type == 'required'
-            ? 'Please enter a password'
-            : undefined
-        }
+        {...register('password', { required: false })}
```
jordy2254, line 99: I don't think this register is required now that it's false... I'd imagine this would be the default right?

jordy2254:
I think given I'm not touching this area of code and this is fixing a bug present in the UI it's safe to merge for now. Ultimatly a user could access the graphql api directly and do the actions. Anything surrounding authentication like this needs to be cohesive with the backend and thus this should be merged.

@subtlepseudonym can you have a quick look at my comment I've left please to do with the change?

subtlepseudonym:
Your comment makes sense to me. I'm not near my dev machine, but can take a look during the week.

jordy2254 (approved):

kkovaletp (approved):

PR #829, opened 2023-04-08
Fix RAW file support for docker non-root users

subtlepseudonym:
Supporting RAW file formats requires running `darktable-cli`, which fails in docker for non-root users with the following error:
```bash
path lookup '/.config/darktable' fails with: 'No such file or directory'
```
Both `/.config/darktable` and `/.cache/darktable` must be present to run `darktable-cli`. Normally, these directories would be silently created on first run, but non-root users don't have permission to create directories.

This change creates those directories in the docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,8 +78,10 @@ RUN apt update \
   && apt install -y curl gpg libdlib19.1 ffmpeg exiftool libheif1
 
 # Install Darktable if building for a supported architecture
+# And create darktable directories for non-root users
 RUN if [ "${TARGETPLATFORM}" = "linux/amd64" ] || [ "${TARGETPLATFORM}" = "linux/arm64" ]; then \
-  apt install -y darktable; fi
+  apt install -y darktable && \
+  mkdir -p /.cache/darktable /.config/darktable; fi
 
 # Remove build dependencies and cleanup
 RUN apt purge -y gpg \


fowlerc74:
I have this same issue. Glad to see it was a simple fix. I would love if this could get merged

rhoot:
I had the same issue, but this did not fix it for me without modifying the paths. For some reason mine have to be created at `/app/.cache` and `/app/.config` rather than just `/.cache` and `/.config`. Even if I run `darktable-cli --version` manually from within the container, with `/` as the current dir.

Dockerfile
```diff
@@ -78,8 +78,10 @@ RUN apt update \
   && apt install -y curl gpg libdlib19.1 ffmpeg exiftool libheif1
 
 # Install Darktable if building for a supported architecture
+# And create darktable directories for non-root users
 RUN if [ "${TARGETPLATFORM}" = "linux/amd64" ] || [ "${TARGETPLATFORM}" = "linux/arm64" ]; then \
-  apt install -y darktable; fi
+  apt install -y darktable && \
+  mkdir -p /.cache/darktable /.config/darktable; fi
```
kkovaletp, line 84: The `.cache` and `.config` folders are created in the user's home folder: in case, the image is running under the root, they are created in the `/root` folder. In case, it is running under a non-root user, it will be in the `/home/<user>` folder. There should be permission to create files and folders in user's home folder, so I don't see a need to do it at the image creation time (which will require also `chown` and `chmod` commands to be run, as Dockerfile is executed by root and folders, created here, are owned by root as well). It makes sense only if you also add the `USER` declaration at the end of the Dockerfile, which will result in the creation of the image, always running under specified non-root user

kkovaletp: but if for some reason the `darktable`, executed under non-root user, wants to create these folders in the Dockerfile's WORKDIR location, then from my point of view it makes more sense to set the RWX permissions to the folder and all its content for all users by the command `chmod -R 777 /app` - this is the application's working folder and the application has to have permissions to do anything there, even if run in non-root mode.

kkovaletp: I've updated my PR #863 with the changes, proposed in the previous comment. Please try to build and run the image from that PR and test whether the problem is fixed for you

kkovaletp:
@jordy2254, as the fix, proposed by this PR, is implemented in my #863 and @subtlepseudonym didn't respond to my comments, I propose to decline this PR.

subtlepseudonym:
> @jordy2254, as the fix, proposed by this PR, is implemented in my #863

This PR includes the missing darktable files whereas #863 changes the permissions for /app subfolders to be wide open (777). I believe adding the required files is a better solution to providing full write access to paths that are likely to be bound to volumes.


> and @subtlepseudonym didn't respond to my comments, I propose to decline this PR.

My bad! Which comments did I miss?

Dockerfile
kkovaletp: thanks for your feedback. I'll do a review of #863 and try to include the fix for this issue

kkovaletp: Today I did a code review of my PR #863 and rewrote it to build an image, configured to run under a non-root user.
As part of this change, I investigated the issue that Darktable fails to run because of missing 2 mentioned folders and with inability to create them. As I see, there is no need to create them in the `/` folder, as Darktable wants them to be in the user's home folder. So, once the `/home/user` folder exists and is owned by the user, under which Photoview is running, Darktable creates those folders inside it successfully and runs fine.

PR #843, opened 2023-05-11
Update docker-compose.example.yml for Traefik reverse proxy

ftoppi:
Add labels for Traefik reverse proxy, assuming it runs in host network mode

### DIFF
--- a/docker-compose.example.yml
+++ b/docker-compose.example.yml
@@ -15,6 +15,7 @@ services:
   photoview:
     image: viktorstrate/photoview:2
     restart: always
+    # Remove the following ports section if you enable Traefik reverse proxy labels
     ports:
       - "8000:80"
     depends_on:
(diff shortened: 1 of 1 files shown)


kkovaletp:
As mentioned in the #863 in Readme changes, I think that this shouldn't be merged into the project, as it creates a false impression that we support the integration with Traefic and should step in in case of some problems with it, while actually from my PoV such external integrations should be outside of the project.

viktorstrate:
Agree with @kkovaletp but it would be nice to add the example to the documentation website here: https://photoview.github.io/en/docs/installation-reverse-proxies/

ftoppi:
Feel free to reject the PR, I just made a suggestion that was fitting in my opinion.

kkovaletp:
Here is the PR to the doc page: https://github.com/photoview/photoview.github.io/pull/26
@viktorstrate - is this what you expect?

viktorstrate:
Yes, I've merged your PR. I'll close this one then ☺️

PR #861, opened 2023-07-01
PoC: support showing guests photos of a user with environment variable USER_TO_SHOW_GUESTS

KevinWang15:
Hello, and thanks for this fantastic project.

My use case is a little different, I'd like to self host this project, upload my photos, and make them all accessible to visitors without requiring them to log in.

Since this use case isn't currently covered, I did some hacking and created a proof-of-concept solution.

WIth this PoC, we could add env var `USER_TO_SHOW_GUESTS=admin`, and all visitors could see `admin`'s photos without needing to log in.

When 'admin' wishes to log in and perform administrative activities, they may still do so because '/login' is still present.

Since this PoC is merely a PoC, it shouldn't be merged. Maybe we can start from this point and see if we can add this feature.

---

Known problems with this PoC:
* "Initial Setup" will not work. (You have to disable `USER_TO_SHOW_GUESTS` to do "Initial Setup" first, then enable it, restarting the server)
* GraphQL authorization is still not correct (e.g. Guests could still create "share"s of an album). This should be easy to fix.
* It's a hack and it's very dirty


jordy2254:
@KevinWang15 given this is a POC and hasn't been looked at for a while could you please convert this PR to a draft. 
Having an issue raised for this would also be good so we can add it to the road map and establish how we'd like this functionality to work.

kkovaletp (changes requested):
Finalize the PI
I've submitted this review to block the Merge functionality from accidental merging

jordy2254:
@KevinWang15  I'm going to close this, I've created issue #930 to discuss this further and flesh out the idea. As you've correctly highlighted there are currently a few issues with the implementation. Feel free to start the discussion on the new issue so we can work out as a community how to implement this the best into photoview.

PR #650, opened 2022-02-20
Create a Development dockerfile

gwynm:
Modelled on the Production dockerfile, but simplifies the build process.
We mount the source code from the host machine and build it, then run
the servers in development mode, using Air to get live-reloading in Go.

Test photos are from thiscatdoesnotexist.com.

PR #863, opened 2023-07-07
Refactor docker deployment and user setup

### LINKED ISSUE #862: Errors in Photoview image
**Describe the bug**
When running Photoview with Docker compose, I consistantly got 2 errors "Error: failed to initialize database" and "No .env file found" is on my both Ubuntu and my Mac

yet the server still works fine and the environments are loaded properly

**To Reproduce**
Fresh install with [Photoview example](https://github.com/photoview/photoview/blob/master/docker-compose.example.yml), replace some hardcoded configs with `.env` file

**Expected behavior**
No misleading errors

**Screenshots**
image

kkovaletp:
Fix #862, address #826, and maybe some other tickets: reimplemented the docker-compose user setup to enhance the product setup experience and cover much more cases:
- extended original docker-compose example with many additional improvements
- added .env and moved all user-changeable variables there
- added Makefile with the most useful shortcuts to simplify product maintenance
- documented all of this with inline comments and updated the Readme

Please feel free to give your feedback and ask questions if you have any.


kkovaletp:
As a response to issue #876 I added the development environment setup. It could be enhanced by mounting the code folders to the Photoview container and running the server using dev server tools, like nodeamon, but these improvements require a good understanding of the code and dev environment, so if you see some benefits from that, as a developer, implement (and test) that from your side.

c13mn14k:
> While optimizing the Dockerfile, I noticed that Debian is used as a base image. While it is stable, it is old-fashioned and outdated compared to some newer Debian-based images (like Ubuntu). From my point of view, it is not an optimal base image. Does anybody have some objections against switching to ubuntu:latest? Maybe you see some even better options, taking stability and security into account.

I think both Debian and Ubuntu may be bloated for photoview, I'll try to write a Dockerfile using Alpine.

c13mn14k:
Your PR looks like a piece of nice work! I have one issue with it though, I think introducing a Makefile has some serious drawbacks:
- it adds another complexity layer - running docker through make _feels_ weird and these commands in makefile could be just simply bash scripts
- it requires that a new user installs Make to run photoview through docker, and 
- that the user is familiar with Make.
- using Make (another complexity layer) makes additional assumptions about user setup,
- commands used in Make do not simplify work much.

For me Make shouldn't be presented as the go-to way to setup photoview with docker. That setup should require docker only. Using Make creates complexity (system complexity and complexity for users) and additional assumptions: about users' OS and users' setup.

As an example, my homeserver is very thin besides being a docker host - thanks to docker I don't have a lot of packages, including make. I also combine all services related to similar areas inside one folder, so I have photoview together with other media services in one compose.

My proposition is:
- Restore Docker Compose as the recommended way to run photoview with docker, remove reference to make in this section of docs
- Add documentation about Makefile as a section of setup: `## Getting started - Setup with Docker and Make`.
- (other solution) Delete makefile and move commands used there to
  - documentation
  - a few bash scripts inside a `scripts/` folder

This is a small change, but I think important. Otherwise your PR looks like a very helpful change for new users!

kkovaletp:
@c13mn14k, thanks for your review and detailed feedback!
I'll think about your proposals and implement some changes, however, I disagree with removing Make completely:
- I think that this is a common and convenient way to simplify user interaction: 1 simple-to-read file with all the scenarios vs a set of bash scripts (1 per scenario) or 1 script with more complex syntax and without auto-completion
- Makefile doesn't hide anything: it is a plaintext file, easy to read, and more: documentation contains a directive to go through it and review the commands and scenarios in the scope of the initial setup - user is free to modify it or not use at all, running the commands directly
- You did a review from your point of view, owning a server with minimal OS setup, and yes, from this point of view your comments and proposals look valid and logical. However, I believe that this is not a typical use case compared to a fully functional OS setup. Also, most probably, owners of such servers with a minimal OS are much more skilled, as managing it requires a deeper dive into the OS specifics.

docker-compose example/docker-compose.example.yml
```diff
@@ -0,0 +1,110 @@
...
+      ## *Additional* media folders can be mounted like this (set variable in .env file):
+      # - "${HOST_PHOTOVIW_MEDIA_FAMILY}:/photos/Family:ro"
+
+  db:
+    image: mariadb:lts
+    container_name: photoview-mariadb
+    restart: unless-stopped
+    stop_grace_period: 5s
+    ## Optimized MariaDB startup command for better performance and compatibility
+    command: mariadbd --innodb-buffer-pool-size=512M --transaction-isolation=READ-COMMITTED --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci --max-connections=512 --innodb-rollback-on-timeout=OFF --innodb-lock-wait-timeout=120
```
DaCHack, line 67: @kkovaletp Can these flags be added to the compose file when
- using mariadb:10.5 before
- and using without any flags before
without breaking the existing database?
Thanks!

kkovaletp: I guess, you're asking about the command flags in line 67. From my understanding, yes, but to make sure double-check the MariaDB docs. Also, I'd suggest doing a full DB backup and then setting up a new Photoview instance with my config, mapping to the new DB volume folder, running the DB, and restoring the backup. In this way, you ensure there is no data loss and corruption

kkovaletp:
Set it temporary to Draft state to make sure it is aligned with the current product workflow

kkovaletp:
Action Item: to review and try to implement #650 here, as the DEV dockerfile is already here

kkovaletp:
> [ ] to review and try to implement [Create a Development dockerfile #650](https://github.com/photoview/photoview/pull/650) here, as the DEV dockerfile is already here

@jordy2254, would you be able to help me with this action item from the dev perspective?
As we've discussed earlier, right now the dev-docker file is logically different from the main one only by having a single-stage build procedure, so all the code and intermediate artifacts are available, as well as the build tools.

However, to provide a truly useful docker file, we need to replace the statically built executable web server binary with a dynamically rebuilding web server for development (like [nodeamon](https://www.npmjs.com/package/nodemon) for frontend and backend or [an alternative tool](https://techinscribed.com/5-ways-to-live-reloading-go-applications/)).

For this type of change, I'd need a developer, who can configure the corresponding tools from our project's code perspective and provide the step-by-step guide for me together with all needed configs.

Then I'll use them to create frontend and backend docker files for developers in our project.

Alternatively, I'll exclude the dev-dockerfile functionality from the PR completely, so that later someone else can create a new PR for that.

jordy2254:
@kkovaletp of course I can help, I think the development docker file provides less community value than the additional features of the permissions system and the self service password change? So as soon as those are done I shall pick it up, unless you feel we should focus this with a bit more urgency?

kkovaletp:
@jordy2254, I'd agree with you that the dev environment is a specific feature for a relatively small subset of our community, so I don't see any urgency here. 
On the other side, this PR is the 1 of a few left not being merged to `master` from those, being merged into the #910. As we defined earlier, merging the PRs, which participated in #910, is our next milestone for the release, while user management functionality is not part of it. 
So, from the release point of view, probably it would be better to exclude the dev environment from this PR, finalize everything else to get it ready, merge, and release a new version (when everything is ready). Then we could prepare a new PR for the dev environment for the next release.
What do you think?

jordy2254:
@kkovaletp Yeah I agree, focusing the effort's onto the other changes that are required for it to be merged, and then making a new issue for the outstanding piece would be best. There's still a few action items left on this as well as some of those things we've discussed on discord, are you planning on doing those pieces? Did you need any support?

kkovaletp:
@jordy2254 I plan to finish this PR and move it to the Ready state ASAP, and then I'll review the other leftovers.
Let's not mix things and discuss here only this PR. We still can discuss the release on Discord)

Dockerfile
```diff
@@ -20,75 +20,67 @@ ARG COMMIT_SHA
...
 WORKDIR /app
 
-COPY api/data /app/data
-
-RUN apt update \
+# Create a user to run Photoview server
+RUN groupadd -r photoview \
+  && useradd -r -g photoview photoview \
+  && mkdir -p /home/photoview \
+  && chown -R photoview:photoview /app /home/photoview \
```
Omar007, line 67: This can be greatly simplified:
1. `useradd` can take care of the user group (`-U`) as well as the home directory (`-m`, optionally with `-d /path/to/user/dir` if you want to set a custom directory)
2. The app dir shouldn't be chowned so this can be removed

```suggestion
RUN useradd -r -U -m photoview \
```

Dockerfile
```diff
@@ -20,75 +20,67 @@ ARG COMMIT_SHA
...
+  && apt purge -y gpg \
   && apt autoremove -y \
   && apt clean \
   && rm -rf /var/lib/apt/lists/*
 
-COPY --from=ui /app/dist /ui
-COPY --from=api /app/photoview /app/photoview
+COPY --chown=photoveiw:photoveiw api/data /app/data
+COPY --chown=photoveiw:photoveiw --from=ui /app/dist /ui
+COPY --chown=photoveiw:photoveiw --from=api /app/photoview /app/photoview
```
Omar007, line 83: Likewise here, these should not be chowned. I would also suggest grouping all the application files used to run photoview under `/app`; move the `/ui` directory so it's also placed in `/app` (don't forget to update the `PHOTOVIEW_UI_PATH` to `/app/ui` in that case).

```suggestion
COPY api/data /app/data
COPY --from=ui /app/dist /app/ui
COPY --from=api /app/photoview /app/photoview
```

Dockerfile
```diff
@@ -20,75 +20,67 @@ ARG COMMIT_SHA
...
 
 ENV PHOTOVIEW_LISTEN_IP 127.0.0.1
 ENV PHOTOVIEW_LISTEN_PORT 80
```
Omar007, line 86: As a side-note, the default here should probably be changed to `8080` if we want it to work out-of-the-box without the need to grant any special capabilities (and don't forget about the `EXPOSE` below in that case).

Dockerfile
Omar007: One side-note on the `/app` ownership; the Dockerfile currently sets it as the WORKDIR, which in turn results in the cache location defaulting to `/app/media_cache` if not otherwise specified by the user by setting `PHOTOVIEW_MEDIA_CACHE`.  
To ensure a good out-of-the-box experience for that situation, the above suggestion could be expanded to prepare this default directory for the end-user, e.g.:
```suggestion
RUN useradd -r -U -m photoview \
  && install -dm755 -g photoview -o photoview /app/media_cache \
```

Omar007 (changes requested):
Since I provided some feedback on this subject through #829 , I figured I'd provide a proper review of the Dockerfile specifically. Hope this helps :+1:

Dockerfile
kkovaletp: We provide a web-server app, so its default HTTP port should be 80. Then a user can safely map it to whatever free port, he\she wants.
We provide the `docker-compose.yml` with the port mapping and document the procedure of the initial setup. Also, the port 8080 might be used as well, so I don't see a reason for this change.

Dockerfile
kkovaletp: > Likewise here, these should not be chowned. I would also suggest grouping all the application files used to run photoview under `/app`; move the `/ui` directory so it's also placed in `/app` (don't forget to update the `PHOTOVIEW_UI_PATH` to `/app/ui` further down the Dockerfile in that case).

I've already asked about why not to chown the app dir.

Thanks for the idea of grouping the app files in 1 folder - I'll add this as well

Dockerfile
Omar007: > Why? This is the application folder, why the application user shouldn't own it?  

The application should be available for running by the (any) user but you don't want the user (or anything and everything getting its privileges) to just modify w/e it wants in those applications.  
Especially for internet facing services doing so would be a risk; the moment an attacker manages to break in, they'll immediately have the user's privileges and thus the capabilities to edit all the files that are served to end-users. When those files are instead owned by root and only executable for other users, this can not be done without also somehow managing a privilege escalation first.  
The executing user should only ever need to have execution permissions for the application. The only ownership it needs is over its own user directories.  

When you look at your own system, you should see very similar behaviour. Your (/usr)/(s)bin isn't owned by your user but by root. You use a package manager, run with root permissions, to install and update your applications but you then execute the individual applications using your own user's privileges. It does not have the capabilities to change those binaries but any runtime data is written to your user's folders (e.g. the XDG directories for data, config and cache, dot-files in $HOME, etc.)

Omar007: Fwiw, an alternative for the default media cache situation could also be to just set it to the user dir as part of the Dockerfile, e.g. `ENV PHOTOVIEW_MEDIA_CACHE=/home/photoview/media_cache`, it doesn't need to be prepared in that case as the home is already set up during `useradd`.

Dockerfile
Omar007: Port 80 is a privileged port. It either requires root access or assignment of CAP_NET_BIND_SERVICE. You can not use this port otherwise. What port you use outside of the container doesn't matter, that can be w/e you want and does not need to be changed.

Dockerfile
kkovaletp: > One side-note on the `/app` ownership; the Dockerfile currently sets it as the WORKDIR, which in turn results in the cache location defaulting to `/app/media_cache` if not otherwise specified by the user by setting `PHOTOVIEW_MEDIA_CACHE`. To ensure a good out-of-the-box experience for that situation, the above suggestion could be expanded to prepare this default directory for the end-user, e.g.:
> `&& install -dm755 -g photoview -o photoview /app/media_cache \`

Why do we need to drop chowning the root application folder, but then create in advance and chown the nested folder for the cache? what is the real benefit of such self-limiting actions?

I didn't even test this, but if this will work and by now Photoview doesn't require the ownership of the /app folder, it might be changed in the future and this limitation is not so obvious - which creates a potential headache for future debugging and maintenance (similar to the issue with the Darktable, failed to run without having write access to the user's home folder)

Dockerfile
Omar007: To clarify, there's 2 layers of ports here. The compose file defines a port mapping for Docker; you're asking your container runtime to open port X and forward it to port Y inside the container. What port X is doesn't matter and is up to the user as long as their container runtime has the permissions to bind the chosen port. This is about port Y, the one inside the container. Since the container is not run with root or assigned special capabilities, it can not bind port 80. I.o.w., without either as a user deciding to run as root after all, assigning CAP_NET_BIND_SERVICE to the container or setting PHOTOVIEW_LISTEN_PORT manually to something above 1024, this will not run out-of-the-box if it's not changed.

I suggested port 8080 for this value as that is generally the most commonly used port for non-privileged HTTP but anything above 1024 is technically fine.

kkovaletp: > Port 80 is a privileged port. It either requires root access or assignment of the CAP_NET_BIND_SERVICE capability. You can not use this port otherwise. What port you use outside of the container doesn't matter, that can be w/e you want and does not need to be changed.

Ok, so what is the problem? Sorry, I still didn't get the point)
Now the app is successfully run under the `photoview` user with 80 port internally. Why should we change it?

Dockerfile
Omar007: > Why do we need to drop chowning the root application folder, but then create in advance and chown the nested folder for the cache?  

The only reason this is needed is because the Dockerfile sets the WORKDIR to /app. There's in total 2 options you can take to prevent this from being needed in the first place:

1. Set the WORKDIR to the user's $HOME
2. Set the PHOTOVIEW_MEDIA_CACHE (see also https://github.com/photoview/photoview/pull/863#discussion_r1576304311 )

One of these would have my preference over the other case anyway but I wasn't sure what the expectations around WORKDIR and PHOTOVIEW_MEDIA_CACHE are so I went with the default for that to base my first suggestion on.


> what is the real benefit of such self-limiting actions?  

There is no self-limiting action here. There's only fixing inconsistencies, preventing bad practices and resolving security problems with the proposed solution.  


> it might be changed in the future and this limitation is not so obvious  

That should never be made to happen. Binary and runtime/config dirs should _always_ be kept separate. The application should never, ever, write in it's binary directory.  


> which creates a potential headache for future debugging and maintenance (similar to the issue with the Darktable, failed to run without having write access to the user's home folder)  

Which had nothing to do with the application folder but everything with the user directories, exactly that which this PR should resolve and straighten out in the first place doesn't it??

Dockerfile
Omar007: https://github.com/photoview/photoview/pull/863#discussion_r1576323967  

It might be that the container runtime and/or the version of a given runtime has some impact on this as well. I believe Docker at some point changed its default configuration to allow the use of privileged ports by containers but I don't know in which version, nor of any other container runtime that allows this, hence the above note.

kkovaletp:
@Omar007, Thanks for your detailed explanations! Now I see your point and the benefits.

I tried to include all your suggestions (which then broke the image and I had to troubleshoot the issues and find solutions), so now it works again with the suggested changes. Please review the PR again, but now please check the changed files in the next order: Readme > Dockerfile > example.env > docker-compose.example.yml > Makefile - so you see the whole config set and better understand the idea.

You've provided valuable feedback and suggestions, so I'd appreciate it if you could review the complete solution and provide your opinion as well

Omar007 (commented):
It looks like you basically incorporated everything, very nice! I've added a few minor change suggestions but that's really just polishing at this point imo. Nice job  :+1:   

In regards to the other files beyond the Dockerfile, the README seems to correspond with the changes made so I don't have anything to add here.  
More generally though, I'm not sure about the addition of the Makefile system for something as basic as a deployment and such. Looking through the PR it doesn't seem like I'm the first to make this remark either but I genuinely don't know what the right answer is here.  
If I made the PR in question, I would personally have split it up since the changes don't really have a relation to each other; container fixes/improvements is one, Makefile wrapper is two and I'd say the watchtower addition could've been a third one.  

However, I'm not the author nor even a maintainer of this repo.Additionally, since this doesn't impact the container quality, nor does it add or remove anything relevant to me personally I don't have a personal stake in it either. As such, I can only provide more generalized observations in this regard:
- Most home users probably know/have Docker set up but I would not expect the same for make. Instructions reflecting the most common use, using the most commonly used tooling for the target demographic is probably best. I don't think `make` is it.
- Pure (container) servers are generally not set up with `make` software present.
- It adds another layer, obfuscating the relevant actions. This is fine if you can read Makefiles but otherwise...

More than that I can't provide you in regards to the other files/changes.

kkovaletp:
Thanks, @Omar007, for the valuable feedback! I'll try to implement the suggested changes, as most of them make sense to me too.

I'd like to ask you about the Makefile more:
I believe that a regular Photoview user doesn't have deep knowledge of docker, and even with them it is a time-consuming and iterative task to find an optimal sequence of commands for a product, so providing them as part of the solution is better.
Also, a user is interested in performing short, clear, and meaningful commands from the product level to manage it in a daily routine instead of remembering and typing longer docker-compose commands (in some cases there are more than 1 for a scenario). So, providing an abstraction layer for regular users is good, as it simplifies the routine. However, I also would like to make it readable for users, so if someone wants to understand what is going on under the hood or modify something according to the use case, it should be easy and straightforward - makefile from my perspective is the best option, as it provides linear lists of commands, separated by scenarios, so it should be really easy to understand it.
If thinking about an alternative, I'd say, it could be a SH script with all the same logic inside and the same list of simple options as parameters: `photoview.sh start`, `photoview.sh stop`, `photoview.sh backup`, ... But this solution will require a much more complex structure inside the script because of the SH language rules.
A workaround to simplify a bit the internal structure is to separate the `photoview.sh` into a list of individual scripts: 1 per scenario, like `start.sh`, `stop.sh`, `backup.sh`... but then we'll have a lot of files in the product root folder, which is not a problem, but not so clean as having 1 Makefile)

So, what do you think about this topic? Do you see a better solution?

Omar007:
> So, what do you think about this topic?  

Based on your message, I think the bigger question I would like to pose is; what are the expectations of the user and what is the workflow you desire from your users as well as what are you willing to make 'official' and thus support users with.  
If the expectation are:
- A regular user doesn't have deep knowledge of Docker but will know make.
- If a regular user doesn't know Docker or make, we want them to learn make instead of Docker.
- A regular user will always have Docker and make installed or will install them both (and knows how to on their platform).
- If a user has _any_ problem getting up and running using make, he can find support for it here.
Then this is probably fine. If not, this should be changed imo.  

My personal expectations and thus approach would be different. Not because that approach is necessarily better, it's just different because my expectations are different. I would expect:
- A regular user to already have (some) knowledge of Docker but not make (e.g. they're running other containerized services already or are using off-the-shelf systems).
- If they don't know Docker, I want them to learn that as opposed to make. They're much more likely to get use out of Docker knowledge for their (future) use case(s) from my perspective.
- No matter the platform, if it has support for running containers ootb, it's more than likely powered by some version of Docker.
- New tools/dependencies should be introduced sparingly or not at all if it can already be satisfied with standard tools and some documentation for optional/more advanced operations. New tools/dependencies also adds a burden to the team; does your whole team know make already? If not, what's your truck factor? Are you willing to support and maintain not just this new tool/code but also the new set of questions this opens the door for and you now have to deal with? Etc.  

> Do you see a better solution?  

Not really. In the end I still feel like this is a very subjective change. Standard (POSIX) scripts as you suggest could be more portable and it eliminates the dependency on additional software but it's still custom stuff abstracting away standard and reusable tooling.  
Whether those are individual scripts or a common script with args doesn't matter too much. You could just group them in a sub-directory instead of dumping them in the root  ;)

jordy2254:
@kkovaletp 
Im going to keep this short as in my phone...
I see a trend including myself that is 3 people that have said the make file doesn't achieve anything and adds a layer of complexity without any value. Ultimately it seems like this is a feature that you'd like and not a feature that the community would like. 

Someone without a technical knowledge that doesn't know docker isn't going to get on with make.  If we want tobuse make for assistance as ice said in the past it should be done utilising and aiding a bare metal installation from source.

docker-compose example/docker-compose.example.yml
```diff
@@ -0,0 +1,159 @@
...
+      # - "/dev/nvidia-uvm-tools:/dev/nvidia-uvm-tools"
+      ## Video4Linux Video Encode Device (h264_v4l2m2m)
+      # - "/dev/video11:/dev/video11"
+    volumes:
+      ## Example:
+      ## - "/host/folder:/container/folder"
+      - "/etc/localtime:/etc/localtime:ro" ## use local time from host
+      - "/etc/timezone:/etc/timezone:ro"   ## use timezone from host
+      ## Uncomment the next line if PHOTOVIEW_DATABASE_DRIVER is set to `sqlite` in the .env
+      # - "${HOST_PHOTOVIW_LOCATION}/database:/home/photoview/database"
```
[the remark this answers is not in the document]
kkovaletp: 1 small typo, made at the beginning, leads to the whole project being affected because of copy-paste))

kkovaletp:
Thanks for your detailed feedback! I'll try to explain my vision)

> What are the expectations of the user and what is the workflow you desire from your users as well as what are you willing to make 'official' and thus support users with?

From my understanding and the experience of supporting the project, I'd say that the typical representative of the target audience is a user, who has a media collection on an FS of home server and needs an easy self-hosted service to share it to home devices and optionally with external users. So, a list of minimal requirements for a typical user from the target audience of the product would be:
- Basic understanding of Linux and Docker
- Ability to get the system on their home server to work in server mode with Docker running self-hosted services using docker-compose files to define them
- Ability to install new software on the server, as needed (in most cases is included in the previous one)
- Ability to read and understand the Readme and inline documentation of the configuration files in the project
- Ability to follow the instructions from the documentation, mentioned in the previous item
- If something is not clear in the documentation, or it doesn't work in the desired way on a supported system and environment, it is expected from the user to ask for support or submit an issue with a detailed description of the problem and the environment

> to learn make

I saw this phrase several times from at least 2 people in the scope of this PR discussion, and, actually, don't understand the point: taking into account the scope of Makefile usage and the list of "make" features being utilized, do you really think that someone should really _learn make_? I'm not an expert in "make" and usually do not understand many things in Makefiles of Linux packages, intended to be built locally from sources, but here "make" is used as a way to define subsets of shell commands per use case, so when user runs `make use_case`, the corresponding commands executed in the predefined order. For me, the syntax and structure of Makefile with such definitions is clear and straightforward to anyone, familiar with Bash and docker-compose. Please explain me your point of view on this issue to let me understand your point)
But anyway, if you and @jordy2254 think that this is the wrong way, I'll transform the Makefile into a set of SH scripts, as mentioned earlier. And sure, I'm ready to support the Makefile or those scripts, depending on what would be the accepted solution.

> If they don't know Docker, I want them to learn that as opposed to make

Learning a new tool (especially such a widely used one) is always good, however, we're not the educational project. The product is built to solve a particular set of user's needs, so from my perspective, as simpler and straightforward our product is to set up and maintain, as more beneficial it is for the end user, who is not going to focus on the technical side of the project (if doesn't want to)

DaCHack:
Hey guys,
I know my contribution to this project is Limited and your technical expertise is much more Advanced.

Just thought it might be of value to share my use case from an end user perspective:
I run all my selfhosted services in docker and with limited time I value an out-of-the-box initial setup experience (putting the fun of tinkering with the tools and tailoring their setup to my needs afterwards aside).

All my containers are fully Set up and configured via Portainer. I want to add a compose script to a new stack and Click on deploy - done. I do not want to connect to my Server first, download files, run scripts (sh or makefile) and afterwards have troubles Managing the Stack in portainer because it was started outside.

i know that this approach is not super professional, but it is convenient since I am not a developer and manage my services in very limited free time.

Hope this helps with the solution to be found.

Omar007:
>     * Basic understanding of Linux and Docker
>     * Ability to get the system on their home server to work in server mode with Docker running self-hosted services using docker-compose files to define them
>     * Ability to read and understand the Readme and inline documentation of the configuration files in the project
>     * Ability to follow the instructions from the documentation, mentioned in the previous item
>     * If something is not clear in the documentation, or it doesn't work in the desired way on a supported system and environment, it is expected from the user to ask for support or submit an issue with a detailed description of the problem and the environment

Based on these points, I would personally completely forgo any custom abstractions for sure (be it make or (posix) shell scripts).  
If the original 'get it up and running' docker compose command really was not sufficient to get someone checking these boxes going, I feel like just adding/linking information for some extra use-cases in the README as is done for some 'advanced' use cases should probably have sufficed.  

> and optionally with external users

(side note; arguably, if you expected them to optionally deal with external users, I would've kind of expected some of those advanced use-cases to be abstracted as well)



>     * Ability to install new software on the server, as needed (in most cases is included in the previous one)

It's this point that I even think would become irrelevant when forgoing the make abstractions (though portable posix scripts could also do so, these would still require/rely on shell access)  



> do you really think that someone should really _learn make_? I'm not an expert in "make" and usually do not understand many things in Makefiles of Linux packages, intended to be built locally from sources, but here "make" is used as a way to define subsets of shell commands per use case, so when user runs `make use_case`, the corresponding commands executed in the predefined order. For me, the syntax and structure of Makefile with such definitions is clear and straightforward to anyone, familiar with Bash and docker-compose. Please explain me your point of view on this issue to let me understand your point)

Learning starts at a different point for everyone. While you and me know of make and roughly know how to write a Makefile to some level or another, from my personal experiences, it would be the safest bet to assume most home users just trying to run a container wouldn't even know of make as being a piece of software, let alone what a Makefile is and how to deal with that.  
Our 'learning' is at a stage where we'd be talking about stuff inside the Makefile but their stage of learning starts immediately when they encounter the word/command `make`.  

This is also why I said I'd rather 'have them learn Docker as opposed to make'. If either command is basically new to them, I would want them to know that for use-case 'run', they can do `docker compose up` rather than `make start`; the next time they have a compose file, they can now re-use that knowledge and do that same thing. This doesn't work otherwise.  



> But anyway, if you and @jordy2254 think that this is the wrong way, I'll transform the Makefile into a set of SH scripts, as mentioned earlier. And sure, I'm ready to support the Makefile or those scripts, depending on what would be the accepted solution.

I've never said it's wrong. It's just a very subjective (and opinionated) change. It's not up to me to decide if that is 'right' (good) or 'wrong' (bad) as I don't have a stake in this part of things. Not as a user, developer, maintainer, owner, person providing support, etc.



> Learning a new tool (especially such a widely used one) is always good, however, we're not the educational project. The product is built to solve a particular set of user's needs, so from my perspective, as simpler and straightforward our product is to set up and maintain, as more beneficial it is for the end user, who is not going to focus on the technical side of the project (if doesn't want to)

I'm not saying it has to be an educational project or exercise but certain changes have certain consequences, whether you want them or not. You're relying on the assumption that adding make (or custom scripts) on top of standard tooling makes things simpler and more straight forward but does it really? Or does it _for you_.  

In this case, you're introducing a new tool (make) to the project to create custom abstractions on top of standard tooling and you expect users to use that as their starting point. The consequence of that is that you'll have to account for everyone in the team to become (to some degree) responsible for and familiar with make, the contents of the Makefile, your users to not just ask for help with Docker things but also on the nuances and set up of your custom abstractions and the relation between it and what happens in Docker.  



Again, I do not have nor can provide a final answer. In the end, it's up to the project owners and maintainers who will have to deal with this. If you are going to take on that responsibility (and not just for the coming few weeks), I don't see why I should have a problem with it (be it a Makefile, posix scripts or heck, a python script or something funky like that); in the end it does not affect me so I can not in good conscience call this right/good/beneficial or wrong/bad/detrimental. All I can say is, that on any point in my path to where I am now, I would personally also have much more preferred to just have the Docker command to get it up and running.

kkovaletp:
> Hey guys, I know my contribution to this project is Limited and your technical expertise is much more Advanced.
> 
> Just thought it might be of value to share my use case from an end user perspective: I run all my selfhosted services in docker and with limited time I value an out-of-the-box initial setup experience (putting the fun of tinkering with the tools and tailoring their setup to my needs afterwards aside).
> 
> All my containers are fully Set up and configured via Portainer. I want to add a compose script to a new stack and Click on deploy - done. I do not want to connect to my Server first, download files, run scripts (sh or makefile) and afterwards have troubles Managing the Stack in portainer because it was started outside.
> 
> i know that this approach is not super professional, but it is convenient since I am not a developer and manage my services in very limited free time.
> 
> Hope this helps with the solution to be found.

@DaCHack, thanks for your feedback and description of your use case.

As mentioned in the Readme, Makefile (or future SH scripts) is optional and you're free to not use this. As far as I remember, Portainer allows to register `.env` file in the same way as the compose file, so you don't need to copy anything to your server in this case.

However, (again from the time when I tried Portainer) it manages compose services in a pretty strange way, adding some limitations. So, I'd say that for now, we do not support Portainer (as well as other similar wrappers on top of Docker).

This means that you still can use anything you want on your server, but if things go wrong because of the Portainer's impact, we cannot help you.

BTW, check Dockge - we don't support it as well, but from my PoV it is better than Portainer and can manage services, started outside of it.

jordy2254:
@kkovaletp @Omar007 

> Portainer's impact, we cannot help you.

I've seen you say this a few time now, who's_ decided this? Supporting common docker management tools seems like something we should do, granted I don't use it personally, but spinning up a sample docker compose works fine... seems a bit of a weird stance to take, especially without any input from anywhere else?

> Readme, Makefile (or future SH scripts) is optional and you're free to not use this.

Overall I'm still failing to see the benefit or any real value from this. Maybe I'm missing something, maybe I have some blinder's on, or maybe it's not been described correctly I can't say.

But if someone is self hosting photoview, I fail to see why they wouldn't  at the very least have a basic understanding of docker and how to use it the scripts (or whatever) that just invoke docker are just adding something else with from what I can still tell no real use case for it bar someone who has no knowledge of the tool they are using to host it, they'd still need to install docker and get it configured correctly. just some food for thought below on when someone has an issue
- Who do you intend to support user's when they want to adapt the script?
- Who do you intend to make changes to the docker compose file when a user wants something else that's not in there?
- Who do you intend to support user's when something is wrong with there environment?
- What about when a user can't install docker themselves?
- What about when a user has issues with their docker install?

Again I'm not against a make file or scripts if it's for a good reason, one such reason would be `Bare Metal` installs where someone wants to not host with docker but compile and host from source directly. Again this will also need some technical expertise such as the setup of the go env, ability to diagnose build issues from missing dependencies on their system.

Finally to finish this, I don't think I've ever seen any self hosted project utilise a make file or shell scripts in this way. Unless of course you can point to any?

kkovaletp:
> I've seen you say this a few time now, who's_ decided this? Supporting common docker management tools seems like something we should do

You have a strong position to not provide any wrapper on top of the Docker compose commands, but now you'd like to support Portainer and other similar wrappers. It doesn't look to me as a consistent position, so please explain.

> I don't use it personally, but spinning up a sample docker compose works fine... seems a bit of a weird stance to take, especially without any input from anywhere else?

I've tested Portainer for several complex docker-compose setups, and based on my testing results I made the conclusion that it isn't transparent in terms of a service's behavior: 
- it adds additional limitations to docker-compose usage scenarios
- services, running through Portainer, do not always behave the same as running through vanilla compose command
- it doesn't add much value in the case of managing a host with Docker only, its power is in managing self-hosted Kubernetes environments or cloud deployments.
That's why I do not want to support it in the scope of this project. However, as I wrote, it doesn't mean that Photoview users cannot use Portainer. They just should be ready to maintain all the differences between the vanilla docker by themselves.

> Overall I'm still failing to see the benefit or any real value from this.

Ok, let me try to explain the main benefits 1 more time:
- Imagine that you're not an experienced Docker user. You can follow the provided instructions to set a containerized app up, and you can search for some management commands in the docker documentation, trying to compile an appropriate command + its params to fix some issue, but you don't want to search for the command and remember it every time, you need to start \ stop \ backup the app - in this case having a list of short and logical commands looks better
- Please look into the `backup` target's execution flow in the currently committed Makefile. Assuming that you'd like to have periodic backups of the app's data according to best practices, what would you prefer: having a script to handle everything inside, so you can simply call it and have a new backup created, or running all those commands manually every time and making sure that you didn't forget or miss something? Other targets are way simpler, but some of them contain some additional flags and other useful improvements, which should be found by each user individually if there is no such wrapper.

> * Who do you intend to support user's when they want to adapt the script?

As the script is optional and provided as a template, I'd expect users with more experience to modify it according to their needs (if they still prefer it over plain docker commands). And if there is something commonly useful or a bug, it should be tracked in the same way as other issues here (new issue + PR).

> * Who do you intend to make changes to the docker compose file when a user wants something else that's not in there?

How is it related to the existence of a script? Anyway, the same as now: issue + PR but anyone from the community. I'm going to review them and contribute fixes, as the main contributor of the enhanced compose setup.

> * Who do you intend to support user's when something is wrong with there environment?

If they follow our recommendations, I'd expect the community and us, as the maintainers, to provide the support. If they use something specific, like Portainer, or host the app on some not supported environment, it would depend on the case and whether there is someone in the community with the knowledge and experience to help. If not then we would decline such a request, as not a supported one.

> * What about when a user can't install docker themselves?

As I mentioned earlier, I'd expect some set of minimal requirements from a user, including the ability to do a basic host setup and install all the dependencies of the app corresponding to the deployment scenario of choice.

> * What about when a user has issues with their docker install?

They should contact Docker support to solve those issues as a prerequisite to starting a support case here.

> I'm not against a make file or scripts if it's for a good reason, one such reason would be `Bare Metal` installs where someone wants to not host with docker but compile and host from source directly.

This is a good idea, but it is not in the scope of this PR, as well as it should be prepared by someone, experienced in building the code for hosting the app directly on the host, which is not my case.

> I don't think I've ever seen any self hosted project utilise a make file or shell scripts in this way. Unless of course you can point to any?

I saw several. Some of them were closed-source projects, so I can't point you to them. I don't remember right now any open-source projects to quickly post here the corresponding links. But anyway, if something isn't a common practice, it doesn't mean that this is not useful.

kkovaletp:
I think that I've done all the planned changes and action items for this PR.
@jordy2254, let's finalize our discussion regarding the Makefile \ scripts, so I can make the necessary changes and finally switch the PR to Ready state.

jordy2254:
@kkovaletp I think the makefile/scripts thing is something we're both seeing from opposite ends and won't necessarily come to a final agreement on. Maybe based on your examples... documentation improvements providing snippets may be a good middle ground? And potentially more visible than a make file or script to the community?

Ultimatly although were both maintaining I think the final decision on that segment should come down to @viktorstrate on his opinions as he is the repo owner.

kkovaletp:
Ok, then I'll switch it to Ready state (as it is ready from my PoV) and all remaining changes will be tracked through code review requests

kkovaletp:
This is a big PR with a lot of changes, so here is the proposed code review path to make it as clear and straightforward, as possible:
Readme > Dockerfile > .env > docker-compose > Makefile - in this way (I hope) you'll better understand the idea behind and the implementation.

docker-compose example/docker-compose.example.yml
```diff
@@ -0,0 +1,166 @@
...
+      # - "/dev/dri:/dev/dri"
+      ## Nvidia CUDA
+      # - "/dev/nvidia0:/dev/nvidia0"
+      # - "/dev/nvidiactl:/dev/nvidiactl"
+      # - "/dev/nvidia-modeset:/dev/nvidia-modeset"
+      # - "/dev/nvidia-nvswitchctl:/dev/nvidia-nvswitchctl"
+      # - "/dev/nvidia-uvm:/dev/nvidia-uvm"
+      # - "/dev/nvidia-uvm-tools:/dev/nvidia-uvm-tools"
+      ## Video4Linux Video Encode Device (h264_v4l2m2m)
+      # - "/dev/video11:/dev/video11"
```
viktorstrate, line 58: Should we add this to the documentation website?

viktorstrate (changes requested):
I have added some comments in the code.

Generally I think that the example is fairly nice. I don't have strong feelings on the Makefile, I think it's fine to have it as it can provide convenience to the user and they can choose to use it if they want. As long as the Makefile is not required to perform some actions and _only_ for convenience. I'm maybe thinking about backups here.

My major concern is that the example is very complicated (because it does a lot of things). And it is also very opinionated. Maybe not everyone wants to use Watchtower or they have another solution that they want to use, the same goes for backups.

I think the old `docker-compose.example.yml` which only had the bare minimum to get Photoview up and running is much easier to approach for someone who might be new to Docker or wants to try Photoview for the first time.

One solution might be to keep the old and simple `docker-compose.example.yml`, but provide a fully featured and opinionated setup along with it, maybe we could add a README file to the example directory to explain how it works? Then we could add all the nice features that we want, including backup etc. without worrying about complexity for users who want to have their own setup.

I don't think we need to introduce breaking changes in this pull request, as all the changes don't need it or could easily be implemented in a way that doesn't.

Dockerfile
```diff
@@ -20,84 +20,81 @@ ARG COMMIT_SHA
...
-COPY ui/package*.json /app/
-RUN npm ci --omit=dev --ignore-scripts
-
-# Build frontend
 COPY ui /app
-RUN npm run build -- --base=$UI_PUBLIC_URL
+WORKDIR /app
+RUN npm ci --omit=dev --ignore-scripts \
+  # Build frontend
+  && npm run build -- --base=$UI_PUBLIC_URL
```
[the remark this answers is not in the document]
kkovaletp: No, we've discussed this topic in the Discord: https://discord.com/channels/794576839813234688/1223295288668717146/1229127778679390319

Here is my answer for transparency:
Regarding layers: building production image on CI won't benefit much (if at all) from build cache, as in our project merges to master happen not so often. Having a well-optimized image for users is much better, as it pulls faster and uses less space on the host. It is even more important for ARM-based systems, where usually system partition is limited in space and performance. 
In the case of the PR testing CI job, I'm not sure how exactly the docker cache works in Github, but from my experience with other CI/CD services, the docker cache is related to the branch, so every new PR will start a new thread of cache and won't utilize the cache from other PRs or master, so here there are no benefits from having more layers too.

Dockerfile
```diff
@@ -20,84 +20,81 @@ ARG COMMIT_SHA
...
 
-COPY --from=ui /app/dist /ui
+WORKDIR /home/photoview
+COPY api/data /app/data
+COPY --from=ui /app/dist /app/ui
 COPY --from=api /app/photoview /app/photoview
 
 ENV PHOTOVIEW_LISTEN_IP 127.0.0.1
-ENV PHOTOVIEW_LISTEN_PORT 80
+ENV PHOTOVIEW_LISTEN_PORT 8080
```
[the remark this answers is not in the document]
kkovaletp: it is required by the "run as non-root user" improvement. We discussed this here: https://github.com/photoview/photoview/pull/863#discussion_r1576311663

docker-compose example/docker-compose.example.yml
kkovaletp: I can if you'd want it to be there, but I don't see a benefit: the compose file is read every time the configuration is done, it is self-descriptive with the comments (as to me), putting it in the documentation creates a duplicate of the info with no additional value

Dockerfile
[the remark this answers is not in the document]
viktorstrate: Are you sure that this is the case for Docker environments as well?
I just found this post: https://forums.docker.com/t/non-root-user-able-to-bind-to-port-80-why/140524

Maybe we should make an issue keeping track of the changes that we want to make for version 3.0?

kkovaletp:
> My major concern is that the example is very complicated (because it does a lot of things).

Yes, this PR is 1 year old and includes fixes for many reported (and some not reported) issues in the project. It is not possible to fix all of them with a simpler file. Taking comments out of it is not an option, as users might not understand the logic and will make many configuration mistakes without the in-line explanations.

> Maybe not everyone wants to use Watchtower or they have another solution that they want to use, the same goes for backups.

Sure, that is why Watchtower is marked as optional. The reason, why it is enabled by default, is to encourage users to use it and have always the latest images running. This will decrease the potential # of issues, submitted for previous builds, while there is a new build with the fix. However, if you see more benefits in getting it disabled (commented out) by default, please share your thoughts.

Backups are configured in the Makefile, which is optional. There is only a variable in the `.env`, which can be safely ignored as long as the Makefile's backup target is not called. I can add this to the comment of the var in the `.env`

> I think the old `docker-compose.example.yml` which only had the bare minimum to get Photoview up and running is much easier to approach for someone who might be new to Docker or wants to try Photoview for the first time. One solution might be to keep the old and simple docker-compose.example.yml, but provide a fully featured and opinionated setup along with it

Yes, it is. But also, it doesn't contain many optimizations and fixes, so keeping it in the repo would create a false impression that both are equivalent from a fixes perspective and only the difference in advanced configuration with more features provided in this file, while it is not the case.

> maybe we could add a README file to the example directory to explain how it works?

Now this Readme is included into the config files in the form of in-line comments, which makes these configs self-descriptive and straightforward to configure, while following the configuration path from the project's Readme. Extracting the comments to the external Readme wouldn't simplify the configs, but add more complexity to users, as they would need to constantly switch between the config and the Readme to understand how to modify it according to their cases.

> I don't think we need to introduce breaking changes in this pull request, as all the changes don't need it or could easily be implemented in a way that doesn't.

1st of all, please let me know, what do you mean by "breaking changes"? I communicated with engineers, who understand it a bit differently. 
You mentioned the change of the internal port inside the container, as a breaking change. What does it break exactly in the case of using the correct configs on the host (from this PR)? Another breaking change might be the requirement to use configs from this PR to make the new image work fine. And this is the only way possible to provide the fixes, implemented in the PR. The configs from the current `master` cannot be used in any case.

Dockerfile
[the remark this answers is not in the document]
kkovaletp: We discussed this as part of the @Omar007 port change proposal. While I can confirm that the recent version of Docker allows using 80 port by a non-root user inside a container, it is not clear, starting from which version of Docker this change is introduced.
I can assume that some Linux distros or devices with proprietary Linux distros might still provide an older Docker version, where this is not yet implemented.
So, making sure that Photoview works according to Linux best practices on all supported platforms, looks more beneficial than keeping port 80 with the hope that it will work correctly everywhere.
I don't see a negative impact of switching the port number. Please share your thoughts if there is something negative, related to this change.

kkovaletp:
And thanks a lot for your review and the feedback provided, @viktorstrate )
I got straight to the point and forgot to type this(

Dockerfile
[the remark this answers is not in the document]
Omar007: I see @kkovaletp covered the reasoning already but @viktorstrate is also technically correct in that it can be 'breaking'.  
If a user had been running as root or granted the required capabilities (or had been on a Docker version greater than ...?) and has never explicitly set `PHOTOVIEW_LISTEN_PORT` themselves at any point and where to just swap out the image, it would be 'breaking' in the sense that their port-forwards/mapping would stop working if they don't explicitly set `PHOTOVIEW_LISTEN_PORT=80` when they update.  

Since this is a defaults change but also one that is backward compatible (it does not require these users to change assigned permissions or capabilites; they could start dropping these now but they don't need to), I wouldn't say this would require a major version bump, postponing to a major release or anything like that, but I would at the very least make sure this is mentioned on w/e the tag ends up being.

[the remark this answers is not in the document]
kkovaletp: Thanks @Omar007 for the explanation!
If we review the case, where a user updates an image, but doesn't update the compose and .env files (which doesn't make much sense from fixes and improvements delivery, but can happen in reality) nothing is going to break, as the compose file from the current `master` contains the `PHOTOVIEW_LISTEN_PORT=80`: https://github.com/photoview/photoview/blob/4133694bc23d53db3fa484d04283e75a57ef00db/docker-compose.example.yml#L27
Of course, if this line was removed or commented out, it wouldn't work, but that is not the recommended config.

BTW, while answering this comment, I noticed the hardcoded port in the `EXPOSE` and `HEALTHCHECK`, so I replaced it with the variable, then tested that it works fine with the `8080` and `80` values (setting `80` in the compose file for the same image)

kkovaletp:
Based on our recent discussion I set the `photoview` image tag to `2` - the current major version. This will prevent auto-updates to the next major version and enhance stability.
If this PR merge is going to trigger an increment of the major version, another PR with consistent version updates across the code would be required, including the tag in the compose file example as well.

kkovaletp:
According to the discussion in the product's Discord about this PR and the future product development strategy in general, it looks like this PR is going to be merged in some undefined future (if it will be at all).
It fixes A LOT of issues, so I assume that many users are waiting for it. If you are one of them, I'd recommend you to check my fork: [https://github.com/kkovaletp/photoview](https://github.com/kkovaletp/photoview) - it already contains all the fixes and improvements and will get more in the future

viktorstrate:
I really like the changes proposed in this PR, and I'm really sad to hear that this is getting dragged out so much and you start losing motivation. For me this change would be awesome if two things was changed:

1. The port number was changed back to `80`, but keep everything else (eg. running non-root). And then expect users to use an up to date version of Docker.
2. The original bare minimum `docker-compose.example.yml` is added back in the root directory, for those who want to configure it themselves.

kkovaletp:
@viktorstrate, thanks for your feedback and the summary of your expectations. Here are my short comments on them (the long and detailed explanations are here in the thread or the Discord - I don't remember already):
1. While it is not the best solution according to the Linux best practices, it works fine, so I can do this change
2. I disagree with this, as the original compose file is buggy and cannot provide the same fixes as the new one. So, by providing both we'd create a huge misunderstanding for users and a false impression that they are equal at least from the quality level, and differ in functionality only, while it is not the case.

viktorstrate:
Can you elaborate and help me understand what things would cause misunderstandings for users?

I like to have a bare minimum unopinionated example for users that is quick and easy to understand, and the let the users chose.

kkovaletp:
@viktorstrate, instead of starting another cycle of the same discussion, I've created the `minimal` compose file using the new file as a source and removing (almost) everything optional from it. Also, I've updated the readme and mentioned it in the docker section.
Please check my last commit to this PR to find these changes.
Is this what you expected to get or do you still prefer the original compose file?

viktorstrate (approved):
It looks good to me now, thank you for making the changes :)

kkovaletp:
I've found some minor bugs in config and will push a fixing commit, so please don't merge yet

kkovaletp:
I committed the fixes for the bugs, found y-day. Feel free to merge the PR.

viktorstrate:
Thank you! I will merge this one just to move forward.

A lot of changes has been made but nothing breaking so if we find issues later on, let's just handle and fix it then. If someone still have concerns with the changes here, feel free to open a new issue or start a discussion on Discord.

PR #874, opened 2023-08-10
Fix UI not serving media and tests expecting localhost:3000 as base URL

davidecavestro:
It started as just an attempt to fix failing tests, emulating what done at https://github.com/photoview/photoview/blob/master/ui/src/components/photoGallery/presentView/PresentMedia.test.tsx#L60

But then I started the dev UI and seen that media got 404, since their url are prepended with the UI domain, then stripping the double slash of the orignal url protocol.

Probably I'm wrong as I ignore the intent of the original code - i.e. proxying media at the client side? - but this way it works for me


kkovaletp:
I've merged this PR into #910 - a temporary alternative to the master while the project is on hold, so users still could get the latest Photoview improvements and fixes by building an image locally from this PR code

viktorstrate (approved):
Looks good to me, thank you for submitting the PR.

Issue #387, opened 2021-05-31
More robust handling of video processing


bobobo1618:
**Is your feature request related to a problem? Please describe.**
When I import a tree that includes videos, photoview spawns CPU-intensive `ffmpeg` processes of the form:

```
/usr/bin/ffmpeg -i <input> -vcodec h264 -acodec aac -vf scale='min(1080,iw)':'min(1080,ih)':force_original_aspect_ratio=decrease /app/cache/<output>.mp4
```

_Most_ of the videos included in my tree are already encoded using web-compatible codecs, so this CPU-intensive transcode is unnecessary, the desired result could be achieved by remuxing the the video, potentially on the fly, which is a much lighter process.

Additionally, [the current logic](https://github.com/photoview/photoview/blob/b8196cdd6763f1094c578255f43915296855c1e4/api/scanner/media_type/media_type.go#L136-L141) about whether a video is web-compatible is built on MIME types, which are container based and not robust. An MP4 container can, for example, contain a variety of codecs that are incompatible with web browsers.

**Describe the solution you'd like**
Break video treatment into three classes, from the current two:
- Leave as-is
- Remux
- Transcode

Information about the codecs and container can be collected with `ffprobe` or `mediainfo`.

This decision flow should be:
- Does the file contain web-compatible video and audio?
  - No: Transcode
  - Yes
    - Is the file in a web-compatible container?
      - Yes: As-is
      - No: Remux

**Describe alternatives you've considered**
If possible, it would be ideal if remux could be performed on-the-fly, as this would eliminate the need for extra disk-space. It would however inhibit, without introducing some kind of complex indexing mechanism, the ability to seek through video.

Another alternative would be to offer an option that disables transcoding entirely.

viktorstrate:
Thank you for the detailed issue.

This sounds like a solid and possible solution. I unfortunateley know very little about the internals of video formats and I'd have to read up on remuxing to understand the exact possibilities and potential solutions.

Even without that, I think there might be a lot of formats that are directly web compatible that are not currently detected as such. And using `ffprobe` (which is already used to gather other metadata) could potentially be a better solution than simply looking at the file extension.

> Another alternative would be to offer an option that disables transcoding entirely.

#279 would allow to disable transcoding entirely

kkovaletp:
This ticket is partially implemented. Let's double-check that there are no blind spots and missing/unhandled cases before closing it.

Issue #674, opened 2022-04-11
Ability to download the original video


fmnamado:
**Describe the solution you'd like**

Hello,
I can download the web optimized video directly from Photoview.
Is it possible to add the possibility of download the original/high quality version also from Photoview?

viktorstrate:
That is not possible, but it would be a nice addition.

rdhir001:
Yes: it seems that small 4K-HDR size videos will play very well - I am having difficulty with large size files.

What do you mean by 'download'? - or do you mean viewing the file on your TV?

... In Partnership and Service, 

rdhir001

fmnamado:
No, download the file.
When I am in a PC, in a browser, for photos I have the option of download the thumbnail, low rez, or ORIGINAL PHOTO.
In video I just have the option of downloading the "web optimized video" 

is is not very important, but if I have the option of downloading the original in photos, why not in videos also?

rdhir001:
OK - then why not download original video from Google Photos where the original video resides?

Am I missing anything?

... In Partnership and Service,

rdhir001

rdhir001:
Sorry - I thought you had PhotoView App for Smart TV:

***

https://www.photoviewapp.com

***

Regards.

... In Partnership and Service,

rdhir001

fmnamado:
No problem but I would say you are on the wrong github?

This github is about photoview self hosted app as you can see here https://github.com/photoview and here https://photoview.github.io/

how you get up to this thread? :)

rdhir001:
Being new to github, I first asked for help on the wrong thread.

Well, it was good to converse with you - you are thorough, and very responsive.

E . N . J . O . Y! a joyful life.

Regards.

~ rdhir001

fmnamado:
oh OK no problem, I found out strange is this is a thread I began to ask for help, and you were asnwering or asking other thing.

No problem, hope you have also a good life :)

DaCHack:
Funny thread here :laughing: 

I would like to +1 this feature request.
I have my files on a SMB-NAS that is mounted in the photoview container so I like to have only the converted files in usual access without the delay to my NAS, but if I select a single file or also a full album to be downloaded it would be good to have the option to go back to the originals and download them.

Same is true for videos that I want to show on a big screen: Once selected through the web interface I would love to play the original (being linked to and streamed from the NAS then would be fine).

kkovaletp:
[...]

PR #886, opened 2023-09-02
recode videos incompatible with popular browsers

### LINKED ISSUE #606: Videos taken on Pixel 6 Pro using Google camera doesn't play on chrome and firefox
**Describe the bug**
Videos taken from Google camera on Pixel 6 Pro doesn't play. I can hear the audio but the image is stuck as the initial frame.
And the issue seem to be with chrome and Firefox. I can play the videos fine in safari.

**To Reproduce**
1. Take a video from pixel 6 Pro using gcam
2. Sync the video to photoview
3. Try playing it from photoview

**Expected behavior**
Able to play videos on all browsers

**Screenshots**

**Your environment:**

 - Are you using Docker? [yes, docker compose
 - Database latest
 - Browser [e.g. chrome, firefox]
 - Version latest

**Additional context**
None

kabakaev:
This PR fixes #606 and partly #387, except a remux of existing video stream.

@bobobo1618, can you give an example of video files worth remuxing?

My video archive has only a handful of records from a digital video camera, which saved H.264 video and AC3 audio in MPEGTS container. But even there, I'd prefer to spend some CPU time to re-encode that hardware-encoded H.264 stream through libx264 to avoid artifacts, improve compression (Main vs. High profile) and to validate the stream.

Another possible use case would be to remux videos in order to move the MP4 `moov` atom in front of `mdat` for faster start of video playback. But I guess it's better to prepare such a remux on-the-fly instead of on-disk caching.

Anyone willing to make use of this PR should remove the "original" media_urls and then press `Scan` button for the affected users.
For example, H.264-encoding of all google pixel videos can be triggered by this SQL statement:
1. Make backup of the database
2. `delete from media_urls where media_name LIKE 'PXL%.mp4' AND purpose=='original';`

A combined code of this PR and #883 is built and pushed to container registry at docker.io/kabakaev/photoview:recode-incompatible-video

@mawi675lr81, @ouafnico, @CorneliousJD, would you like to test your video files with this fixed version (avoid running on production though, or make a backup)?

The combined code is pushed at https://github.com/kabakaev/photoview/tree/recode-incompatible-video

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,15 +74,12 @@ WORKDIR /app
 COPY api/data /app/data
 
 RUN apt update \
-  # Required dependencies
-  && apt install -y curl gpg libdlib19.1 ffmpeg exiftool libheif1
-
-# Install Darktable if building for a supported architecture
-RUN if [ "${TARGETPLATFORM}" = "linux/amd64" ] || [ "${TARGETPLATFORM}" = "linux/arm64" ]; then \
-  apt install -y darktable; fi
-
-# Remove build dependencies and cleanup
-RUN apt purge -y gpg \
+  `# Required dependencies` \
+  && apt install -y curl gpg libdlib19.1 ffmpeg exiftool libheif1 \
+  `# Install Darktable if building for a supported architecture` \
+  && if [ "${TARGETPLATFORM}" = "linux/amd64" ] || [ "${TARGETPLATFORM}" = "linux/arm64" ]; then \
+    apt install -y darktable; fi \
+  && apt purge -y gpg \
   && apt autoremove -y \
   && apt clean \
   && rm -rf /var/lib/apt/lists/*
--- a/api/scanner/scanner_tasks/processing_tasks/process_video_task.go
+++ b/api/scanner/scanner_tasks/processing_tasks/process_video_task.go
@@ -55,75 +55,97 @@ func (t ProcessVideoTask) ProcessMedia(ctx scanner_task.TaskContext, mediaData *
 		return []*models.MediaURL{}, errors.Wrap(err, "error getting video content type")
 	}
 
-	if videoOriginalURL == nil && videoType.IsWebCompatible() {
-		origVideoPath := video.Path
-		videoMediaName := generateUniqueMediaName(video.Path)
-
-		webMetadata, err := ReadVideoStreamMetadata(origVideoPath)
-		if err != nil {
-			return []*models.MediaURL{}, errors.Wrapf(err, "failed to read metadata for original video (%s)", video.Title)
-		}
-
-		fileStats, err := os.Stat(origVideoPath)
-		if err != nil {
-			return []*models.MediaURL{}, errors.Wrap(err, "reading file stats of original video")
-		}
-
-		mediaURL := models.MediaURL{
-			MediaID:     video.ID,
-			MediaName:   videoMediaName,
-			Width:       webMetadata.Width,
-			Height:      webMetadata.Height,
-			Purpose:     models.MediaOriginal,
-			ContentType: string(*videoType),
-			FileSize:    fileStats.Size(),
-		}
-
-		if err := ctx.GetDB().Create(&mediaURL).Error; err != nil {
-			return []*models.MediaURL{}, errors.Wrapf(err, "insert original video into database (%s)", video.Title)
-		}
-
-		updatedURLs = append(updatedURLs, &mediaURL)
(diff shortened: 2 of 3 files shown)


kkovaletp:
@kabakaev Why not to include the mentioned SQL query as the migration code, affecting all original videos? In this case, when this PR is merged, the fix will be automatically applied to all user's envs after the next rescan. 

A really small amount of users actually read the "What's new" of a release, so if some manual steps are required for the fix, in most cases that won't be performed

kabakaev:
This may be a good idea. 
Unfortunately, the given SQL statement is not idempotent. In other words, it can only be run once.
ALso, it should not delete the "original" media_urls, which are already compatible with web browsers, such as MP4 files with H.264 and AAC codecs.

We may call `ffprobe` for every scanned video file, but it will make the scan job painfully slow. The only viable and performant option would be to extend the `media_urls` table with videoCodecs, audioCodecs names and container options. If we fill them once, then future improvements may be easier to implement. For example, on-the-fly recode of MP4 container for fast start would be fast enough.
@viktorstrate, how do you think, is it possible to extend the `media_urls` with those 3 fields?
Alternatively, we may create two new tables: a `media_codecs` table listing all known codecs and a `media_urls_codecs` many-to-many table joining `media_urls` and  `media_codecs`.

kkovaletp:
@kabakaev As the query is not idempotent, it should be run inside the transaction, so we can be sure that nothing is corrupted in case of some interruption or failure.

As far as I see from the DB structure, there is no need to extend the `media_urls` table with additional fields or create new tables, as we already have the `video_metadata` table, whose purpose is to contain such data. Please check whether it is, and if not, we potentially need to enhance or extend the data in this table.

According to the schema, the connection between the `media_urls` and `video_metadata` tables can be done through the `media` table.

```SQL
select * 
from media_urls mu 
	join media m on mu.media_id = m.id 
	JOIN video_metadata vm on vm.id = m.video_metadata_id 
where mu.purpose = 'original';
```

Anyway, analyzing potential "damage", I don't see a big problem with removing more "original" media_urls than technically needed: the next scan will add them back spending not so much additional time and resources. It will be definitely faster than running `ffprobe` for each video file

kkovaletp:
@kabakaev, I've created PR #910 - a temporary alternative to the master while the project is on hold. I'd like to merge this PR as well, but I don't believe users will run manual steps after pulling new code and rebuilding the image, so I'd ask you to prepare a migration for the DB without an extension for the `media_urls` table and using existing data and relations, as proposed in my previous comment. This migration should cover all the cases when this PR might be useful (remove videos of all formats that need to be rescanned).
Do we need to remove corresponding files from the cache as well? If original files are going to be rescanned, will that remove previously converted files or they will stay and just waste space?

viktorstrate:
The database can be automatically migrated to solve the problem of deleting the old records.

This is done in `api/database/database.go`

kkovaletp:
@jordy2254, what do you think about this PR? My questions are in the comment above

kevinfiol:
Any noise on this one? Videos recorded by Pixels seem to have issue playing.

kkovaletp:
> Any noise on this one? Videos recorded by Pixels seem to have issue playing.

it is still waiting in the queue(
if you're a developer and can help with the leftovers, feel free to fork, create a branch from this PR, implement the required changes, and open a new PR for this issue

kkovaletp:
@googollee, as you're working now on the Ffmpeg-related topic, can you please review this PR and provide your feedback: does it still add a value to the current product implementation?
If you see a benefit in merging this PR, can you please add the migration, mentioned in the comments? Or let me know your thoughts about the migration logic, proposed there

googollee:
I think converting videos to a proper size is still necessary, since some of users accessing them through the internet https://github.com/photoview/photoview/issues/1068. The original videos are pretty large, especially 4k is popular recently.

kkovaletp:
> converting videos to a proper size is still necessary,

Yes. I agree. The more important reason is not the size, but the combination of the container, video/audio codecs, and the compression params, expected by our player on the UI side.
We have another issue #674 with the request to provide access to the original media, where we could implement the ability to access the original media.

Apart from that, do you see some positive changes and improvements in this PR compared with the current codebase? Or we could safely close it

googollee:
No, I don't see such changes.

kkovaletp:
Ok, then I'm closing this PR as not planned.
Thanks for the contribution and the idea of this functionality, but currently it doesn't fit the product architecture.

PR #910, opened 2024-01-28
NOT FOR MERGE rebuild photoview image

kkovaletp:
This PR is not for merging.
It is created as a workaround for the fact that the project is on hold, so the image on the Docker Hub was built a long time ago, which leads to some bugs and vulnerabilities not being fixed in the base image and installed 3rd-party components.

With this PR anyone could easily rebuild the Photoview image locally based on the project source code and the latest available 3rd-party components according to the dependencies defined in the project.

See the README in this PR for the step-by-step guide. Here is the copy of the guide for your convenience:

> This section describes how to get Photoview up and running on your server with Docker.
> Make sure you have Docker and docker-compose installed and running on your server.
> `make` should be installed as well if you'd like to use provided `Makefile`, which is optional (see step 4 for more details).
> `7zz` should be installed in case, you'd like to use it in scope of the backup scenario instead of the default .tar.xz format. Read the comment in the `Makefile`, located in the `backup` section for more details.

1. Clone this repo to the folder on your server, where you expect to host the Photoview internal data (database and cache files).
   - `cd <path to the folder on your server, where you expect to host the Photoview internal data>`
   - `git clone https://github.com/kkovaletp/photoview.git .`
   - `git checkout NOT-FOR-MERGE-rebuild-photoview-image`
   - `git pull`
2. Rename `docker-compose.example.yml` and `example.env` files and remove the `example` from their names (so, you need to have `.env` and `docker-compose.yml` files).
3. Open these files and the `Makefile` in a text editor and read them. Modify where needed according to the documentation comments to properly match your setup. It is better to go through the files in the next order: `.env`, `docker-compose.yml`, and `Makefile`.
4. In case, you don't have `make` installed in your system or don't want to use it for the Photoview management activities, you could use the same commands from the `Makefile` and run them in your shell directly, or create your own scripts. Make sure to apply or replace the variables from your `.env` first in this case. `Makefile` is provided just for your convenience and simplicity, but is optional.
5. Start the server by running the following command (or corresponding sequence of commands from the `Makefile`):

```bash
make all
```

If the endpoint or the port hasn't been changed in the `docker-compose.yml` file, Photoview can now be accessed at http://localhost:8000

> NOTE:
> This sequence will build the Photoview image from local sources replacing the original
> `viktorstrate/photoview:latest` image, published on the docker hub by the repo owner.
> This is a workaround of the fact that the project is not actively maintained,
> and that image is pretty outdated, having some vulnerabilities and old software.
> If you want to use the original image, do not build the local one
> (exclude the `build` step from the sequence in the `Makefile`)
> and remove the local `viktorstrate/photoview:latest` image.
> Next time you start the service, the original image will be pulled from the docker hub.


kkovaletp:
I've just merged PR #874 to this branch to fix failing tests
I'm going to review and merge some other open PRs here to make this branch a better alternative to the `master`, containing the latest fixes and improvements until the repo owner returns and merges those PRs into the `master`.

kkovaletp:
BTW, this PR was based on #863, so it contains all the fixes from that PR too

DaCHack:
@kkovaletp Many thanks for the work you put into this!
Idea in favor of all non-dev users: Can't we agree in setting up a separate repository that does nothing else than syncing this rep, applying this PR and pushing a ready-to-go image to docker hub? This way there is no need for people to always build the image from scratch.

I just see that you are even able to create workflows in this rep to create the docker image. Why not pushing them directly to docker hub? (if needed on a separate account)

kkovaletp:
> I just see that you are even able to create workflows in this rep

No, these workflows and automatic checks were defined by the repo owner and run for each PR update independently to the PR owner

> Can't we agree in setting up a separate repository that does nothing else than syncing this rep, applying this PR and pushing a ready-to-go image to docker hub?

It is a good idea, which needs to be investigated. Right now I'm not sure that it will work end-2-end. Also, in any case, that would be another Docker Hub registry, not the one, created by the repo owner and mentioned in the README of the `master` branch - as nobody else has access to that registry to push images.

kkovaletp:
@DaCHack ,


> Can't we agree in setting up a separate repository that does nothing else than syncing this rep, applying this PR and pushing a ready-to-go image to docker hub?

I've tried to integrate the CI pipeline on my forked repo, but without success:
- I investigated the possibility of activating the GitHub Actions (which are disabled by default for forked repos), but found out that in this case, I'll end up with constantly failing pipelines for `master` and all other branches, as they have the config from the original repo. If I'll change that, my repo will be out of sync. Another issue is that I'll have to guess how the environment variables are configured for the CI in the original repo to try to replicate the same here. So, it looks like using the GitHub actions on a forked repo, while keeping it in sync with the original one isn't a good idea.
- Then I integrated this branch with CircleCI and configured it to build and push the docker images. However, once the pipeline was valid from the CircleCI perspective, I found out that for some reason CircleCI is unable to build docker images for other platforms except for amd64, so I'll need to maintain this CI pipeline, my Docker Hub registry, the Makefile to support both pull and build just to provide the and64 image. In contrast, all ARM-based ones still need to be self-built.

With this, I decided to revert all my changes in the branch, as I don't see enough benefits that are worth such an extra complex setup. Feel free to propose a simpler and working solution for building all the images on a CI and hosting them on the Docker Hub registry.

DaCHack:
Thanks for checking and putting the effort in this.
I'll think about it and get back as time allows.

timbroder:
Any ideas why I might be getting errors like this?

```
ERROR: Invalid interpolation format for "build" option in service "photoview": "$(date +%Y-%m-%d)"

ERROR: Invalid interpolation format for "build" option in service "photoview": "$(git rev-parse --short HEAD)"
```

I'm running `docker-compose version 1.29.2, build unknown`

kkovaletp:
> Any ideas why I might be getting errors like this?

strange... no ideas. I tested the build on kubuntu 22.04 (amd64) and everything was working fine.
please double-check that you strictly follow the steps from the readme or the 1st post in this PR.
`*BUILD_DATE` and `*COMMIT_SHA` build args in docker-compose are resolved from the Makefile's build command's build-args, where they are defined as sub-shell commands, so they are executed in bash right at the moment, the value is going to be provided to the main command parameters. Make sure that this chain is not broken and works correctly on your env.

docker-compose.example.yml
```diff
@@ -1,51 +1,133 @@
...
+        BUILDPLATFORM: linux/amd64
+        TARGETPLATFORM: linux/amd64
+        #REACT_APP_API_ENDPOINT: http://localhost:4001/
+        UI_PUBLIC_URL: /
+        VERSION: selfbuilt
+        REACT_APP_BUILD_VERSION: selfbuilt
+        BUILD_DATE: $(date +%Y-%m-%d)
+        REACT_APP_BUILD_DATE: $(date +%Y-%m-%d)
+        COMMIT_SHA: $(git rev-parse --short HEAD)
+        REACT_APP_BUILD_COMMIT_SHA: $(git rev-parse --short HEAD)
```
timbroder, line 20: This solved my issue [here](https://github.com/photoview/photoview/pull/910#issuecomment-1932482370). Thanks!

```suggestion
        BUILD_DATE: $$(date +%Y-%m-%d)
        REACT_APP_BUILD_DATE: $$(date +%Y-%m-%d)
        COMMIT_SHA: $$(git rev-parse --short HEAD)
        REACT_APP_BUILD_COMMIT_SHA: $$(git rev-parse --short HEAD)
```

kkovaletp: Thanks, I tested it and it works for me. Have no idea why it was working with 1 $ before...
I've pushed this change to the repo

kkovaletp:
I've merged here PR #917 with an improvement to thumbnail creation for video files.

kkovaletp:
I decided to add the `vulnerabilities` folder to this PR with the latest scan results from Anchore Syft and Grype of the source code and the built Docker image.

My main intention is to provide you with all the info about the product's security level, so you can self-host it with an adequate level of protection and isolation. But also, it would be great if these reports could trigger some new PRs from our community developers with updated dependencies to fix at least already patched ones.

Please note that the Grype report about the image contains components from the product, as well as from the base Docker image, which means that we can fix only the vulnerabilities from the product, while we rely on the base image maintainers to keep that image up-to-date and that is why I recommend to periodically rebuild your Photoview image even if there are no new updates here in the repo.

kkovaletp:
I've merged here PR #918, which adds the change password functionality to the settings page of non-admin accounts

kkovaletp:
As the project is back to active development, no new merges are expected in this PR.
We're going to merge all original PRs, which are part of this one, into the `master` and when it is done, so `master` contains all the same features and bugfixes, this PR will be closed. I'll add a new comment here before that

kkovaletp:
If you're still waiting for some fixes and improvements, not merged into the `master` compared to this PR, I'd recommend you to check my fork: [https://github.com/kkovaletp/photoview](https://github.com/kkovaletp/photoview) - it already contains all the fixes and improvements from this PR

PR #922, opened 2024-03-28
add ImageMagick Support

### LINKED ISSUE #923: add ImageMagick Support
Currently, support for `RAW` format is available via `Darktable`, but `Darktable` may not be installed or used on some machines, and an alternate `RAW` format support option is required. In the future, I think this way through external tools, can be selected at will through the configuration file, not limited to the current few tools.

Here, I want to add `ImageMagick `as another external tool that supports `RAW` format, and on my current machine (TrueNAS base FreeBSD 13.1), only that tool can be installed successfully.

Here is the PR I created: #922

fierceX:
Fixes #923 In the current `RAW` format support, which is supported by `Darktable`, but on some machines the package may not be installed, I added the `ImageMagick` package as a `Darktable` alternative to support the `RAW` format

@ kkovaletp I compiled the test on your branch.

### DIFF
--- a/api/scanner/media_encoding/encode_photo.go
+++ b/api/scanner/media_encoding/encode_photo.go
@@ -115,6 +115,11 @@ func (img *EncodeMediaData) EncodeHighRes(outputPath string) error {
 			if err != nil {
 				return err
 			}
+		} else if executable_worker.ImageMagickCli.IsInstalled(){
+			err := executable_worker.ImageMagickCli.EncodeJpeg(img.Media.Path, outputPath, 70)
+			if err != nil {
+				return err
+			}
 		} else {
 			return errors.New("could not convert photo as no RAW converter was found")
 		}
--- a/api/scanner/media_encoding/executable_worker/executable_worker.go
+++ b/api/scanner/media_encoding/executable_worker/executable_worker.go
@@ -15,11 +15,13 @@ import (
 
 func InitializeExecutableWorkers() {
 	DarktableCli = newDarktableWorker()
+	ImageMagickCli = newImageMagickWorker()
 	FfmpegCli = newFfmpegWorker()
 }
 
 var DarktableCli *DarktableWorker = nil
 var FfmpegCli *FfmpegWorker = nil
+var ImageMagickCli *ImageMagickWorker = nil
 
 type ExecutableWorker interface {
 	Path() string
@@ -30,6 +32,10 @@ type DarktableWorker struct {
 }
 
 type FfmpegWorker struct {
+	path string
+}
+
+type ImageMagickWorker struct {
 	path string
 }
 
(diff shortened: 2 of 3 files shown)


jordy2254:
Thanks for your PR, I've had a quick look and can't see any issue for this could you please start by raising an issue so we can link them together for easier search-ability in the future. As part of raising that issue it would be good if you could explain why having imagemagick in addition to darktable would be a good system to have as an alternative beyond someone might not have darktable installed.

Thanks again for your PR
Jordan

viktorstrate (approved):

kkovaletp:
This is a good idea to provide an alternative to the DarkTable for architectures, where it is not available.

@fierceX it would be great if you could do some research and provide a brief comparison results of these 2 products: 
- Maybe it makes sense to migrate from DarkTable to ImageMagick on all platforms? 
- What benefits and disadvantages do you see?
- Does ImageMagick provide all the DarkTable functionality that we use in Photoview?

Another question about the PR: I see the source code changes, but don't see the ImageMagick installation as part of the image build process. Do we already have it installed? I don't see a place, where it is installed into the image.

fierceX:
@kkovaletp I didn't add this external tool to Docker. I think if the original Darktable tool works well, then I don't need to install it. I added this tool to compile and install it. On my machine, Darktable cannot be installed.

kkovaletp:
> if the original Darktable tool works well, then I don't need to install it

@fierceX we have the next code in the dockerfile:
```docker
if [ "${TARGETPLATFORM}" = "linux/amd64" ] || [ "${TARGETPLATFORM}" = "linux/arm64" ]; then \
    apt install -y darktable; fi
```
So, for the platforms, where Darktable is installed, there is no need for your PR, while for other platforms your PR won't work because the ImageMagick tool is missing.
Of course, if you install Photoview to your system directly instead of using Docker, you should manage all the dependencies manually, but for Docker (as our main distribution option), we need to make sure that all the dependencies are included and the image is fully-functional for all supported platforms (with all limitations clearly described and documented)

fierceX:
@kkovaletp First of all, Docker is not the only way to deploy, and valid code that is not integrated into Docker also has value in being merged (I'm sure many people compile and install it directly). Secondly, if ImageMagick is not installed, it will not be an error, just like not installing Darktable, except that it cannot handle RAW files. ImageMagick and Darktable are not in conflict, but in substitution.

jordy2254:
@fierceX 
Thanks for raising an issue and explaining why this is needed it makes a lot more sense now. I'm wondering... should we be making this configurable for the event the user has both installed but wants to choose?

jordy2254:
Hey @fierceX Thanks again for the PR, I'm going to close this PR as it's now covered by your other request in #929 If you feel this is a mistake feel free to open another PR, if you need any help please don't hesitate to ask.

PR #927, opened 2024-04-01
feat(api): implement permissions system

jordy2254:


api/database/database.go
```diff
@@ -153,8 +154,13 @@ func SetupDatabase() (*gorm.DB, error) {
 	return db, nil
 }
 
-var database_models []interface{} = []interface{}{
+var database_models_1 []interface{} = []interface{}{
```
viktorstrate, line 157: I think the tests don't work since you make two database models and try to migrate them both. Why did you add a `database_models_1`?

api/database/migrations/permissions.go
```diff
@@ -0,0 +1,154 @@
...
+
+import (
+	"github.com/photoview/photoview/api/graphql/auth"
+	"github.com/photoview/photoview/api/graphql/models"
+	"github.com/pkg/errors"
+	"gorm.io/gorm"
+)
+
+// MigrateForPermissions Migrates new permissions into the system and upgrades system roles according to the new permissions.
+func MigrateForPermissions(db *gorm.DB) error {
```
viktorstrate, line 11: We should test this migration thoroughly before releasing, since it is a little more involved.
Just to make sure a lot of users databases doesn't get corrupted.

api/graphql/schema.graphql
```diff
@@ -134,37 +158,41 @@ type Mutation {
     id: ID!
     username: String
     password: String
-    admin: Boolean
-  ): User! @isAdmin
+    roleId: ID
+    admin: Boolean @deprecated(reason: "The roles system replaces the old admin boolean.")
```
viktorstrate, line 162: I really like that you deprecated the old method. Then when we want to make a 3.0 release at some point, we can refactor out all the deprecations.

api/graphql/schema.graphql
```diff
@@ -1,9 +1,16 @@
 directive @isAuthorized on FIELD_DEFINITION
-directive @isAdmin on FIELD_DEFINITION
+directive @hasPermission(permission: Permission!) on FIELD_DEFINITION
```
viktorstrate, line 2: Having a `@hasPermission` directive is awesome :)

viktorstrate (changes requested):
I think there's a problem with the `database_models` in database.go, see comment.

api/database/database.go
jordy2254: @viktorstrate the reason for this was due to some query size constraints atleast within mysql. I was going to investigate further to find a better solution. I will take a further look st this and hope its the cause I'd the failures

api/database/migrations/permissions.go
jordy2254: Fully agree on this one, I've tested the scenarios I can think of albeit on a small dataset and only a few user's. These are as below.
- Migrating from an old version of photoview with various admin users, normal user & demo user
- Addition of a new permission
- Addition of an existing permission to a system user group doesn't add to the user group
- Addition of a new permission to a system user group automatically adds to the user group

jordy2254:
Just an update for anyone interested on this. First of all I'll need some user's to test migrations if anyone see's this and is interested ping me a message on discord.

I've not had much time to look at this as I had hoped but finally managed to get a look this afternoon at some of the test faliures and it's not mostly working. below is what's left/what I've changed since last pushing

- The test suite is now nearly fixed, I've implemented database transactions for all tests rather than the delete everything system for the end of a test that was implemented before.
- I have one more test set to fix which is the scanner tests, this is because of the scanner creating a transaction per thread which locks the outer transaction causing errors, I'm yet to investigate this one but have some ideas.
- UI Changes still need doing the initial pass of these will be the ability to select a user role for a user only and not create and amend the roles themselves.

I'm hoping I'll be able to finish this off during the week depending on how much dev time I get but I won't promise that just yet.

api/api
jordy2254, line 1: This needs to be removed... whoops

ui/src/Pages/SettingsPage/Users/RoleSelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const items: DropdownItem[] = []
+
+  if (error) {
+    return <div> Error</div>
+  }
+  if (loading) {
+    items.push({ label: 'Loading...', value: '' })
+  } else {
+    if (!data?.roles.find(role => role.id === props.selected)) {
+      items.push({ label: 'Please Select', value: '' })
```
jordy2254, line 32: Missing translation

PR #940, opened 2024-04-25
fix: remove logging of database url in place for just dialect

### LINKED ISSUE #937: [Security vulnerability]: Photoview outputs the DB connection string as a plain text to the log
**Describe the bug**
Photoview service outputs to the log the complete DB connection string, including all secrets, even password

**To Reproduce**
1. Start the Photoview
2. Check the log (in the case of Docker, it is the log of the photoview conatiner)
3. See error: Connecting to MYSQL database: <the plain text connection string>

**Expected behavior**
It should never be sent to the log.
In the case of debugging, it can be sent in TRACE log level, but never on INFO

**Your environment:**

 - Are you using Docker? [yes]
 - Database [mariadb LTS]
 - Version [master]

**Additional context**
It is important, as logs are the typical target for malicious actors and never should contain secrets

jordy2254:
Fixes #937

### DIFF
--- a/api/database/database.go
+++ b/api/database/database.go
@@ -72,21 +72,21 @@ func GetSqliteAddress(path string) (*url.URL, error) {
 
 func ConfigureDatabase(config *gorm.Config) (*gorm.DB, error) {
 	var databaseDialect gorm.Dialector
-	switch drivers.DatabaseDriverFromEnv() {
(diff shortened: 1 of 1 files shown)


kkovaletp (approved):

jordy2254:
Merging even though coverage report error as change is just changing logging.

Issue #850, opened 2023-05-30
Photoview page "crashes" when trying to display sidebar for media with invalid gps data


lesensei:
**Describe the bug**
One of our family camera seems to use a default "max" value for latitude and longitude data when the GPS can't fix a position. This translates to the following:
```
# exiftool -location:all photo.JPG
Location                        :
Country                         :
State                           :
City                            :
Landmark                        :
City 2                          :
GPS Version ID                  : 2.3.0.0
GPS Latitude Ref                : North
GPS Longitude Ref               : East
GPS Altitude Ref                : Above Sea Level
GPS Status                      : Measurement Void
GPS Measure Mode                : 2-Dimensional Measurement
GPS Dilution Of Precision       : 0
GPS Img Direction Ref           : Magnetic North
GPS Img Direction               : 202.5
GPS Map Datum                   : WGS-84
GPS Processing Method           : GPS
GPS Area Information            :
GPS Altitude                    : 0 m Above Sea Level
GPS Latitude                    : 17056881 deg 51' 12.15" N
GPS Longitude                   : 17056881 deg 51' 12.15" E
GPS Position                    : 17056881 deg 51' 12.15" N, 17056881 deg 51' 12.15" E
```
The scanner doesn't choke on this and integrates the values in the database (with maker/model for ref):
```
MariaDB [photoview]> select camera,maker,gps_latitude,gps_longitude from media_exif where id = 1;
+----------+-----------+-----------------+-----------------+
| camera   | maker     | gps_latitude    | gps_longitude   |
+----------+-----------+-----------------+-----------------+
| DMC-TZ40 | Panasonic | 17056881.853375 | 17056881.853375 |
+----------+-----------+-----------------+-----------------+
```
However, the problem arises when trying to display the info sidebar for one of the affected pictures (from the dev console):
```
Error: Invalid LngLat latitude value: must be between -90 and 90
    bt https://[fqdn]/assets/mapbox-gl.9b1e4062.js:8
    convert https://[fqdn]/assets/mapbox-gl.9b1e4062.js:8
    jumpTo https://[fqdn]/assets/mapbox-gl.9b1e4062.js:1081
    Map https://[fqdn]/assets/mapbox-gl.9b1e4062.js:1085
    D https://[fqdn]/assets/MapboxMap.284406f0.js:86
    Zs https://[fqdn]/assets/index.5073f743.js:32
    [...]
```

[...]

PR #951, opened 2024-05-21
Validate incoming GPS data

kkovaletp:
Validate incoming GPS data and throw an error if it is incorrect, storing Null values.
This is the fix for the #850 

I've created this PR with Bing Copilot's help and didn't find the correct way to add the migration, which should find all incorrect GPS data in the DB and replace it with Null values.
I'd really appreciate it if someone could explain how to add a new migration correctly. I'll then write it and add it to the PR. At least I hope that it won't be much more difficult than writing the fix)


api/scanner/exif/exif_parser_external.go
```diff
@@ -196,6 +198,14 @@ func (p *externalExifParser) ParseExif(media_path string) (returnExif *models.Me
 		newExif.GPSLatitude = &latitudeRaw
 	}
 
+	// **Added GPS data validation**
+	// https://github.com/photoview/photoview/issues/850
+	if found_exif && (latitudeRaw < -90 || latitudeRaw > 90 || longitudeRaw < -90 || longitudeRaw > 90) {
```
jordy2254, line 203: The below reads nicer and smoother the comments above this are not required. However, I personally I would at this stage extract the retrieval and validation of latitude and longitude into a function which returns the latitude and longtude to set on the data.

```go
func extractValidGpsData(fileInfo) (float, float) {
 // extract latitude and longitude, validate and return values.
}
```

```suggestion
	if found_exif && ( math.Abs(latitudeRaw) > 90 || math.Abs(longitudeRaw) > 90) {
```

As you've said before you're not a dev if you need more guidance feel free to reach out :)

api/scanner/exif/exif_parser_internal.go
```diff
@@ -142,8 +142,16 @@ func (p internalExifParser) ParseExif(media_path string) (returnExif *models.Med
 
 	lat, long, err := exifTags.LatLong()
 	if err == nil {
-		newExif.GPSLatitude = &lat
-		newExif.GPSLongitude = &long
+		// **Added GPS data validation**
+		// https://github.com/photoview/photoview/issues/850
+		if lat < -90 || lat > 90 || long < -90 || long > 90 {
+			returnErr = errors.New(fmt.Sprintf("Incorrect GPS data in the %s Exif data: %f, %f, while expected values between '-90' and '90'", media_path, long, lat))
```
jordy2254, line 148: - No comment
- I don't believe the nil setting is required on this based on the context I can see, we should be able to invert the condition and if it's within the correct range apply it.

My same comment re: `math.Abs()` applies here too.

api/scanner/exif/exif_test.go
```diff
@@ -75,6 +75,22 @@ func TestExifParsers(t *testing.T) {
...
+				assert.Nil(t, exif.GPSLongitude)
+			},
+		},
+		{
+			path: "./test_data/CorrectGPS.jpg",
+			assert: func(t *testing.T, exif *models.MediaEXIF) {
+				assert.NotNil(t, exif.GPSLatitude)
+				assert.NotNil(t, exif.GPSLongitude)
+				assert.True(t, *exif.GPSLatitude >= -90 && *exif.GPSLatitude <= 90)
+				assert.True(t, *exif.GPSLongitude >= -90 && *exif.GPSLongitude <= 90)
```
jordy2254, line 91: I think unless we're going to add more test conditions for edge cases etc. the NotNil assertions are fine here.

jordy2254 (changes requested):

kkovaletp:
@jordy2254, I managed to fix syntax errors, however, now getting a failed unit test from an unexpected area... Please take a look at your free time, but I think that this code refactoring with GPS data processing extraction into a separate function needs to be reverted and done in a separate PR, as it is not directly related to the purpose of this PR, while I don't have GO programming skills to understand all hidden consequences of my changes(
Please let me know what you think)

jordy2254:
@kkovaletp  With regards to the migration I still want to introduce a better more standard migration tool if in the meantime you can follow the same format I did in this pr that would be great (chuck the migration in a go file within a migrations file, and export the function) https://github.com/photoview/photoview/pull/927/files#diff-78952107b5c08f78733a3be1140445096581151507716a74eed4d54b70d66f4c

Below is the sql (for reference)
```sql 
update media_exif
set gps_latitude=0 and gps_longitude=0
where abs(gps_longitude) > 90 OR abs(gps_latitude) > 90
```

Below should be the sql written in gorm;
db.Model(&MediaExif{}).
		Where("ABS(gps_longitude) > ?", 90).
		Or("ABS(gps_latitude) > ?", 90).
		Updates(MediaExif{GPSLatitude: 0, GPSLongitude: 0})

> **NOTE** I havn't checked all dialects for this but sqllite supports it so hopefully they all will

jordy2254:
@kkovaletp 
Your failing test appears to be an issue with the assertions on your gps data. Floating point number's are not precise and loose precision as such I expects it's due to this, the assertions shouldn't be using `assert.True` and instead should be using `assert.EqualValues(t, *exif.GPSLongitude, 1.12345)`

image

kkovaletp:
Hi @jordy2254, thanks for your help!
Finally, I completed the development of this PR with the code working and tests passed. Please review it.
However, I have a question about the coverage:
- According to the Codecov report, my test for the parser runs only for the internal parser code. Why the external parser is not covered by all the same tests, while it is mentioned in the testing configuration (the code of that test)? How to make it run for the external parser code as well?
- I've added the test for my new migration and made it work. I see that it is passed in the tests report (on the GitHub Actions page), but Codecov doesn't show this migration file in the list of changed files at all. How can it be included in the report? BTW, the coverage in the testing report for the migration is shown at 4.5% - it is not possible, as my test covers all possible cases. I wanted to see the detailed report in Codecov, but the migration is missing there. Do you have some ideas as to why the coverage is so low?

jordy2254:
- This one per the image below is because if the exiftool isn't found it just doesn't add it to the parsers and thus only provides a small log message. 
image

- To answer why the coverage is low I think unit coverage of migration steps are a bit pointless, they should be immutable and never change once the PR is merged, changes to migration steps should be done as a new section of code.

Code wise I'm happy with the change now, I'm not sure if you want to do more investigation into that test suite or not I'll leave that to you.

jordy2254 (approved):

kkovaletp:
@jordy2254, thanks for your review and feedback)
- I'll try to check whether I'll be able to understand why exiftool is not found and what to do to make it be detected and used by the test
- As I'm not a developer and don't feel enough confidence about my changes, I'd prefer to cover them by unit tests as much as I can. I agree with you that the unit test for migration is a one-time benefit, but it was an interesting experience for me to implement the test, as well as in this way I can be quite more sure that the changes in the migration work as expected and don't break anything else (there are still some chances, but significantly lower with unit test being running)

kkovaletp:
> * I'll try to check whether I'll be able to understand why exiftool is not found and what to do to make it be detected and used by the test

I found the cause: both ffmpeg and exiftool are not installed on the system, where API unit tests are running.

The strange thing is once I've added them to the installation step, all tests for the external runner failed. At least 2 of failed tests were failed because of the different precision of internal and external tools. It will be easy to fix by rounding the value to the internal tool's level. But I'll need more time to understand why other tests failed. So far I can assume that the execution sequence is incorrect and that somehow internal tool's tests impact on the same tests for the external tool (but this is just 1st assumption and I need to investigate it deeper). @jordy2254, what do you think?

kkovaletp:
@jordy2254, I finally fixed all the tests, but I also had to modify a bit the source code, so please review this PR again to make sure that I didn't make some unwanted or stupid changes?)

BTW, even after all tests run now for both `internal` and `external` parsers (at least they were failing before I fixed them, so I hope that they run now), I still see almost no coverage for the `external` parser on the https://app.codecov.io/gh/photoview/photoview/blob/GPS-validation/api%2Fscanner%2Fexif%2Fexif_parser_external.go
Can you please check that the tests are running for both parsers and that coverage is updated on the CodeCov side? If both are true, then I don't understand why the coverage is so low (especially for the external parser)...

kkovaletp:
BTW (another one), when I did unit test failures investigation and fixing, I noticed that the same function `ParseExif(media_path string)`, executed from internal or external parser's code, will return different results (at least for the case of the `stripped.jpg` test image): it could return a primitive `nil` or an object with all properties set to `nil`, except the `ShotDate`, which is set to the date and time of the scan.

To me, it looks like a 'code smell', as if both internal and external parsers are implementations of the same 'abstract parser' object, they should follow the same contract and return values of the same structure/type for the same incoming data.

@jordy2254, what do you think? Do you see other similar 'code smells' in this module? Which of them does it make sense to refactor in the scope of this PR and which - in some future dedicated PR?

kkovaletp:
@googollee , please review this PR and try to answer my last 2 comments, originally written to @jordy2254 (as he is quite busy now and might look on this after some time)

api/database/database.go
```diff
@@ -194,6 +195,11 @@ func MigrateDatabase(db *gorm.DB) error {
 		log.Printf("Failed to run exif fields migration: %v\n", err)
 	}
 
+  // Remove invalid GPS data from DB
```
googollee, line 198: The indentation seems wrong. It should be tabs instead of spaces. Could you run `go fmt` here?

api/database/migrations/exif_invalid_gps_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/photoview/photoview/api/database/migrations"
+	"github.com/photoview/photoview/api/graphql/models"
+	"github.com/photoview/photoview/api/test_utils"
+)
+
+func TestExifMigration(t *testing.T) {
+  envFile, err := os.Open("/home/runner/work/photoview/photoview/api/testing.env")
```
googollee, line 18: `go fmt` here.

api/database/migrations/exif_invalid_gps_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	scanner := bufio.NewScanner(envFile)
+	for scanner.Scan() {
+		line := scanner.Text()
+		parts := strings.SplitN(line, "=", 2)
+		if len(parts) != 2 {
+		  t.Fatalf("Invalid line in environment file: %s", line)
+		}
+		key, value := parts[0], strings.Trim(parts[1], "'")
+		os.Setenv(key, value)
```
googollee, line 32: Do we not have a function to set env from a file?

api/database/migrations/exif_invalid_gps_test.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	db := test_utils.DatabaseTest(t)
+	//defer db.Exec("DELETE FROM media_exif") // Clean up after test
```
googollee, line 36: Should we uncomment this?

api/scanner/exif/exif_parser_external.go
```diff
@@ -63,6 +63,31 @@ func sanitizeEXIF(exif *models.MediaEXIF) {
 	}
 }
 
+func extractValidGpsData(fileInfo *exiftool.FileMetadata, media_path string) (*float64, *float64) {
+	var GPSLat, GPSLong *float64
+
+  // GPS coordinates - longitude
```
googollee, line 69: `go fmt` here.

api/scanner/exif/exif_test.go
```diff
@@ -65,16 +66,61 @@ func TestExifParsers(t *testing.T) {
 		},
 		{
 			path: "./test_data/stripped.jpg",
-			assert: func(t *testing.T, exif *models.MediaEXIF) {
-				assert.Nil(t, exif)
+			assert: func(t *testing.T, exif *models.MediaEXIF, err error) {
+				assert.NoError(t, err)
+				if exif == nil {
```
googollee, line 71: Is there any reason to change this test case? If the file doesn't have exif, this should be `nil`, but not an instance with zero value.

kkovaletp: the reason is that the internal and external parsers return different results, as I mentioned in my previous comment: in 1 case it is a simple `nil`, while in another - an object with empty values

api/database/migrations/exif_invalid_gps_test.go
kkovaletp: I'm not sure about the cleanup here, as having media with GPS data in the testing DB doesn't break other tests.
In general, I know that it is better to generate and clean data for each test, I wasn't sure about the current tests suite execution, so I didn't enable the cleanup.
If you confirm that it is better to do this, I'll uncomment the line.

api/database/migrations/exif_invalid_gps_test.go
kkovaletp: I don't know) if there is something better, please suggest what would be the replacement)
I'm not a developer and write code based on my general understanding of the functionality and with help from AI)

googollee: OK. I'll do this later.

api/database/migrations/exif_invalid_gps_test.go
googollee: I hope to clean the test data after the test. It's friendly to run tests locally.

api/scanner/exif/exif_test.go
googollee: SG.

api/database/migrations/exif_invalid_gps_test.go
kkovaletp: ok, I'll uncomment the cleanup after the workflows are enabled in the repo, as now they got disabled for some reason(

googollee (approved):

Issue #632, opened 2022-01-21
Timeline show one date from all years


savegame:
**Describe the bug**
First i add 2017 year to indexing, than 2018 ... Than when i try view in time line dates like "from 2018 and earler", its show 2017 dates first

**To Reproduce**
1. First add first album from one month or year ( as in my case, first i add 2017 year for indexing )
2. Then add next album ( in my case its 2018 year ) 
3. Open timeline, and set up "from 2018 and earler"
4. It still show first 2017 year, and if it has same dates as in 2018, it show both pictures in one line from both years

**Expected behavior**
Should show 2018 first, then 2017

**Screenshots**

![изображение](https://user-images.githubusercontent.com/16311332/150502795-f421184a-903c-4ce8-a8e6-6ed3d7846701.png)  
![изображение](https://user-images.githubusercontent.com/16311332/150502356-458691b2-a09a-4892-bf8b-bb42bf10a390.png)  

**Your environment:**
Feel free to remove and add what you think might be useful.

 - Build it in jail in FreeBSD 12.2
 - Database: sqlite ( just for testing this software, later i'll try with postgresql ) 
 - Browser: Firefox 96.0 ( Ubuntu 21.10 )
 - Version: 2.3.9

**Additional context**
Host system is TrueNAS, in jail ( FreeBSD 12.3 )

savegame:
P.S. in my previous case, i use photoview in this algorithm (with SQLite): 
1. add folder for scanning /media/2017/01  ( for testing ) 
2. delete this album in settings
3. add this album again in settings and press scan
4. delete it again in settings
5. add /media/2017 as album and press scan
6. add /media/2018 as album and press scan 
- after than i was missing 01 folder from 2017 year ( it shown as empty ) , and bug, described upper

savegame:
@viktorstrate looks like with postgres bug is not reproduced ( later i'll try reproduce this bug on separate jail, just for test purposes (with SQLite again))
and i have another bug =)   
looks like in inforation panel, path to pictures album has wrong format 
![изображение](https://user-images.githubusercontent.com/16311332/152947543-96a5e9c5-89ef-478c-8926-efb047785bbb.png)   
real album path is   
![изображение](https://user-images.githubusercontent.com/16311332/152947795-0cca672a-3c95-491b-bed5-f4e07a54048f.png)

viktorstrate:
Right, thank you for looking into this.
This seems to be related to sqlite then. Sqlite is very unstable see #618 for example.
[...]

PR #954, opened 2024-05-24
Fix timeline order with sqlite3. It should be ordered as MySQL and PGSQL like `(year, month, day)`.

googollee:
When using `sqlite` as database, `Timeline` shows media ordering by the album's day of the year. If I have albums like:

- Album1 on 2020-12-31
- Album2 on 2021-01-01

My timeline shows media in `Album1` first, then media in `Album2`, even `Album2` is more recent.

I see `PostgreSQL` and `MySQL` are ordering with `(year, month, day)`, which is reasonable. Only `sqlite` is ordering with `(the day of the year)`. I don't know why `sqlite` has a special behavior here.

This patch matches the sqlite behavior with other databases.

### DIFF
--- a/api/graphql/models/actions/timeline_actions.go
+++ b/api/graphql/models/actions/timeline_actions.go
@@ -24,7 +24,7 @@ func MyTimeline(db *gorm.DB, user *models.User, paginate *models.Pagination, onl
 			Order("media.date_shot DESC")
 	case drivers.SQLITE:
 		query = query.
-			Order("strftime('%j', media.date_shot) DESC"). // convert to day of year 001-366
+			Order("strftime('%Y-%m-%d', media.date_shot) DESC"). // convert to YYYY-MM-DD
 			Order("albums.title ASC").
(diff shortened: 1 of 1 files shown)


kkovaletp:
@googollee, thanks for the PR! Nice catch!
However, looking into the way, the original code is written, it looks like that ordering was implemented intentionally, and I have no idea why it was needed. It would be great to find out why it was implemented in that way before merging this PR to not break something else.

@jordy2254, @viktorstrate, Do you know, remember, or at least have some assumption on why the original ordering was implemented for SQLite DB case differently than for other DBs?

viktorstrate:
Probably just a mistake :)

kkovaletp:
Let's wait for @jordy2254 opinion. Maybe he has some ideas, anyway he is going to do the code review before merging the PR.

mih:
I applied this patch and it achieves the desired behavior change for me. Thanks a lot!

It also addresses #632 for me.

jordy2254 (approved):

PR #960, opened 2024-06-02
Optimize user's album filter.

### LINKED ISSUE #705: [Refactoring] Very slow album view for huge photo collection in deep directory trees.
I'm currently trying out this application with a photo/media collection of around 200+ GB. The “special” thing about it is quite organized in very deep directory trees. I don't know if that is the reason, but it's something I though could be important.

The main problem is that loading this entire library into photoview and then wanting to access the album page, it takes around 5-6 seconds or even longer before it displays anything. I thought there might be some error, so I had a look at the logs and found that the SQL Statements were taking so long. Especially this one: `SELECT * FROM `albums` WHERE id IN...`. After the IN keyword there are thousands of IDs listed that as far as I can guess the IDs of the different subdirectories are.
I don't know how one could improve this but I thought it was something that should be known.

Just to show how large this can become:
image

googollee:
To resolve #705.

- Remove filling albums to the user.
  - It's pretty slow if a user has a lot of albums.
- Remove checking if the user owns the album by `album_id IN ()`.
  - It's pretty slow as the `user_albums` table doesn't have a proper `user_id` index.
- Check if (user_id, album_id) exists in `user_albums` relationship table.
  - The check uses the index (user_id, album_id) as the primary key.
- Updated `album_actions_test.TestAlbumPath()` as a directory and all parent directories should be associated with the user.
  - From the root path, all sub-directories are pushed into the queue and should be associated with the user according to [`scanner_user.FindAlbumsForUser()`](https://github.com/photoview/photoview/blob/master/api/scanner/scanner_user.go#L210).
  - From my real database, a directory and all its parent directories are associated with the user.

```
sqlite> WITH recursive path_albums AS (SELECT * FROM albums anchor WHERE anchor.id = 176 UNION SELECT parent.* FROM path_albums child JOIN albums parent ON parent.id = child.parent_album_id) SELECT * FROM path_albums WHERE EXISTS (SELECT 1 FROM user_albums WHERE user_id=1 AND album_id = 176 LIMIT 1);
id|created_at|updated_at|title|parent_album_id|path|path_hash|cover_id
176|2024-05-20 15:36:35.729790324+00:00|2024-05-20 15:36:35.730176668+00:00|.......|24|/media/photos/......|e60b0bab33bbc14032a04b25e59009b8|
24|2024-05-20 15:36:24.645216255+00:00|2024-05-20 15:36:24.645989796+00:00|.......|2|/media/photos/........|459b9499d75792864768bbe494f79390|
2|2024-05-20 15:36:22.798409883+00:00|2024-05-20 15:36:22.799627995+00:00|......|1|/media/photos/......|1e3b17ed968875c070088df3f3dd81f7|
1|2024-05-20 15:36:13.829977475+00:00|2024-05-20 15:36:13.829977475+00:00|.......||/media/photos|26807d2efa623879f289dbb13cfaf931|
sqlite> SELECT * FROM user_albums WHERE user_id = 1 AND album_id IN (176, 24, 2, 1);
album_id|user_id
1|1
2|1
24|1
176|1
```

### DIFF
--- a/api/graphql/models/actions/album_actions_test.go
+++ b/api/graphql/models/actions/album_actions_test.go
@@ -12,7 +12,7 @@ import (
 func TestAlbumPath(t *testing.T) {
 	db := test_utils.DatabaseTest(t)
 
-	album := models.Album{
+	album := &models.Album{
 		Title: "Three",
 		Path:  "/one/two/three",
 		ParentAlbum: &models.Album{
@@ -30,9 +30,9 @@ func TestAlbumPath(t *testing.T) {
 	user, err := models.RegisterUser(db, "user", nil, false)
 	assert.NoError(t, err)
 
-	db.Model(&user).Association("Albums").Append(album.ParentAlbum.ParentAlbum)
+	db.Model(&user).Association("Albums").Append(album.ParentAlbum.ParentAlbum, album.ParentAlbum, album)
 
-	albumPath, err := actions.AlbumPath(db, user, &album)
+	albumPath, err := actions.AlbumPath(db, user, album)
 	assert.NoError(t, err)
 	assert.Len(t, albumPath, 2)
 	assert.Equal(t, "Two", albumPath[0].Title)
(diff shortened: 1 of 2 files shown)


googollee:
I think the behavior is not the same as the previous query. I'll improve it.

PR #962, opened 2024-06-02
Optimize user's album filter. #705

### LINKED ISSUE #705: [Refactoring] Very slow album view for huge photo collection in deep directory trees.

googollee:
To resolve #705.

- Remove filling albums to the user.
  - It's pretty slow if a user has a lot of albums.
- Remove checking if the user owns the album by `album_id IN ()`.
  - Scanning the whole table is pretty slow because the `user_albums` table doesn't have a proper `user_id` index.
- Check if (user_id, album_id) exists in `user_albums` relationship table.
  - The check uses the index (user_id, album_id) as the primary key.

```
sqlite> explain query plan WITH recursive path_albums AS (SELECT * FROM albums anchor WHERE anchor.id = 176 UNION SELECT parent.* FROM path_albums child JOIN albums parent ON parent.id = child.parent_album_id) SELECT * FROM path_albums WHERE EXISTS (SELECT 1 FROM user_albums WHERE user_albums.user_id=1 AND user_albums.album_id = id LIMIT 1);
QUERY PLAN
|--CO-ROUTINE path_albums
|  |--SETUP
|  |  `--SEARCH anchor USING INTEGER PRIMARY KEY (rowid=?)
|  `--RECURSIVE STEP
|     |--SCAN child
|     `--SEARCH parent USING INTEGER PRIMARY KEY (rowid=?)
|--SCAN path_albums
`--CORRELATED SCALAR SUBQUERY 3
   `--SEARCH user_albums USING COVERING INDEX sqlite_autoindex_user_albums_1 (album_id=? AND user_id=?)

sqlite> explain query plan WITH recursive path_albums AS (SELECT * FROM albums anchor WHERE anchor.id = 176 UNION SELECT parent.* FROM path_albums child JOIN albums parent ON parent.id = child.parent_album_id) SELECT * FROM path_albums WHERE id in (1,2,4,5,3,6,7);
QUERY PLAN
|--CO-ROUTINE path_albums
|  |--SETUP
|  |  `--SEARCH anchor USING INTEGER PRIMARY KEY (rowid=?)
|  `--RECURSIVE STEP
|     |--SCAN child
|     `--SEARCH parent USING INTEGER PRIMARY KEY (rowid=?)
`--SCAN path_albums
```

### DIFF
--- a/api/graphql/models/user.go
+++ b/api/graphql/models/user.go
@@ -165,18 +165,8 @@ func (user *User) FillAlbums(db *gorm.DB) error {
 }
 
 func (user *User) OwnsAlbum(db *gorm.DB, album *Album) (bool, error) {
-
-	if err := user.FillAlbums(db); err != nil {
-		return false, err
-	}
-
-	albumIDs := make([]int, 0)
-	for _, a := range user.Albums {
-		albumIDs = append(albumIDs, a.ID)
-	}
-
 	filter := func(query *gorm.DB) *gorm.DB {
-		return query.Where("id IN (?)", albumIDs)
+		return query.Where("EXISTS (SELECT 1 FROM user_albums WHERE user_albums.user_id = ? AND user_albums.album_id = id LIMIT 1)", user.ID)
 	}
 
 	ownedParents, err := album.GetParents(db, filter)


kkovaletp:
@googollee, thanks for your PR. Addressing performance issues is an important improvement for the product!

As I'm not a developer, it is difficult to understand from the code how exactly the logic has been changed, so I rely more on your PR's description with a lot of checks being removed. And it looks strange to me, as they were added to the code on purpose (I'm not 100% sure which, but still).

@jordy2254, we need your review and feedback here as a backend developer.

But until then, let me ask you some questions:
- does the updated query do 100% the same job and return the same results as the original one (but faster and in an optimal way)? If yes, my other questions don't make sense, but if not really, then:
- Instead of removing some of the checks and filters, because they are slow, wouldn't it be better to optimize the DB schema by optimizing links between tables, adding needed indexes, pagination to those tables, and moving some of the logic to DB triggers and procedures? Asking that I understand that we support 3 different DBMSs and at least SQLite doesn't support everything mentioned, so we might need to do some research and implement the best combination of those optimizations, which would work for all 3 DBMSs.

googollee:
hi @kkovaletp,

Thanks for the quick response. I'll explain what it was and what I changed.

- Before:
  - Load all albums belonging to a user: `user.FillAlbums(db)` and `albumIDs`.
  - Search for an album and all parent albums: `album.GetParents()`
    - Make sure that all these albums fall in the group of albums belonging to the user: `query.Where("id IN (?)", albumIDs)`
- After:
  - Search for an album and all parent albums: `album.GetParents()`
    - Make sure all these albums belong to the user: `query.Where("EXISTS (SELECT 1 FROM user_albums WHERE user_albums.user_id = ? AND user_albums.album_id = id LIMIT 1)", user.ID)`

`user_albums` records the relationship between users and albums. `SELECT 1 FROM user_albums WHERE user_albums.user_id = ? AND user_albums.album_id = id LIMIT 1` returns an record if the given user has the album with `id`, and returns no records if the given user doesn't have one. `SELECT 1` avoids loading fields from the table because the query doesn't use any fields; `LIMIT 1` avoids the query checking more than one record. `EXISTS ()` checks if there are any records, and if there are, it returns `true`, otherwise returns `false`.

It should not remove any checks or filters.

kkovaletp:
@googollee, thanks for your explanation.
So, you've optimized the query to work faster while doing all the same checks and filtering. Nice job!

Do you see the need to add more indexes to the DB to speed up the query execution even more? Does it make sense to include them in this PR or it is better to make a new PR with all missing indexes across all tables?

googollee:
> Do you see the need to add more indexes to the DB to speed up the query execution even more? Does it make sense to include them in this PR or it is better to make a new PR with all missing indexes across all tables?

I find many SQL querying the `user_albums` table, and adding an index may speed up these SQL. But it's not easy to add such indexes with GORM framework. So I'm not sure. One possible way is adding indexes in [`database.MigrateDatabase()`](https://github.com/photoview/photoview/blob/df1610b13cbf3625955b01f430422b3b805c895e/api/database/database.go#L176), but it's far away from the definition of [`user_albums` table](https://github.com/photoview/photoview/blob/df1610b13cbf3625955b01f430422b3b805c895e/api/graphql/models/album.go#L17), which I don't like. As I can optimize the query without adding indexes, I'd like to keep it simple and put the decision of indexes off until we need them.

kkovaletp:
> I find many SQL querying the `user_albums` table, and adding an index may speed up these SQL. But it's not easy to add such indexes with GORM framework. So I'm not sure. One possible way is adding indexes in [`database.MigrateDatabase()`](https://github.com/photoview/photoview/blob/df1610b13cbf3625955b01f430422b3b805c895e/api/database/database.go#L176), but it's far away from the definition of [`user_albums` table](https://github.com/photoview/photoview/blob/df1610b13cbf3625955b01f430422b3b805c895e/api/graphql/models/album.go#L17), which I don't like. As I can optimize the query without adding indexes, I'd like to keep it simple and put the decision of indexes off until we need them.

I see. Thanks for the explanation. Let's see what would be @jordy2254 point of view...
Potentially, adding indexes would speed up the DB operations even after your PR gets merged, so this is something useful in any case. Probably it does make sense to do it for all affected tables in a separate PR to be more focused on the purpose of each PR and provide more consistent set of indexes there.

googollee:
@Tuxie I agree that SQL optimization should optimize for `LIMIT 1`, especially as `(user_id, album_id)` is a combined primary key and should be unique. I prefer explicitly than implicitly and keep `LIMIT 1` here.

jordy2254 (approved):

PR #972, opened 2024-06-22
Urgent small fixes in the Dockerfile and Compose configs

kkovaletp:
Freeze UID and GID for the `photoview` user inside the image; added alternative permission configs description to the Readme and Makefile (as a comment); switched the default tag in the compose from `2` to `master`, as this is the most common user misunderstanding according to the issues and Discord questions.

The tag change in the compose files is needed, as the `2` image is too old, and contains a lot of fixed bugs, which users keep reporting. It is not obvious for them to switch the tag. Also, I don't believe the release is happening soon. Anyway, we could change the tag back right before the release.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,8 @@ ARG TARGETPLATFORM
 # See for details: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 # Create a user to run Photoview server
-RUN useradd -r -U -m photoview \
+RUN groupadd -g 999 photoview \
+  && useradd -r -u 999 -g photoview -m photoview \
   # Required dependencies
   && apt-get update \
   && apt-get install -y curl gnupg gpg libdlib19.1 ffmpeg exiftool libheif1 sqlite3 \
--- a/README.md
+++ b/README.md
@@ -105,6 +105,11 @@ All the photo galleries can do a lot of what I need, but no single one can do it
 
    If command(s) return `Permission denied` error, run them under the user, owning corresponding files and folders. Alternatively, run them adding `sudo ` before the command: this will switch the execution context to `root` user and ask for the root password. You have to have permission to run `sudo` in the system.
 
+   If you don't want to give required permissions to `others` group for your files, alternatively, you can:
+
+   - create a group on your host with GID=999 and make all the files and folders inside volumes of the `photoview` service being owned by this group; then set the appropriate permissions to the `group` section.
+   - create on your host a group with GID=999 and a user in this group with UID=999; then change the ownership of all the files and folders inside volumes of the `photoview` service to this user; then set the appropriate permissions to the `user` section.
+
    If you configured other mounts with media files from other locations on the host (like HOST_PHOTOVIEW_MEDIA_FAMILY or anything else), you need to run the same commands, as in the `Makefile` `readable` target, for each media root folder on your host manually: copy each command to your shell and replace the variable with the absolute path to an additional media root folder without the trailing `/`. Run both commands for each additional root folder.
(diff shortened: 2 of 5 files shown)


viktorstrate:
Regarding changing the tag to master. Why don't we make more releases, what's holding us back right now?

Dockerfile
```diff
@@ -62,7 +62,8 @@ ARG TARGETPLATFORM
 # See for details: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 # Create a user to run Photoview server
-RUN useradd -r -U -m photoview \
+RUN groupadd -g 999 photoview \
+  && useradd -r -u 999 -g photoview -m photoview \
```
viktorstrate, line 66: What implications does this have on existing setups. Could this break existing setups such that Photoview no longer has access to the files?

kkovaletp:
@viktorstrate,

> Regarding changing the tag to master. Why don't we make more releases, what's holding us back right now?

Right now we have at least #797 waiting for the finalization and merge. As far as I understand from @jordy2254 comments, a small code change is required, but the original contributor is not responsive, so I asked @jordy2254 to step in and make the needed change, following by merge of the PR. Waiting for this to happen.

While this is the only PR left not merged from the original release plan, we have now 1 more PR #965, fixing an annoying UI issue. The fix is implemented and tested manually, but I cannot finish unit tests development for the module, as they keep failing no matter what I try from AI suggestions. I need some help from a developer to debug and fix those unit tests before the PR can be reviewed and merged. I asked @jordy2254 for help and waiting for his free time to deep dive into the frontend and find out the way to fix them...

Basically, the main issue with the release is that our project needs more experienced developers with the maintainer role, so that we don't have a bottleneck with PR review and consultancy. Saying this, I completely understand that we all here work for fun in our free time and don't get any salary, so no one is obligated to spend a minimal amount of hours per week on the project. But still: the issue is not solved. Some time ago I posted the next invite on my [Linkedin page](https://www.linkedin.com/feed/update/urn:li:activity:7204541422807560192/), but didn't get any responses and don't see volunteers in the project. You might want to repost it for your network or post something similar to your friends) I don't know how else I can help to solve this problem...

Dockerfile
kkovaletp: Right now the image is created with the user and group, having UID=999 and GID=999, but as the command gets the 1st free ID, there is no guarantee that in the future after base image update these IDs won't change. That is why even while users see the IDs inside the container, they can't rely on them when managing FS permissions on the host. 

With this change, the IDs would be frozen/hardcoded, but still the same as currently used. This means, that those, who already configured their permissions on the host with the current IDs, will not see any difference, while all others will be able to do the same and rely on the IDs. If in the future at least 1 of the IDs is locked and not available at the build time, the image build will fail and we'll get the explicit error about this issue instead of automatically using some other ID and breaking users' setup.

viktorstrate:
But is the issue blocking the release, why are we waiting for it to be merged?

kkovaletp:
@viktorstrate, 

#797 is an improvement. Not a bug fix. So, it is not blocking the release.
#965 is a bug fix for an annoying user-visible issue, but it is not a blocker and the product can be used "as is".

I'll understand if you'd like to push the release without these 2 PRs being included. However, if we can make them ready and merge before the release, I think that the final release build would be much better.

Please let me know your decision, so that I can update this PR with the correct tag so that it is included in the release anyway.

viktorstrate:
I think it's better to have smaller and more frequent releases, rather than recommending users to use the master tag ☺️

kkovaletp:
@viktorstrate, ok. I've reverted back the tag-related changes in the compose files.
Please approve (I've enabled auto-merge for this PR, so once all jobs are passed and there is approval, it will be merged automatically).

Regarding the release, if you think that the product is ready for the release (after merging this PR), feel free to perform the needed actions, as an owner and leader of the project)

viktorstrate (approved):

PR #976, opened 2024-06-27
Bump go version and modules.

googollee:
- Bump all Go modules to the newest version,
  - Except for pinning `libheif@v1.15.1` to match the C lib version in Debian bookworm.
  - The next Debian stable version `trixie` will have the newest `libheif`. We could bump the version at that time.
- Bump Go version to `1.22`.
  - `github.com/jackc/pgx/v5` requires to Go `1.20`, while Debian `bookworm` pins Go with `1.19`. We can't use Go provided by Debian `bookworm`.
  - As we need to use a new Go version, I prefer to choose the newest one.
  - Updated `go_wrapper` to `set_go_env` to work with `golang` image.
- Fixed legacy `ENV` in `Dockerfile`.

This PR deprecates #968 and #974.


Dockerfile
```diff
@@ -28,22 +28,21 @@ RUN npm ci --omit=dev --ignore-scripts \
...
-ENV CGO_ENABLED 1
+ENV CGO_ENABLED=1
 
 # Download dependencies
-RUN chmod +x /tmp/install_build_dependencies.sh \
-  && chmod +x /go/bin/go \
-  && /tmp/install_build_dependencies.sh \
+RUN chmod +x /tmp/scripts/*.sh \
+  && /tmp/scripts/install_build_dependencies.sh \
+  && . /tmp/scripts/set_go_env.sh \
```
kkovaletp, line 45: I'd prefer the more explicit form: `source /tmp/scripts/set_go_env.sh`: both work the same (as far as I know), but this is better readable and easier to understand)

.github/workflows/tests.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: ^1.16
+          go-version: ^1.20
```
kkovaletp, line 57: Don't we need to set here the same version, as in the Dockerfile? I guess that it is better to run unit tests with the same GO version than the PROD code

kkovaletp (commented):
I've added a few comments. They are not critical, so feel free to address or skip them)

Dockerfile
googollee: I tried `source` before. `source` doesn't work with `/bin/sh`. It's a command in `bash`. As `docker build` uses `/bin/sh` by default, it can't use it here.

.github/workflows/tests.yml
googollee: Yeah. I hesitated here which version to test. I agreed to test with the same version.

Dockerfile
kkovaletp: oh, good to know)
But isn't this should instruct the docker build to use bash:
`SHELL ["/bin/bash", "-o", "pipefail", "-c"]`

I've added it on the line 63, as a fix for a linter warning, but maybe it makes sense to add the same line before each RUN with more than 1 command inside?

googollee: I didn't check the detail. When I tried with `source`, the error showed it was using `/bin/sh`. I don't keep that log. Let me try it again.

kkovaletp (approved):
LGTM

kkovaletp:
@googollee, after merging this PR and updating my forked repo I found the next issue:
- Don't we need to update the GO version in the `go.mod` as well: https://github.com/photoview/photoview/blob/e36b09753a267ab36bb9e449a966d12bed15a36e/api/go.mod#L3 ?

googollee:
Oh, yes. In theory, we could keep it as the code in photoview doesn't require the higher version. In practise, I want to bump it to the latest `1.22`, and the following development could benefit from the latest go version. Photoview is an app and it's no need to take care of the compatible. I'll submit a PR quickly.

kkovaletp:
I've updated this version in my forked repo and it seems to work fine, but I got this warning from the CodeQL:
```
Invalid Go toolchain version
[View file](https://github.com/kkovaletp/photoview/blob/master/api/go.mod#L1-L1)
As of Go 1.21, toolchain versions [must use the 1.N.P syntax](https://go.dev/doc/toolchain#version).

1.22 in api/go.mod does not match this syntax and there is no additional toolchain directive, which may cause some go commands to fail.
```
So, I set the version to `1.22.0`

- Also, I've replaced the hardcoded GO version in the workflow with the `go-version-file: ${{ github.workspace }}/api/go.mod` (and moved this step 2 steps later: after the fetching branches step) to automatically use the same version without the need of always updating it here. Then I added the `cache: false` to remove the warning that it cannot find 'go.sum' and fail to cache dependencies. We don't need the caching here, as we have it as a separate step with the per-platform cache.

googollee:
Sounds good to me. @kkovaletp do you want to submit PR? I can review it.

kkovaletp:
@googollee , I mean that the tests workflow looks like this in my repo now:
```
    steps:
      - name: Check out code into the Go module directory
        uses: actions/checkout@v4

      - name: Fetch branches
        run: git fetch --all

      - name: Set up Go
        uses: actions/setup-go@v5
        id: go
        with:
          go-version-file: ${{ github.workspace }}/api/go.mod
          cache: false

      - name: Cache Go dependencies
        uses: actions/cache@v4
        with:
          path: ~/go/pkg/mod
          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
          restore-keys: |
            ${{ runner.os }}-go-
...
```

kkovaletp:
If you started the PR, you can go ahead) If not, I can do it if you'd prefer)
What is better for you)

PR #980, opened 2024-06-28
Update the minimal GO version for API module

kkovaletp:
Update GO version in the `go.mod` and read it by the job to set up the testing environment for unit tests

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -51,14 +51,18 @@ jobs:
         working-directory: api
 
     steps:
-      - name: Set up Go
-        uses: actions/setup-go@v2
-        with:
-          go-version: ^1.22
-        id: go
-
       - name: Check out code into the Go module directory
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
+
+      - name: Fetch branches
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -51,14 +51,18 @@ jobs:
...
         id: go
-
-      - name: Check out code into the Go module directory
-        uses: actions/checkout@v2
+        with:
+          go-version-file: ${{ github.workspace }}/api/go.mod
+          cache: false
 
       - name: Cache Go dependencies
         uses: actions/cache@v2
```
googollee, line 68: I see you updated this to `v4` in the comment. Do you want to do the same here?

googollee: As it's already updating the workflow, I think we could update all dependencies to the latest version.

kkovaletp: I thought that it is better to create another PR for that to not mix things together: in addition to updating the versions I planned to migrate at least independent serverless tools from my repo here as well to increase the code testing and lining level

googollee (approved):
LGTM

PR #981, opened 2024-06-28
Enhance Github Actions workflows

kkovaletp:
Migrate part of workflow from my forked repo to the main repo with updating versions of used actions:
- Updated versions of used actions, DB images, and the runner VM to the latest ones
- Added Dockle image scanner job (runs in `master` because needs to pull the image, which is pushed only for this branch) with reporting to the Security tab of the repo
- Optimized existing jobs with better steps configuration
- Added Anchore Grype dependency scanner job with reporting to the Security tab of the repo (all issues) and to the comment in the PR's discussion (only dependencies with fixed versions released)
- Added Hadolint linter of Dockerfile with reporting to the Security tab of the repo and to the comment in the PR's discussion
- Added NPM Audit step to UI tests with reporting to the Github Actions step output
- Added ESLint step to UI tests with reporting to the Github Actions step output
- Configured Unit tests to output results also in the standard JUnit format with coverage reports in Lcov (for UI tests) for future integrations

### DIFF
--- a/.dockleignore
+++ b/.dockleignore
@@ -1,0 +1,5 @@
+# Allow the `latest` tag
+DKL-DI-0006
+
+# Don't care about signatures
+CIS-DI-0005
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,29 +30,31 @@ jobs:
           git fetch --depth=1 origin +refs/tags/*:refs/tags/*
           echo "tag=$(git describe --tags `git rev-list --tags --max-count=1`)" >> $GITHUB_OUTPUT
 
+      - name: Prepare matrix
+        id: matrix
+        run: |
+          if [[ -n "${{ steps.gettag.outputs.tag }}" ]]; then
+            echo "matrix=$(jq -cn --arg tag '${{ steps.gettag.outputs.tag }}' '{ref: ["refs/heads/master", $tag]}')" >> $GITHUB_OUTPUT
+          else
+            echo "matrix=$(jq -cn --arg ref '${{ github.ref }}' '{ref: [$ref]}' )" >> $GITHUB_OUTPUT
+          fi
+
     outputs:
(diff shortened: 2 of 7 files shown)


kkovaletp:
@viktorstrate , I need your help with making the project configuration changes, required by actions, which try to output results to the PR comment:
- I need the `COMMENT_TOKEN` be generated by you and created in the project settings > Secrets and variables > both Actions and Dependabot categories (the same value in both)
- It should be your Personal Access Token with the next scopes granted: `repo:status`, `public_repo`
Then I'll uncomment the token parameter in the workflow to use it. 

It appears that with the default token it doesn't work even for regular PRs, not only Dependabot's

kkovaletp:
In general, this is a good idea.
But not all jobs can be converted easily to scripts, as some of them use and rely on GitHub functionality and actions, provided by GitHub.
Also, there might be potentially more complex code compared to the current list of actions (more efforts for maintenance).

Anyway, as I said, in general, this is a good idea, let's dive 1 level deeper and list what exactly you propose to convert to scripts: which jobs?

googollee:
At least scripts to test against MariaDB and PostgreSQL would be helpful. I'm thinking about some `docker-compose` files to set the environment with different databases and run tests in these env. But I'm not sure if it works well with GitHub Action. I can try to build these scripts this weekend.

For this PR, I see Code Analysis failed. After this is fixed, we could submit this PR first, and work on scripts in the next one.

kkovaletp:
The Code analysis is failed because it wants to create a comment in the discussion with the execution results, but it got 403 error. We need @viktorstrate to create a secret for this job, then I can use it to fix the error.

Also, I assume that this 403 error might be the cause of the workflows stop running in the repo completely( This might be some paranoid protection from GitHub, and if my assumption is correct, we need @viktorstrate to enable them back as well

PR #983, opened 2024-06-30
order root albums by path on settings page

Ajnasz:
Once you have more than a few paths assigned to a user, it becomes quite hard to oversee the list.

Personally, I would add my albums to Photoview one by one, so I can manage the processing better. For example, I can review and fix the face recognition for one new album at a time, and I don't need to deal with too much data.
Since I started to add the albums in a random order, it's relatively hard to find which album is added and which is not.

An easy fix is to order the photo paths by their name.

- Extended the query of the rootAlbums to order them by path

### DIFF
--- a/api/scanner/scanner_user.go
+++ b/api/scanner/scanner_user.go
@@ -55,7 +55,11 @@ func FindAlbumsForUser(db *gorm.DB, user *models.User, album_cache *scanner_cach
 	}
 
 	var userRootAlbums []*models.Album
-	if err := db.Where("id IN (?)", userAlbumIDs).Where("parent_album_id IS NULL OR parent_album_id NOT IN (?)", userAlbumIDs).Find(&userRootAlbums).Error; err != nil {
+	if err := db.
+		Where("id IN (?)", userAlbumIDs).
+		Where("parent_album_id IS NULL OR parent_album_id NOT IN (?)", userAlbumIDs).
+		Order("path ASC").
+		Find(&userRootAlbums).Error; err != nil {
 		return nil, []error{err}
 	}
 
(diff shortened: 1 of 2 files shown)


kkovaletp:
@Ajnasz, thanks for your PR with the improvement!
I like the idea of sorting the users' media library paths by name on the Settings page.
(As I'm not a developer) Please make sure that these changes are focused on the mentioned area and don't impact other product functionality, because I see that a lot of backend code is changed.

Ajnasz:
> Please make sure that these changes are focused on the mentioned area and don't impact other product functionality, because I see that a lot of backend code is changed.

Hi @kkovaletp!

Those are automatically generated changes, so basically didn't change the actual logic.
But I changed my commit and removed the unrelated diffs from the commit, so it should reflect only what is really needed.

api/scanner/scanner_user.go
```diff
@@ -55,7 +55,11 @@ func FindAlbumsForUser(db *gorm.DB, user *models.User, album_cache *scanner_cach
 	}
 
 	var userRootAlbums []*models.Album
-	if err := db.Where("id IN (?)", userAlbumIDs).Where("parent_album_id IS NULL OR parent_album_id NOT IN (?)", userAlbumIDs).Find(&userRootAlbums).Error; err != nil {
+	if err := db.
+		Where("id IN (?)", userAlbumIDs).
+		Where("parent_album_id IS NULL OR parent_album_id NOT IN (?)", userAlbumIDs).
+		Order("path asc").
```
googollee, line 61: Could you upper the key word `ASC`?

api/graphql/schema.graphql
```diff
@@ -280,7 +280,9 @@ type User {
   "All albums owned by this user"
   albums: [Album!]! @isAdmin
   "Top level albums owned by this user"
-  rootAlbums: [Album!]! @isAdmin
+  rootAlbums(
+    order: Ordering,
```
googollee, line 284: Do you plan to change the order in UI later? I don't think currently web UI has a way to change the order of the path. I think we could update the SQL query only, but not adding an order parameter in API.

Ajnasz: I didn't plan any other updates to this. I changed the pr, removed the changes from the API.

api/scanner/scanner_user.go
Ajnasz: changed

googollee (approved):

PR #985, opened 2024-06-30
Proposal for a restructuring of DB access to enable better testing; Feedback requested

AndGem:
Hey all,

I wanted to see if there is an easy way to restruture the code a bit so that DB interactions are separated from the business logic. This should allow easier testing.

I've prepared this change (that also introduces new dependencies which should be discussed) that showcases one way of doing this. Happy to hear some feedback around this.


googollee:
Hi @AndGem,

Mocking data is easily turning into below cases as the project evolves:

- Invalid test data when new conditions are added to the model
- Complex test data without good semantical meaning
- Unnecessary test data as they don't fail in tests

Also, mocking doesn't help in testing 3 databases. The mocking data is a 4th database, and it doesn't have ORM framework support.

For data testing, I'd like to have comprehensive methods of the model and inject necessary data through these methods during testing. If you want to test some cases without proper data, it's better to add a new method to the model and inject data through the method during testing.

Thanks for understanding these concerns.

AndGem:
Got your points, and thanks again for the feedback :). Will close the PR now.

PR #988, opened 2024-07-03
Remove loading env from tests.

googollee:
The env file doesn't exist in local test env and the local test won't succeed with it. I run tests with `TestMain()` as how other tests do, like `api/graphql/models/models_test.go` and load env with flags.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: ["*"]
+    branches: [master]
   pull_request:
     branches: [master]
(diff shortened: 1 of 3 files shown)


kkovaletp (approved):
Approving the PR mostly based on the passed run, which includes this test, also with some high-level understanding of the code change.
But anyway, this is just the test, not the app code, so even if there is some non-blocking issue, we could fix it later)

kkovaletp:
@jordy2254 FYI, feel free to provide your thoughts on this PR even after it is merged: we always can create a new one to improve the code if needed

PR #989, opened 2024-07-06
Add nodemon for ui.

googollee:
`npm run mon` launches `nodemon` to monitor ui code changes and `build`/`start` the service again when changes happen.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -10,7 +10,7 @@
   "description": "UI app for Photoview",
   "scripts": {
     "start": "vite",
-    "mon": "nodemon --watch './**' --ext 'ts,json,tsx,js,html,css' --ignore './dist/' --exec 'npm start'",
+    "mon": "nodemon",
     "build": "vite build",
     "lint": "eslint ./src --max-warnings 0 --config .eslintrc.js",
     "test": "vitest",
@@ -18,6 +18,19 @@
     "genSchemaTypes": "apollo client:codegen --target=typescript --globalTypesFile=src/__generated__/globalTypes.ts --passthroughCustomScalars && prettier --write */**/__generated__/*.ts",
     "extractTranslations": "i18next -c i18next-parser.config.js",
     "prepare": "(cd .. && ./ui/node_modules/.bin/husky install)"
+  },
+  "nodemonConfig": {
+    "watch": [
+      "./**"
+    ],
+    "ext": "ts,json,tsx,js,html,css",
+    "exec": "npm start",
+    "ignore": [
+      "./dist/"
+    ],
+    "execMap": {
+      "ts": "ts-node-esm"
+    }
   },
   "dependencies": {
     "@apollo/client": "^3.6.9",
--- a/ui/package.json
+++ b/ui/package.json
@@ -10,6 +10,7 @@
   "description": "UI app for Photoview",
   "scripts": {
     "start": "vite",
+    "mon": "nodemon --watch './**' --ext 'ts,json,tsx,js,html,css' --ignore './dist/' --exec 'npm run build -- --base=/ && npm start'",
     "build": "vite build",
     "lint": "eslint ./src --max-warnings 0 --config .eslintrc.js",
     "test": "vitest",
@@ -74,6 +75,7 @@
     "husky": "^8.0.1",
     "i18next-parser": "^8.13.0",
     "lint-staged": "^13.0.3",
+    "nodemon": "^3.1.4",
     "vitest": "^0.17.1"
   },
   "overrides": {
--- a/ui/package.json
+++ b/ui/package.json
@@ -19,19 +19,6 @@
     "extractTranslations": "i18next -c i18next-parser.config.js",
     "prepare": "(cd .. && ./ui/node_modules/.bin/husky install)"
   },
-  "nodemonConfig": {
-    "watch": [
-      "./**"
-    ],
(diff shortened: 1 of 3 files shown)


googollee:
As I'm working on macOS, the docker container is running in a VM (like lima VM or Docker Desktop). FS notification doesn't work in this situation. Using `--legecy-watch` with `nodemon` could work well on macOS, but it's pretty slow. It's better to work with native env on macOS.

It works well on a Linux env with a container.

viktorstrate (approved):

kkovaletp:
Thanks for adding the file-watching web server for the dev environment! It would be a great addition to your other PR with the development environment inside containers.

The current configuration should work, however, I think that it will be slow, as on each file change it will run the full build process. Instead, the approach, described in the https://typescript.tv/hands-on/effortless-nodemon-setup-with-typescript-and-esm/ article, might be much faster and more optimal for the development case.

What do you think?

googollee:
I'd like to try.

googollee:
@kkovaletp done with `ts-node-esm`. I tried locally changing some text in `tsx` files. It works well.

ui/package.json
```diff
@@ -10,6 +10,7 @@
   "description": "UI app for Photoview",
   "scripts": {
     "start": "vite",
+    "mon": "nodemon --watch './**' --ext 'ts,json,tsx,js,html,css' --ignore './dist/' --exec 'npm start'",
```
kkovaletp, line 13: I guess that with the latest changes (adding the `ts-node`), the command can be simplified:
- `watch`, `ext`, and `ignore` params can be moved to the `nodeamon.json`, which would be the logical place for them, holding the complete config in 1 place
- instead of the `exec` option, it is now much better to put the path to the main TS file (as far as I understand, it is the `src/index.tsx`) - in this way we'd utilize the TypeScript JIT functionality and faster reload on changes

ui/package.json
```diff
@@ -63,6 +63,7 @@
     "@testing-library/jest-dom": "^5.16.4",
     "@testing-library/react": "^13.3.0",
     "@testing-library/user-event": "^14.2.1",
+    "@types/node": "^20.14.10",
```
kkovaletp, line 66: As the UI project is currently working under Node 18, I think that the version of this dependency has to be aligned with it: `"@types/node": "^18.15.3",`

kkovaletp (changes requested):
Couple of missing/forgotten small changes)

ui/package.json
googollee: Removed this dependency.

ui/package.json
googollee: I put nodemon config in `package.json`. I don't know which one is better, but in `package.json` putting all configure together.

googollee: Please reopen this thread if you have any other ideas.

kkovaletp: Hm, if the config is successfully loaded from the `package.json` I'm fine with it, however, looking for it there won't be my 1st guess))

But you still use the `--exec 'npm start'` - doesn't it work with removing this parameter completely and instead of it  put the `src/index.tsx` to the nodeamon command (as in the article):
`"mon": "nodemon src/index.tsx",` 

Without this change adding TS support (looks like) doesn't add much value, because nodeamon doesn't work in the native mode.

Please let me know if you've tried that and it doesn't work for some reason.

googollee: It's not the same way to launch a server in `photoview` as what that document says. With `nodemon src/index.tsx`, `npm run mon` fails:

```
% npm run mon

> photoview-ui@0.0.1 mon
> nodemon src/index.tsx

[nodemon] 3.1.4
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): **
[nodemon] watching extensions: ts,json,tsx,js,html,css
[nodemon] starting `node src/index.tsx`
(node:4321) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/Users/googol/code/photoview/ui/src/index.tsx:1
import React from 'react'
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at wrapSafe (node:internal/modules/cjs/loader:1383:18)
    at Module._compile (node:internal/modules/cjs/loader:1412:20)
    at Module._extensions..js (node:internal/modules/cjs/loader:1551:10)
    at Module.load (node:internal/modules/cjs/loader:1282:32)
    at Module._load (node:internal/modules/cjs/loader:1098:12)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:215:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:158:5)
    at node:internal/main/run_main_module:30:49

Node.js v22.4.0
[nodemon] app crashed - waiting for file changes before starting...
```

googollee: BTW, I believe there is a better way to set up `nodemon` with Typescript. I'm not a frontend expert, so I'd like to wait for others to contribute better.

kkovaletp: Oh, now I see...
Sure, in these circumstances we should go with the working (even not the most efficient) setup. Later, when we'll have a Frontend developer, interested in improving the setup, it will be an easy fix (I haope).
Approving the PR.

kkovaletp (approved):
LGTM

PR #992, opened 2024-07-07
feat: harden security on docker compose examples

rare-magma:
This PR hardens security on the docker compose examples by dropping unnecessary capabilities and  ensuring the containers are not granted any new privileges after initialization.
See more info on rules 2, 3, 4, 6, 7 and 8 in https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-3-limit-capabilities-grant-only-specific-capabilities-needed-by-a-container

### DIFF
--- a/docker-compose example/docker-compose.example.yml
+++ b/docker-compose example/docker-compose.example.yml
@@ -17,9 +17,21 @@ services:
       mariadb:
         condition: service_healthy
     ## Security options for some restricted systems
+    read_only: true
+    cap_drop:
+      - ALL
     security_opt:
-      - seccomp:unconfined
-      - apparmor:unconfined
+      - no-new-privileges:true
+      # not necessary since Docker Engine v20
+      #   - seccomp:unconfined
+      #   - apparmor:unconfined
+    ## Uncomment to enforce resource usage limits
+    # deploy:
+    #   resources:
+    #     limits:
+    #       cpus: "4"
+    #       memory: 4g
+    #       pids: 1024
     environment:
       PHOTOVIEW_DATABASE_DRIVER: ${PHOTOVIEW_DATABASE_DRIVER}
       ## Comment out the next variable in the case PHOTOVIEW_DATABASE_DRIVER is set to `sqlite` or `postgres` in the .env
@@ -104,11 +116,26 @@ services:
     container_name: photoview-mariadb
     restart: unless-stopped
     stop_grace_period: 5s
+    user: "999:999"
     ## Optimized MariaDB startup command for better performance and compatibility
     command: mariadbd --innodb-buffer-pool-size=512M --transaction-isolation=READ-COMMITTED --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci --max-connections=512 --innodb-rollback-on-timeout=OFF --innodb-lock-wait-timeout=120
-    security_opt: ## see https://github.com/MariaDB/mariadb-docker/issues/434#issuecomment-1136151239
-      - seccomp:unconfined
-      - apparmor:unconfined
+
+    ## Security options for some restricted systems
+    cap_drop:
+      - ALL
+    security_opt:
+      - no-new-privileges:true
+     # not necessary since Docker Engine v20 see https://github.com/MariaDB/mariadb-docker/issues/434#issuecomment-1136151239
+      #   - seccomp:unconfined
+      #   - apparmor:unconfined
+
+    ## Uncomment to enforce resource usage limits
+    # deploy:
+    #   resources:
+    #     limits:
+    #       cpus: "4"
+    #       memory: 1g
+    #       pids: 1024
     ## Uncomment next 2 lines if you want to access the database directly
     # ports:
       # - "3306:3306"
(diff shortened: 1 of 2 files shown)


docker-compose example/docker-compose.example.yml
```diff
@@ -17,9 +17,21 @@ services:
...
+      - ALL
     security_opt:
-      - seccomp:unconfined
-      - apparmor:unconfined
+      - no-new-privileges:true
+      # not necessary since Docker Engine v20
+      #   - seccomp:unconfined
+      #   - apparmor:unconfined
+    ## Uncomment to enforce resource usage limits
+    # deploy:
```
kkovaletp, line 29: According to the https://docs.docker.com/compose/compose-file/deploy/, the `deploy` section is used only by Docker Swarm (something like Kubernetes, but from Docker).
We currently don't officially support Docker Swarm and for regular `docker compose` command it is ignored, so currently I don't see a value from adding this template here. Even more: it will create a false impression that there is a way to set some limits, while it isn't true.

docker-compose example/docker-compose.example.yml
```diff
@@ -104,11 +116,26 @@ services:
     container_name: photoview-mariadb
     restart: unless-stopped
     stop_grace_period: 5s
+    user: "999:999"
```
kkovaletp, line 119: Why would we need this user IDs definition for the DB service on a PROD setup? The DB is not exposed by default to the world, it creates and manages the DB-related files under root ownership which is additional protection from some unintended modification on the host system.
Do you see a real need/benefit in this change?

docker-compose example/docker-compose.example.yml
```diff
@@ -104,11 +116,26 @@ services:
...
     ## Optimized MariaDB startup command for better performance and compatibility
     command: mariadbd --innodb-buffer-pool-size=512M --transaction-isolation=READ-COMMITTED --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci --max-connections=512 --innodb-rollback-on-timeout=OFF --innodb-lock-wait-timeout=120
-    security_opt: ## see https://github.com/MariaDB/mariadb-docker/issues/434#issuecomment-1136151239
-      - seccomp:unconfined
-      - apparmor:unconfined
+
```
kkovaletp, line 122: Please remove empty lines inside the same service definition: it is a styling issue, as in this way we could visually identify where the definition of a previous service ends and a new one starts: we have pretty long definitions (including comments), so I'd like to keep the structure as clear as possible)

kkovaletp (changes requested):
@rare-magma thanks for your contribution!
Security topic is an important part of a self-hosted web service and it's great that you propose some improvements here!

I've reviewed the PR and added some comments and a few change requests. Hope to get your answers and updates for them)

docker-compose example/docker-compose.example.yml
rare-magma: @kkovaletp I think the official documentation is misleading, the deploy section does impact docker compose's behaviour (without using Docker Swarm), if I set a low memory limit I see it getting it enforced by the OOM killer in case the container exceeds the threshold

docker-compose example/docker-compose.example.yml
rare-magma: @kkovaletp setting this user id makes the container's entrypoint not modify the database folders to an arbitrary UID (like 0 for root)  here: https://github.com/MariaDB/mariadb-docker/blob/11135d071fd1fe355b1f7fa99b9d3b4a59bb5225/docker-entrypoint.sh#L200-L221
Doing this negates the requirement to add the `CHOWN` capability.

I think your point is valid for the scenario where ab user is mapping a host folder to the DB folder (instead of using a volume)

docker-compose example/docker-compose.example.yml
rare-magma: done in https://github.com/photoview/photoview/pull/992/commits/0719a1e3b663fc03aae700bd43fb3b8c0eccbe5f

docker-compose example/docker-compose.example.yml
rare-magma: you can also confirm this by having a look at the `docker stats` output, the `LIMIT` column will correspond to the values set here

docker-compose example/docker-compose.example.yml
```diff
@@ -17,9 +17,21 @@ services:
...
         condition: service_healthy
     ## Security options for some restricted systems
+    read_only: true
+    cap_drop:
+      - ALL
     security_opt:
-      - seccomp:unconfined
-      - apparmor:unconfined
+      - no-new-privileges:true
+      # not necessary since Docker Engine v20
```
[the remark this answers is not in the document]
kkovaletp: I test on MacOS 14.5 x64

docker-compose example/docker-compose.example.yml
kkovaletp: Thanks, I'll test this as well

docker-compose example/docker-compose.example.yml
kkovaletp: Mapping a host folder as a volume to Photoview services is the default option (part of the examples) because in this way users have more control over hosting these volumes on their host file systems depending on the available limitations and use case.

rare-magma: @kkovaletp I've replaced the hardcoded user with the necessary capabilities in https://github.com/photoview/photoview/pull/992/commits/82d317f00b817e945117dd4ea2f7bd4b9e2c6248

PR #943, opened 2024-05-03
Create CONTRIBUTING.md

blakeschafer:
Creating a resource file for any developer who wants to contribute to this open-source project. Adding more details than what was already provided, giving a detailed walkthrough for newcomers on how to make changes in the code.

PR #995, opened 2024-07-09
Update Set up development environment.

googollee:
Follow up [#993](https://github.com/photoview/photoview/pull/993) to update development environment setup doc.

It's related with [#943](https://github.com/photoview/photoview/pull/943) but not the same part of doc.


README.md
```diff
@@ -158,43 +158,68 @@ Setting up and configuring of all these protections depends on and requires a lo
 
 ## Set up development environment
 
-### Local setup
+### Developing dependencies
+
+- API
+  - Required packages:
```
kkovaletp, line 164: I see that this list is a bit different from the install dependencies scripts from the previous PR (like, I don't see here `libheif` and `g++`, but see the `libexif`, missing in the script). Please double-check and let me know why is that)
I assume that there only the dependencies, needed to build the code, are listed, but since you created scripts to set the local dev env up, I think that there should be the possibility or an additional script to install the developing dependencies as well

README.md
```diff
@@ -158,43 +158,68 @@ Setting up and configuring of all these protections depends on and requires a lo
...
+
+In Debian/Ubuntu, install dependencies:
 
-1. Install a local mysql server, and make a new database
-2. Rename `/api/example.env` to `.env` and update the `MYSQL_URL` field
-3. Rename `/ui/example.env` to `.env`
+```sh
+$ sudo apt install golang libexif-dev libdlib-dev libjpeg-dev libblas-dev libatlas-base-dev liblapack-dev libdlib-dev # For API requirement
+$ sudo apt install reflex sqlite3 # For API optional tools
+$ sudo apt install node # For UI
```
kkovaletp, line 184: I'd really not recommend installing Node from apt or Homebrew. There will be a lot of inconveniences and headaches in this case. Instead, I'd recommend installing NVM (https://github.com/nvm-sh/nvm) and install the latest Node LTS 18 from its repo, following by `nvm install-latest-npm` command

README.md
```diff
@@ -158,43 +158,68 @@ Setting up and configuring of all these protections depends on and requires a lo
...
+In macOS, install dependencies:
 
-Make sure [golang](https://golang.org/) is installed.
+```sh
+$ brew install golang libheif dlib jpeg # For API
+$ brew install reflex sqlite3 # For API optional tools
+$ brew install node # For UI
+```
 
-Some C libraries are needed to compile the API, see [go-face requirements](https://github.com/Kagami/go-face#requirements) for more details.
```
kkovaletp, line 171: Why did you remove the Face-recognition dependency installation link? don't we need it anymore?

README.md
```diff
@@ -158,43 +158,68 @@ Setting up and configuring of all these protections depends on and requires a lo
...
-sudo apt-get install libdlib-dev libblas-dev libatlas-base-dev liblapack-dev libjpeg-turbo8-dev libheif-dev
-# Debian
-sudo apt-get install libdlib-dev libblas-dev libatlas-base-dev liblapack-dev libjpeg62-turbo-dev libheif-dev
-# macOS
-brew install dlib libheif
+### Local setup
 
-```
+1. Rename `/api/example.env` to `.env`
+  - Update `PHOTOVIEW_SQLITE_PATH` if you don't want to put sqlite file under `/api`
```
kkovaletp, line 200: ... or set a different DBMS driver, followed by commenting our the SQLite path variable and uncommenting the corresponding connection string variable for the new driver (and update its value to correspond to your setup)

kkovaletp (commented):
@googollee thanks for preparing this follow-up PR!
Here are my 1st comments after the fast review:

README.md
```diff
@@ -158,43 +158,68 @@ Setting up and configuring of all these protections depends on and requires a lo
...
+    - `libjpeg`
+    - `libblas` 
+    - `libcblas`, recommending using `libatlas-base`
+    - `liblapack`
+  - Optional tools during developing:
+    - `reflex`
+    - `sqlite`
+- UI
+  - Required packages:
+    - `node` >= 18
```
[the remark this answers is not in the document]
googollee: Done.

README.md
googollee: Updated.

README.md
googollee: It's a typo.

I hope to show depending packages clearly in README. Also `scripts/apt/install_build_dependencies.sh` contains cleanup commands. I don't know if it's good to let developers run it directly.

For `homebrew` scripts, I may remove `scripts/homebrew/install_*_dependencies.sh` as the content are already in README. I'll keep `scripts/homebrew/set_compiler_env.sh` as it should be run every time before building API.

README.md
googollee: They are already mentioned in the dependencies session. I make it more clear.

README.md
googollee: Done

README.md
googollee: I updated as I said. Let me know what you think.

googollee:
Added the session of `Set up development environment with Docker`.

README.md
kkovaletp: > `scripts/apt/install_build_dependencies.sh` contains cleanup commands. I don't know if it's good to let developers run it directly

That is a good catch! Definitely, not all of those commands should be executed on a dev env, as some of them are destructive (they are fine for an image, where we're sure there will be no future installations):
- `rm -rf /var/lib/apt/lists/*`
- `apt-get purge -y ${BUILD_DEPENDS[@]}`
These 2 commands should not be executed on a long-leaving environment, like a dev system or a server.

kkovaletp: > I may remove `scripts/homebrew/install_*_dependencies.sh` as the content are already in README

Readme - is a doc, while script helps to simplify the setup by running 1 command instead of many.

README.md
```diff
@@ -171,43 +171,82 @@ Remember, every contribution counts. Let's make this project better together! 
...
+  - Required packages:
+    - `golang` >= 1.22
+    - `g++`
+    - `libc-dev`
+    - `libheif` >= 1.15.1
+    - [go-face Requirements](https://github.com/Kagami/go-face#requirements) 
+        - `dlib`
+        - `libjpeg`
+        - `libblas` 
+        - `libcblas`, recommending using `libatlas-base`
```
kkovaletp, line 186: - `libcblas`, recommend using `libatlas-base` instead

README.md
```diff
@@ -171,43 +171,82 @@ Remember, every contribution counts. Let's make this project better together! 
...
+You can install `node` with other package manager if you like.
+
+### Local setup
+
+1. Rename `/api/example.env` to `.env`
+  - Update `PHOTOVIEW_SQLITE_PATH` if you don't want to put sqlite file under `/api`
+  - To set a different DBMS driver
+    - Comment the SQLite path variable
+    - Update `PHOTOVIEW_DATABASE_DRIVER` with your driver
+    - Uncommenting the corresponding connection string variable for the new driver
```
kkovaletp, line 227: Uncommenting > Uncomment

Dockerfile
```diff
@@ -37,7 +37,7 @@ ARG TARGETPLATFORM
 # See for details: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
-COPY scripts/apt /app/scripts/apt
+COPY scripts /app/scripts
```
kkovaletp, line 40: I don't like the idea of mixing apt and brew scripts, and putting them in the `/scripts` folder, so you have to copy the whole folder to the image, including brew scripts sub-folder (which is not going to be used there in any case)

README.md
```diff
@@ -171,43 +171,82 @@ Remember, every contribution counts. Let's make this project better together! 
...
+  - Required packages:
+    - `node` = 18
+
+In Debian/Ubuntu, install dependencies:
 
-1. Install a local mysql server, and make a new database
-2. Rename `/api/example.env` to `.env` and update the `MYSQL_URL` field
-3. Rename `/ui/example.env` to `.env`
+```sh
+$ sudo apt install golang g++ libc-dev libheif-dev libdlib-dev libjpeg-dev libblas-dev libatlas-base-dev liblapack-dev # For API requirement
```
kkovaletp, line 198: always propose to run `sudo apt update` before installing something, as APT, in opposite to Brew, doesn't run the `update` if it wasn't run for long time

kkovaletp (commented):
A few more comments after a fast review of the recent changes.
Sorry, still didn't have time/energy to perform the complete end-2-end review(

README.md
googollee: Done.

README.md
googollee: Done.

README.md
googollee: I agree `README` is a doc. It should tell the reader details, but not an emergency playbook to operate blindly. Here, the content shows exactly what needs to be installed. If we add or remove any dependencies in the future, it's easy to see the change instead of in scripts.

Also, putting these commands into scripts won't reduce the running commands, if you don't want to install extra packages during building. It requires at least two scripts, build dependencies, and dev tools, because dev tools are not required during building.

README.md
googollee: I believe a Debian/Ubuntu user should know the basic usage of `apt`. It'll depend on their knowledge to run it.

googollee: I'd like to show my follow-up PR to recommend users setting up development environment with Docker, see https://github.com/googollee/photoview/blob/dev-docker/README.md#set-up-docker-development-environment.

Only advanced user goes to local environment, and we could let them decide when to run `apt update`.

kkovaletp: > Only advanced user goes to a local environment

This might sense, but not necessary. Some users do not like the Docker dev workflow and prefer to build and test locally - this is just 1 example of when users will prefer to use local env even if they are not as skilled, as you assume, they should be)

Adding `apt update && ` prefix to installation commands wouldn't cost anything, but will solve a lot of potential issues

README.md
```diff
@@ -171,43 +171,82 @@ Remember, every contribution counts. Let's make this project better together! 
...
+    - `libheif` >= 1.15.1
+    - [go-face Requirements](https://github.com/Kagami/go-face#requirements) 
+        - `dlib`
+        - `libjpeg`
+        - `libblas` 
+        - `libcblas`, recommended using `libatlas-base`
+        - `liblapack`
+  - Optional tools during developing:
+    - `reflex`
+    - `sqlite`
```
kkovaletp, line 190: As we add these tools to the list, but mark them optional, I'd like to see a short explanation of what benefits they might add to a developer's experience: why and for what use case do we recommend using them?

Like:
  - `reflex` - a source code monitoring tool, which automatically rebuilds and restarts the server, running from the code in development: https://github.com/cespare/reflex
  - `sqlite` - the SQLite DBMS, useful to interact with Photoview's SQLite DB directly if you use it in your development environment

README.md
```diff
@@ -171,43 +171,82 @@ Remember, every contribution counts. Let's make this project better together! 
...
+```sh
+$ nvm install 18
+$ nvm use 18
 ```
 
+You can install `node` with other package manager if you like.
+
+### Local setup
+
+1. Rename `/api/example.env` to `.env`
```
kkovaletp, line 222: ... and optionally modify it if needed according to the inline comments

kkovaletp (commented):
This time I did e2e review) hope to detect everything this time)
BTW, as now we have the Contributing.md, doesn't it make sense to move the Setup dev env section from the main Readme to it? Most of the users don't need those instructions...

README.md
googollee: Done.

README.md
googollee: Done.

README.md
googollee: Done

Dockerfile
googollee: I expand the homebrew setup in README.

kkovaletp: This is good, however, my comment was more about the docker image. It is still not a big issue, as the brew-related files are copied to the build stages, not to the final image, but still, it looks a bit messy to me)

If you could briefly explain what is the benefit of this, it might be more clear to me)

googollee: I don't have too many thought on this. I hope to keep the script as simple as possible. Copy all files under `scripts/` so the build script don't need to know the structure.

I don't have experience of writing a good development doc. I just write down what I did to set up my local environment.

kkovaletp (approved):

PR #997, opened 2024-07-17
Add `Set up Docker development environment` in README.md

googollee:


README.md
```diff
@@ -169,9 +169,46 @@ It includes information on our code of conduct, the process for submitting pull
...
+```sh
+$ docker build --target dev-ui -t photoview-ui .
+$ docker build --target dev-api -t photoview-api .
+```
+
+### Start API server with Docker
+
+```sh
+$ cp ./api/example.env ./api/.env
+$ docker run --rm -it -v `pwd`:/app -p 4001:4001 photoview-api /bin/bash
```
kkovaletp, line 191: Here and for the UI: I think creating a dev-compose file and implementing these steps inside is better. Then a user needs just to run 1 single command to manage the env, while still being able to access and manage each of the containers individually if needed.

Dockerfile
```diff
@@ -60,8 +60,22 @@ RUN chmod +x /app/scripts/*.sh \
   # Build api source
   && go build -v -o photoview .
 
-### Copy api and ui to production environment ###
-FROM --platform=${BUILDPLATFORM:-linux/amd64} debian:bookworm-slim AS final
+### Build dev image for UI ###
+FROM ui AS dev-ui
```
kkovaletp, line 64: Because of the line 29 (`RUN npm ci --omit=dev --ignore-scripts`) this won't work for the dev purposes. UI dev image should be built with `npm ci` command in the sequence, but adding another build stage to this Dockerfile, which will be executed every time on CI and never used there, looks like wasting time and resources (as docker build already runs long time on CI).

So, in this case, it looks like it is better to create a separate Dockerfile inside the `ui` repo folder and reference it for dev-image build purposes.

kkovaletp (commented):
A few issues with the current workflow...

README.md
googollee: Not actually. I updated `Dockerfile` to run a server with 1 command now.

Dockerfile
googollee: Updated.

`dev-ui` and `dev-api` won't be built when building `release` image, as `release` image doesn't depend on those stages. So it doesn't waste any more time during CI.

googollee:
```sh
$ docker build -t photoview .
[+] Building 138.3s (23/23) FINISHED                                                                   docker:default
 => [internal] load build definition from Dockerfile                                                             0.0s
 => => transferring dockerfile: 3.29kB                                                                           0.0s
 => [internal] load metadata for docker.io/library/debian:bookworm-slim                                          0.0s
 => [internal] load metadata for docker.io/library/node:18                                                       0.0s
 => [internal] load metadata for docker.io/library/golang:1.22-bookworm                                          0.0s
 => [internal] load .dockerignore                                                                                0.0s
 => => transferring context: 325B                                                                                0.0s
 => [release 1/7] FROM docker.io/library/debian:bookworm-slim                                                    0.0s
 => [internal] load build context                                                                                0.2s
 => => transferring context: 49.04MB                                                                             0.2s
 => [ui 1/4] FROM docker.io/library/node:18                                                                      0.0s
 => [api 1/5] FROM docker.io/library/golang:1.22-bookworm                                                        0.0s
 => CACHED [release 2/7] COPY scripts/install_runtime_dependencies.sh /app/scripts/                              0.0s
 => [release 3/7] RUN groupadd -g 999 photoview   && useradd -r -u 999 -g photoview -m photoview   && chmod +  119.4s
 => CACHED [api 2/5] COPY scripts /app/scripts                                                                   0.0s
 => [api 3/5] COPY api /app/api                                                                                  0.6s
 => CACHED [ui 2/4] COPY ui /app/ui                                                                              0.0s
 => CACHED [ui 3/4] WORKDIR /app/ui                                                                              0.0s
 => CACHED [ui 4/4] RUN npm ci --omit=dev --ignore-scripts   && npm run build -- --base=/                        0.0s
 => [api 4/5] WORKDIR /app/api                                                                                   0.1s
 => [api 5/5] RUN chmod +x /app/scripts/*.sh   && source /app/scripts/set_compiler_env.sh   && /app/scripts/i  133.8s
 => [release 4/7] WORKDIR /home/photoview                                                                        0.4s
 => [release 5/7] COPY api/data /app/data                                                                        0.4s
 => [release 6/7] COPY --from=ui /app/ui/dist /app/ui                                                            0.3s
 => [release 7/7] COPY --from=api /app/api/photoview /app/photoview                                              0.1s 
 => exporting to image                                                                                           3.1s 
 => => exporting layers                                                                                          3.0s 
 => => writing image sha256:04f3d0fd3d87d63b49744d734010b5bce2b86c9198362091043fcdbc5eae71ce                     0.0s 
 => => naming to docker.io/library/photoview 
```

From the staging prefix in `[]`, you can see it only builds `ui`/`api`/`release`, but no `dev-ui`/`dev-api`.

Dockerfile
```diff
@@ -4,9 +4,6 @@ FROM --platform=${BUILDPLATFORM:-linux/amd64} node:18 AS ui
 # See for details: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
-ARG REACT_APP_API_ENDPOINT
-ENV REACT_APP_API_ENDPOINT=${REACT_APP_API_ENDPOINT}
```
kkovaletp, line 8: Why did you remove these 2 lines?

README.md
kkovaletp: If you need to run just the backend server, that would be only 1 command, but if you need to run the frontend, it requires running the backend as well, so in this case, it would be 2 commands in 2 separate shells (as both run interactively).
I see the benefit of running a server interactively: you don't need to run the `docker logs` command in addition to see what is going on there. But also, if you're developing a frontend feature, most probably you're not interested in a backend log. Yes, you can run it in the detached mode. But isn't it better to provide a single command to run both servers? Just asking)

Dockerfile
googollee: With these 2 lines, `.env` doesn't work with a container. As the release stage doesn't have this env variable and the value is an empty string during building, I don't see the reason to keep them.

README.md
googollee: I don't know how a frontend developer works. I want to show more details to new contributors.

I have [another change](https://github.com/photoview/photoview/compare/master...googollee:photoview:test-scripts) for testing with Docker Compose. I don't want to put two compose files. Let me see if I can combine them with one `dev-compose.yaml`.

Dockerfile
googollee: As I work on a combined docker compose file, I can keep these lines and update env in the compose file. I still think they are not used, but I'd like to delegate to a frontend developer to decide.

kkovaletp: For sure, it is better to decide on the Frontend side. I'm also not 100% sure that this var is needed, but since it was there before (even my) changes to the Dockerile, I'd prefer it to be kept until we have a confirmation from a Frontend that it is not needed for any possible use case.

As far as I see, .env can contain whatever variables you want. If a var is not there, it will be empty, if there is a var, not referenced in the compose or Dockerfile, it will be ignored.

This particular var is not for .env and compose. It is for building the frontend inside the build stage on the CI as a 1st prio.

googollee: Revert this change.

README.md
googollee: Done. Please take another look and let me know what you think about it.

kkovaletp: I'm not requesting you to provide a dev-compose.yml. I just write my thoughts and vision based on my understanding and past experience of a frontend dev workflow just to make sure that we took all the aspects into account. Without having a frontend developer in our team we could only guess and assume what might work better for this role...

Dockerfile
kkovaletp: just a question: as now the `dev-ui` is an empty stage, why exactly do we need it instead of referencing the `ui` (parent) stage in the dev-compose and individual `docker run` command?

README.md
```diff
@@ -169,9 +169,33 @@ It includes information on our code of conduct, the process for submitting pull
...
+$ docker compose -f dev-compose.yaml build dev-ui dev-api
+```
+
+### Start API and UI server with Docker Compose
+
+It may take a long time to build dependencies when launching servers first time.
+
+```sh
+$ docker compose -f dev-compose.yaml up dev-api dev-ui
+```
```
kkovaletp, line 192: Sorry for the back-and-forth process, but I didn't mean to replace the `docker run` commands with the single compose command, as for sure `docker run` commands are useful for certain cases. Can you please provide both ways in the doc here, so contributors can decide which is better for them?

Dockerfile
googollee: I know it's empty. I keep it as a symmetry stage between `api` and `ui`.

README.md
googollee: Added below.

kkovaletp (approved):

PR #1003, opened 2024-07-23
Add scripts for unit tests with all databases.

googollee:
- Add services for testing in `dev-compose.yaml`.
- Add scripts to run tests.
- Update GitHub test workflow with test scripts.


kkovaletp:
@googollee, thanks for this PR.
It might provide noticeable improvements in the usability of test execution locally on a dev env.

However, I don't like the idea of running the tests in the same way on CI: 
- your implementation doesn't use the Github caching, so every time all the dependencies are going to be installed as in the 1st time
- tests are going to be executed inside the docker image, which should be slower
- the setup in the case of CI execution is like a russian doll: GitHub workflow defines execution branches according to the DB names and calls SH script, where there is a `docker compose` command, which runs the defined set of services, inside which the tests are executed, writing the results to a volume, representing the checked out code on the Github runner.

Even taking aside the not optimal from the GitHub CI perspective implementation, debugging some tricky cases with the CI execution or enhancing the pipeline with additional linters or other useful tools would be a nightmare...

Sorry, but I'll not approve this change in the CI workflow, so I added @jordy2254 as a reviewer for this PR - maybe he has another opinion or some thoughts...

googollee:
I want to unify the test between different environment, including GitHub workflow. We should test with the same setup to avoid issues related with a special environment.

As our image is built with `debian:bookworm-slim`, GitHub workflow runs tests in ubuntu, which has a lot of small difference. All dependencies are different versions.

> your implementation doesn't use the Github caching, so every time all the dependencies are going to be installed as in the 1st time

I didn't see any steps using the cache in the previous workflow. I believe the cache step is not used. See `if: steps.cache-primes.outputs.cache-hit != 'true'` in [the example](https://github.com/actions/cache?tab=readme-ov-file#restoring-and-saving-cache-using-a-single-action) about how to use cache.

> tests are going to be executed inside the docker image, which should be slower

Running in a container is not slow in Linux. It's a normal process running in a special cgroup.

It adds a extra step to download the image before testing, which slow the workflow. 

> the setup in the case of CI execution is like a russian doll

CI scripts should not contain a complex logic. It's hard to reproduce issues locally. Calling a script makes it easy to reproduce the same thing in different environment. Running in a container through `docker compose` ensure tests won't depend on special CI environment, but run with a target environment.

kkovaletp:
> I want to unify the test between different environment, including GitHub workflow. We should test with the same setup to avoid issues related with a special environment.

I completely understand this idea and in general, I'd agree with it, but not in the case of CI: this is a special case and it can run a bit differently to be optimized for speed, simplicity of debugging, and maintainability.

> As our image is built with `debian:bookworm-slim`, GitHub workflow runs tests in ubuntu, which has a lot of small difference. All dependencies are different versions.

I don't see an issue here, as this is a supported OS for our product and we do not ship it as a docker image only. Running tests on a different platform additionally extends the coverage.

> I didn't see any steps using the cache in the previous workflow. I believe the cache step is not used.

Please check my opened PR #981, where I fixed many similar bugs with the steps configuration.

> Running in a container is not slow in Linux. It's a normal process running in a special cgroup.

Agree, but this is not a typical Docker setup: Each job in the GitHub Actions runs in its own VM, where each step is running as a Docker container. If a step runs Docker inside, it is actually the Docker-in-Docker setup. So, Because of such complexity, I'd expect some performance degradation.

> CI scripts should not contain a complex logic.

Again, I agree with you, but the CI pipeline should be clear and easy to maintain, debug, or extend with some new steps if needed.

googollee:
> but not in the case of CI: this is a special case and it can run a bit differently to be optimized for speed, simplicity of debugging, and maintainability.

There is some way to optimize CI workflow speed with `docker compose`. I can try that.

> Running tests on a different platform additionally extends the coverage.

If we want to cover more platforms, we should include it in the test script, not just set different environments in different locations. It's clearer to users and is easy to update if we change something in the test.

> but this is not a typical Docker setup

In my mind, It is a typical usage of Docker, for unit tests. Docker provides [docker-in-docker](https://www.docker.com/blog/docker-can-now-run-within-docker/) in an early stage. GitHub action works hard to run a test environment like a local host and hide the complexity. This is not a complex thing from the project side.

googollee:
About speed:

|Test|Total|UI|API SQLite|API MySQL|API PostgreSQL|
|---|---|---|---|---|---|
|[master branch](https://github.com/photoview/photoview/actions/runs/9877981467/usage)|17m37s|3m31s|4m22s|4m52s|4m52s|
|[compose test branch](https://github.com/photoview/photoview/actions/runs/10074249250/usage)|12m25s|1m5s|3m24s|3m38s|4m18s|

The new branch actually took less time comparing to the old way. One reason is that the old way always launches `mysql` and `postgres` before any `api` tests, even with `sqlite`, while the new way only launches when necessary. This shows the new way is easy to mantainace dependencies between test stages and relating services, and the test workflow could benefit from it to save running time and resources.

googollee:
Updated with caching buliding images, see [the result](https://github.com/photoview/photoview/actions/runs/10091525490/usage).

- UI: 2m57s -> 1m6s
- API:
   - SQLite: 4m20s -> 1m37s
   - MySQL: 4m16s -> 1m50s
   - PostgreSQL: 4m22s -> 1m39s

kkovaletp:
Hmm, these performance (execution time) measurement results were unexpected to me. It looks like the standard Github's way of executing each action separately in a sequence is not optimal...

However, we still have the complexity issue from reading and understanding the workflow, as well as debugging or enhancing it. This will be a major problem for a contributor who is not so familiar with Docker and needs to change something in the workflow or understand the tests' execution flow.
The workflow now is not so linear from a reading perspective, as it is not in `master`. Also, it is not a standard implementation for GitHub CI. These aspects will make additional barriers for less senior contributors. Also, they add some complexity to the workflow maintenance (because the changes need to be synchronously applied in more places now).

So, while I understand your motivation and now see benefits in this implementation, I'm still not sure about the architecture. To solve this gap, I'd like to see the @jordy2254 opinion on this PR.

googollee:
@kkovaletp I merged the workflow from https://github.com/googollee/photoview/tree/github-container. It requires to grant write permission for workflows, in `Repo Settings/Actions/General/Workflow permissions/Read and write permissions`.

kkovaletp:
> It requires granting write permission for workflows

This is something, only @viktorstrate can help with(
I think that it is much better to ask for a token with all needed permissions to be generated and added as a secret to the Actions and Dependabot sections, then referenced in the pipeline: in this way, the pipeline will run with the same permissions independently from the PR author.
I'm waiting for the same from @viktorstrate in the PR #981

kkovaletp:
As far as I see, you are blocked only with the push action. You can temporarily overcome this by pushing/pulling images from our PROD docker hub registry. Later, when @viktorstrate provides us the token, we could update the job to use Github registry for these images

scripts/install_build_dependencies.sh
```diff
@@ -12,32 +12,28 @@ else
   DEBIAN_ARCH='amd64'
 fi
 
-apt-get update
+apt update
```
kkovaletp, line 15: it is not recommended to use `apt` in scripts. this is the tool for manual execution. for scripts it is better to use `apt-get`, as it is specially designed to be more stable and less colorful

scripts/install_runtime_dependencies.sh
```diff
@@ -1,23 +1,25 @@
 #!/bin/bash
 
-BUILD_DEPENDS=(gnupg2 gpg)
+BUILD_DEPENDS=(gpg)
```
kkovaletp, line 3: Are you sure that we don't need the `gnupg2` package in any possible case? It is ok if not, but I added it from the Darktable installation guide

scripts/test/cleanup.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+docker compose -f "$(dirname $0)/../../dev-compose.yaml" down
```
kkovaletp, line 3: isn't the `down -v` better? I know that we don't use default docker volume folder location, but anyway it might do some additional cleanup...

kkovaletp (commented):
Some comments)

.github/workflows/tests.yml
```diff
@@ -2,22 +2,117 @@ name: Tests
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Cleanup
+      uses: actions/delete-package-versions@v5
+      with:
+        owner: '${{ github.repository_owner }}'
+        token: '${{ secrets.GITHUB_TOKEN }}'
+        package-type: 'container'
+        package-name: photoview/test
+        min-versions-to-keep: 5
```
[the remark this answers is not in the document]
googollee: I tried with `actions/delete-package-versions`, but the `package-version-ids` parameter requires an integer id instead of the tag. See https://github.com/actions/delete-package-versions/issues/101. I also tried to query GitHub API to get such id. The problem is [GitHub package API](https://docs.github.com/en/rest/packages?apiVersion=2022-11-28) requires an organization or a user to query packages, not a repo. To check a repo belonging to an organization or a user, the workflow is far more complex and I'm not happy about such complexity. I know the problem you said.
I don't know if it works with Docker Hub. With Docker Hub, all forked repos can't run the test workflow themselves. I hope not to do that.

scripts/install_build_dependencies.sh
googollee: As we don't depend on the output, I hope to use `apt` as it's familiar to users. Well, I don't have a strong option about it. I'll update.

scripts/install_runtime_dependencies.sh
googollee: Debian requires only with `gpg` to add keys. `gnupg2` is a dummy transitional package, pointing to `gnupg`, and `gpg` is one dependency of `gnupg`. `gpg` is the essential package. If it's not sufficient, it should be `gnupg` here.

.github/workflows/tests.yml
```diff
@@ -2,22 +2,117 @@ name: Tests
 
 on:
   push:
-    branches: [master]
+    branches: ['*']
   pull_request:
     branches: [master]
 
 jobs:
+  cleanup:
```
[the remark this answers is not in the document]
googollee: Done.

.github/workflows/tests.yml
```diff
@@ -2,22 +2,117 @@ name: Tests
...
+    name: Clean test images
+    needs:
+    - test-ui
+    - test-api
```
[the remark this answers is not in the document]
googollee: Done.

scripts/install_build_dependencies.sh
googollee: Done.

scripts/test/cleanup.sh
googollee: Done.

Dockerfile
```diff
@@ -23,74 +25,63 @@ ARG COMMIT_SHA
...
 
-### Build API ###
-FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.22-bookworm AS api
+# Download dependencies
+COPY ui/package.json ui/package-lock.json /app/ui
+RUN npm ci --omit=dev --ignore-scripts
+
+# Build UI
+COPY ui/ /app/ui
+RUN npm run build -- --base=$UI_PUBLIC_URL
```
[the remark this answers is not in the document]
googollee: I hoped that the build step should be cached with `gha`. I'm also OK to split the stages. Done.

.github/workflows/tests.yml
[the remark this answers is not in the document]
googollee: I tried some more different methods and realized that all actions are wrapped scripts running in the runner under `ubuntu-latest`. We could combine those actions and run scripts after that. Finally, I got [this workflow](I tried a few different methods). It cached all layers when building test images and code dependencies, then ran tests in containers. It doesn't push images to any repos so no need to clean up. It still runs tests with `docker compose` commands, but not through a declared container. As GitHub runner declares [providing Docker tools](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#tools) with `ubuntu-latest`, I read it as GitHub supports such feature officially.
Let me know what you think about it.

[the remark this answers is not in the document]
kkovaletp: I got [this workflow] - did you forget to put a link to some other place here?

Dockerfile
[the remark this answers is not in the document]
kkovaletp: I guess, you forgot to update the target in the build job (line 36 in the workflow)

scripts/install_build_dependencies.sh
kkovaletp: The same in the Dockerfile (line 78, 79). Sorry, missed that earlier(

.github/workflows/tests.yml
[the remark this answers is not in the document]
kkovaletp: > To check a repo belonging to an organization or a user, the workflow is far more complex

Yeap, I spent days to make it work in my forked repo, and I think that in the case of merging that PR into the parent repo, I'd have to spend even more time to make it work for it(

I need to see what is the alternative workflow, you mentioned in the 2nd comment, but if we go with this one, at least increase # of versions left to 20 so that we support up to 10 running in-parallel jobs (I guess that this would be more than enough taking into account Github CI performance and how often PRs are changed). Also, please add the step to remove all untagged versions - any job does not use them and they just use the space.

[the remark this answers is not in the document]
googollee: > > I got [this workflow]
> 
> Did you forget to put a link to some other place here?

Sorry, copy/paste the wrong content. See [here](https://github.com/googollee/photoview/actions/runs/10151697901/workflow).

Dockerfile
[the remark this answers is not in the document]
googollee: Right. Done.

scripts/install_build_dependencies.sh
googollee: Done.

.github/workflows/tests.yml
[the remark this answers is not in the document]
googollee: I prefer [that compose workflow](https://github.com/googollee/photoview/actions/runs/10157776885/workflow). It's faster than current one. If missing cache, it takes 5m for API test and 2m for UI test. If hitting cache, it takes 2m for API test and 1m for UI test. As the compose workflow doesn't share images between jobs, it doesn't need to upload/download/cleanup images, which saving a lot of time and reducing the complexity.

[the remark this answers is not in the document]
kkovaletp: yes, but it introduces more complexity in the test execution setup - the main part of the testing job)
from my point of view, we need to provide the clearest possible setup, as in an open-source project it should be clear to even a junior-level developer (with the help from Google and GPT) how the workflow works and to debug in the case of issues.
Even if the local test execution setup is not so clear for such developers, the CI should be the right place to understand the flow and debug it.

I already explained this earlier, so won't add much detail now.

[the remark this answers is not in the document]
googollee: How about expand the command in workflow? Like below:

```
    - name: Test
      run: docker compose -f dev-compose.yaml run test-api-${{ matrix.database }}

    - name: Cleanup
      run: docker compose -f dev-compose.yaml down -v
```

It shows more details in the CI workflow.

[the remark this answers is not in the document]
kkovaletp: not a big difference, as to me...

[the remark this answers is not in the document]
googollee: Outdated.

googollee:
@kkovaletp Sorry about the back and forth many times about this PR. I favor [this way](https://docs.docker.com/build/ci/github-actions/test-before-push/) to organize the test workflow. It's recommended as an [test and push example](https://github.com/marketplace/actions/build-and-push-docker-images#examples) in [actions/build-and-push-docker-images](https://github.com/marketplace/actions/build-and-push-docker-images). It only uses Docker in the workflow, and shows the relationship between steps.

Could you help to review it again? Thank you.

Dockerfile
```diff
@@ -108,5 +88,19 @@ HEALTHCHECK --interval=60s --timeout=10s \
...
+  # Required dependencies
+  && /app/scripts/install_runtime_dependencies.sh
+
 USER photoview
 ENTRYPOINT ["/app/photoview"]
+WORKDIR /home/photoview
+
+COPY api/data /app/data
+COPY --from=ui /app/ui/dist /app/ui
+COPY --from=api /app/api/photoview /app/photoview
```
kkovaletp, line 106: Why did you move lines 91-99 and 102-106 down the file? It looks messy now.
Usually, the structure is:
- copy data to the image
- run some commands
- configure the app (including the workdir if not needed to be set earlier for the RUN)
- expose port
- define healthcheck
- declare the user
- declare the entrypoint

README.md
```diff
@@ -195,9 +196,9 @@ If you don't want to depend on Docker Compose but only Docker, you can launch se
 It may take a long time to build dependencies when launching servers first time.
 
 ```sh
-$ docker build --target dev-api -t photoview-api . # Build image for development
-$ cp api/example.env api/.env
-$ docker run --rm -it -v `pwd`:/app --network host photoview-api # Monitor source code and (re)launch API server
+$ docker build --target api -t photoview/api . # Build image for development
+$ docker run --rm -it -v `pwd`:/app --network host --env-file api/example.env photoview/api \
```
kkovaletp, line 200: Why do you use `--network host` here? It is fine in the scope of the CI, as the runner is anyway isolated. But here for the developer environment, this might cause some issues and introduce unnecessary security issues. Is there a way to avoid this?

README.md
```diff
@@ -207,9 +208,9 @@ The graphql playground can now be accessed at [localhost:4001](http://localhost:
 It may take a long time to build dependencies when launching servers first time.
 
 ```sh
-$ docker build --target dev-ui -t photoview-ui .
-$ cp ./ui/example.env ./ui/.env
-$ docker run --rm -it -v `pwd`:/app --network host photoview-ui # Monitor source code and (re)launch UI server
+$ docker build --target ui -t photoview/ui . # Build image for development
+$ docker run --rm -it -v `pwd`:/app --network host --env-file ui/example.env photoview/ui \
```
kkovaletp, line 212: Same question regarding the `--network host` here. I see that it was introduced by some of your previous merges, but I probably didn't notice this earlier...

scripts/install_runtime_dependencies.sh
```diff
@@ -1,19 +1,21 @@
...
 
 # Install Darktable if building for a supported architecture
 if [ "${TARGETPLATFORM}" = "linux/amd64" ] || [ "${TARGETPLATFORM}" = "linux/arm64" ]; then
-  echo 'deb https://download.opensuse.org/repositories/graphics:/darktable/Debian_12/ /' \
-    | tee /etc/apt/sources.list.d/graphics:darktable.list
+  echo 'deb https://download.opensuse.org/repositories/graphics:/darktable/Debian_12/ /' > /etc/apt/sources.list.d/darktable.list
   curl -fsSL https://download.opensuse.org/repositories/graphics:/darktable/Debian_12/Release.key \
-    | gpg --dearmor | tee /etc/apt/trusted.gpg.d/graphics_darktable.gpg > /dev/null
+    | gpg --dearmor -o /etc/apt/trusted.gpg.d/darktable.gpg
+  gpg --show-keys --with-fingerprint --dry-run /etc/apt/trusted.gpg.d/darktable.gpg
```
kkovaletp, line 13: Is this a debug code, which you forgot to remove?)

dev-compose.yaml
```diff
@@ -2,11 +2,13 @@ name: photoview
 
 services:
   dev-ui:
-    image: photoview/photoview-ui
+    image: photoview/ui
+    profiles:
+    - dev
```
kkovaletp, line 7: why do you need the profile if there are only 2 services defined and both have this profile set? You can do the same with the command `docker compose -f ./dev-compose.yaml <action>`

dev-compose.yaml
```diff
@@ -2,11 +2,13 @@ name: photoview
 
 services:
```
kkovaletp, line 3: As I see here only the API and UI services are defined, please remind me how a developer is supposed to run MariaDB or PostgreSQL DB if needed or preferred.

kkovaletp (changes requested):
A list of comments:
Most of them are questions and styling comments, but there are a few change requests.
Also, as there are significant changes in the Dockerfile, please test and make sure that the previously merged Dev environment (both UI and API) works fine and that nothing is missing.

.dockerignore
```diff
@@ -17,6 +17,7 @@ Makefile
 photos_path
 screenshots
 storage
+photoview.db*
```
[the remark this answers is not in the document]
googollee: Updated in `.gitignore` as well.

.github/workflows/tests.yml
```diff
@@ -2,22 +2,20 @@ name: Tests
...
+    runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
       matrix:
-        database: ['mysql', 'postgres', 'sqlite']
+        database: ['sqlite', 'mysql', 'postgres']
 
     services:
-      mariadb:
+      mysql:
```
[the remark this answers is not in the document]
googollee: Done.

.github/workflows/tests.yml
```diff
@@ -31,7 +29,6 @@ jobs:
           --health-retries=5
         ports:
           - 3306:3306
-
```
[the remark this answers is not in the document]
googollee: Done.

.github/workflows/tests.yml
```diff
@@ -46,109 +43,68 @@ jobs:
...
+        push: false
+        load: true
+        target: api
+        tags: photoview/api
+        cache-from: type=gha
+        cache-to: type=gha,mode=max
+
+    - name: Test
+      run: |
+        docker run --name test --network host \
```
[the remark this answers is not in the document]
googollee: It needs to access mariadb and postgres. As both databases are launched by GitHub Action, they don't have a proper way to define the network. We could either use a deprecated Docker feature [`--link`](https://docs.docker.com/network/links/) to link the test container with database services, or use host network as here to ensure they are in the same network.

Dockerfile
googollee: The previous part won't change frequently. Moving copying binaries down lets Docker cache the previous part between buildings. See [here](https://docs.docker.com/build/cache/).

README.md
googollee: Yes, but it makes the dev environment more complex to set. It requires to [create a new Docker network](https://docs.docker.com/network/) and launch two servers in that network. With that network, the address changes from `localhost` to some random address depending on how a developer configs Docker.

As a development env is not a permanent environment, I prefer convenient over security.

scripts/install_runtime_dependencies.sh
googollee: I hope to keep it as a building log.

dev-compose.yaml
googollee: Well, we may put more services in `dev-compose.yaml` in the future, like databases or test environments. With the giving profile, there's no need to change the command then.

dev-compose.yaml
googollee: I don't have a good way to launch an external db yet. I'll improve it in the future.

.github/workflows/tests.yml
```diff
@@ -2,22 +2,20 @@ name: Tests
...
     branches: [master]
 
 jobs:
   test-api:
     name: Test API
-    runs-on: ubuntu-20.04
-
+    runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
```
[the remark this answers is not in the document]
googollee: Done.

README.md
googollee: Oh, I forgot to say, UI server needs to access API server, so they need to be in the same network, or `--link`.

README.md
googollee: It requires to be in the same network with API server.

Dockerfile
kkovaletp: There is no need to cache anything in the Expose, Healthcheck, User, and Entrypoint declarations. You can safely move them down according to the order, I mentioned. There is nothing to cache in Env declarations as well, but it is ok to have them also at the beginning of the stage, as this is the common place for them.
I see the point for reordering Copy and Run calls and it is ok to be ordered in this way, but I'd like to keep other declarative instructions, which don't impact on cache, to be located in the standard order

README.md
kkovaletp: I'll resolve this thread, as we discuss more-less the same in the other one

dev-compose.yaml
kkovaletp: This is a good point to add to the documentation: the possibility for devs to extend this compose with additional services

dev-compose.yaml
kkovaletp: let's add the 2 DB services commented out to the dev-compose file in the same way as it is done in the PROD compose example. Then if it is needed a dev can simply uncomment it and run.
Sure, there should be corresponding config lines added (also commented) to the Backend service

Dockerfile
googollee: Even those Expose steps don't have too many sizes, they still create new layers for them. From my experience, there is no common structure in a Dockerfile, just ordering with the change frequency. I hope to keep the most important part at the end of the spec.

README.md
googollee: Exposing ports is not a drawback. As it's to set local dev env, exposing ports locally is easy to debug and check if a service runs well.

I'll try to limit the server to listening `localhost` only. It avoids most of security issues.

Besides the network part, do you have any other things to document?

Dockerfile
kkovaletp: Could you please measure the time difference between the current and the latest proposed by me ordering? I expect to see no difference. If my expectation is true, I'd recommend following the common ordering. (If I remember correctly, I even read some recommendations to keep the User and Entrypoint (or CMD) to be the last 2 declarations)

README.md
kkovaletp: > Besides the network part, do you have any other things to document?

Except of already mentioned topics I don't see now something else missing. Anyway, documentation-related changes can be done at any time, as they don't have any dependencies

Dockerfile
googollee: Well, I don't insist on this. Revert as previous steps. The steps below copying won't take much time. I reverted it to the previous order.

dev-compose.yaml
googollee: I updated `dev-compose.yaml` with database, and README about how to use it.

dev-compose.yaml
googollee: Removed dev profile as using profiles with database.

README.md
googollee: Done.

googollee: BTW, `localhost` doesn't work well with `docker compose`, as it can't expose ports on localhost. It needs to listen the address assigned by the network.

.github/workflows/tests.yml
```diff
@@ -2,22 +2,20 @@ name: Tests
 
 on:
   push:
-    branches: [master]
+    branches: ['*']
```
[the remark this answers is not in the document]
googollee: Reverted.

Dockerfile
```diff
@@ -23,71 +23,65 @@ ARG COMMIT_SHA
...
-# Create a user to run Photoview server
-RUN groupadd -g 999 photoview \
+RUN chmod +x /app/scripts/install_runtime_dependencies.sh \
+  # Create a user to run Photoview server
+  && groupadd -g 999 photoview \
   && useradd -r -u 999 -g photoview -m photoview \
   # Required dependencies
-  && chmod +x /app/scripts/*.sh \
-  && /app/scripts/install_runtime_dependencies.sh
+  && /app/scripts/install_runtime_dependencies.sh \
```
kkovaletp, line 81: The trailing `\` is not needed as this is the last command in the chain now.

kkovaletp: @googollee, I found this small typo) 
Everything else looks good to me.

googollee: Good catch. My bad.

kkovaletp (approved):
LGTM

PR #1024, opened 2024-08-19
Add support for GIF images

Jorilx:
Patch to add support for GIF images.
I've tested it using the Docker development environment, please tell me if I'm missing something.
Thanks!

### DIFF
--- a/api/scanner/media_type/media_type.go
+++ b/api/scanner/media_type/media_type.go
@@ -22,6 +22,7 @@ const (
 	TypeWebp MediaType = "image/webp"
 	TypeBmp  MediaType = "image/bmp"
 	TypeHeic MediaType = "image/heic"
+	TypeGif  MediaType = "image/gif"
 
 	// Raw formats
 	TypeDNG MediaType = "image/x-adobe-dng"
@@ -78,6 +79,7 @@ var SupportedMimetypes = [...]MediaType{
 	TypeWebp,
 	TypeBmp,
 	TypeHeic,
+	TypeGif,
 }
 
 var WebMimetypes = [...]MediaType{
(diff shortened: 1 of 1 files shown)


kkovaletp:
@Jorilx, thanks for your PR. We're going to implement the media type autodetect feature instead of the current dictionary-based approach. But until then we could add this improvement if it is ok from the code perspective. @googollee, what do you think?

googollee:
The code is OK. But, does an GIF count as a photo?

@Jorilx Could you describe more about your case here? Do you have a gallary of GIFs? Still or motional?

kkovaletp:
@googollee, while we're waiting for the answer from @Jorilx, I'd like to share my opinion:
- GIF is also a picture. While it is not common to use this format for photos, technically we could support it
- As far as I know, some phones can create motion photos in GIF format
- Once we have the media type autodetect functionality, I expect GIF to be supported automatically by ImageMagick, so there will be no functionality downgrade if we merge it now
- According to the current release plan, the media autodetect feature is going to be included in the next release, so even if we merge this PR, it will be overwritten later by the new code and stay only in the list of changes in the version description, but with no real impact on users, who don't pull the `master` branch/image.

Jorilx:
> @Jorilx Could you describe more about your case here? Do you have a gallary of GIFs? Still or motional?

Yes, I have a gallery of GIFs (comic strips, actually)

googollee (commented):
LGTM

googollee (approved):
LGTM

googollee:
I hope to know the use case. @Jorilx, thanks for sharing.

Jorilx:
Thanks for the quick merge! :)

PR #1025, opened 2024-08-26
Fix release version and date in UI.

### LINKED ISSUE #452: Release Version  undefined ()
**Describe the bug**
When looking in Settings, it says the release version is undefined.

**To Reproduce**
1. Go to Settings
2. Look at "Release Version"

**Expected behavior**
Prints the release number.

**Screenshots**
image

**Your environment:**

 - Are you using Docker? Yes
 - Browser: Firefox
 - Version 2.3.5

**Additional context**

googollee:
Fix #452.

Args are copied from [this commit](https://github.com/photoview/photoview/commit/b2a79a4c9da03ea93f3fdfbae34153b4c310090d).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,6 +62,18 @@ jobs:
             type=semver,pattern={{major}}
             type=sha
 
+      - name: Prepare args
+        id: args
+        run: |
+          VERSION=edge
+          if [[ $GITHUB_REF == 'refs/tags/*' ]]; then
+            VERSION=${GITHUB_REF#refs/tags/v}
+          fi
+
+          echo VERSION=${VERSION} >> $GITHUB_OUTPUT
+          echo COMMIT_SHA=${GITHUB_SHA} >> $GITHUB_OUTPUT
+          echo BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ') >> $GITHUB_OUTPUT
+
       - name: Build and push
         uses: docker/build-push-action@v2
         with:
@@ -70,3 +82,7 @@ jobs:
           push: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
           tags: ${{ steps.docker_meta.outputs.tags }}
           labels: ${{ steps.docker_meta.outputs.labels }}
+          build-args: |
+            BUILD_DATE=${{ steps.args.outputs.BUILD_DATE }}
+            VERSION=${{ steps.args.outputs.VERSION }}
+            COMMIT_SHA=${{ steps.args.outputs.COMMIT_SHA }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,15 +62,9 @@ jobs:
             type=semver,pattern={{major}}
             type=sha
 
-      - name: Prepare args
-        id: args
+      - name: Prepare build date
+        id: date
         run: |
-          VERSION=master
-          if [[ $GITHUB_REF == 'refs/tags/*' ]]; then
-            VERSION=${GITHUB_REF#refs/tags/v}
-          fi
-
-          echo VERSION=${VERSION} >> $GITHUB_OUTPUT
           echo BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ') >> $GITHUB_OUTPUT
 
       - name: Build and push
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -62,6 +62,18 @@ jobs:
             type=semver,pattern={{major}}
             type=sha
 
+      - name: Prepare args
+        id: args
+        run: |
+          VERSION=edge
```
kkovaletp, line 68: The `edge` version is not used anymore in the product. It is fully-replaced by `master`, so please remove this line

.github/workflows/build.yml
```diff
@@ -70,3 +82,7 @@ jobs:
           push: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
           tags: ${{ steps.docker_meta.outputs.tags }}
           labels: ${{ steps.docker_meta.outputs.labels }}
+          build-args: |
+            BUILD_DATE=${{ steps.args.outputs.BUILD_DATE }}
```
kkovaletp, line 86: Please replace this line with the next one:
`BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ')`

.github/workflows/build.yml
```diff
@@ -70,3 +82,7 @@ jobs:
...
+            VERSION=${{ steps.args.outputs.VERSION }}
```
kkovaletp, line 87: Please replace this line with the next one:
`VERSION=${{ github.ref_name }}`
It should be resolved to the correct version for all the supported cases: 
- for builds in `master` branch it will return the branch name, which is the same as the build tag, and the same as the version, we currently use for 'edge' builds
- for builds in tagged commits it will return the tag name, which is what we want to be set as our version as well.
- for PRs it will return the PR ID, but the image is not pushed anywhere, so we don't care)

.github/workflows/build.yml
```diff
@@ -70,3 +82,7 @@ jobs:
...
+            COMMIT_SHA=${{ steps.args.outputs.COMMIT_SHA }}
```
kkovaletp, line 88: Please replace this line with the next one:
`COMMIT_SHA=${{ github.sha }}`
Or if you'd prefer using the Env var (as according to the GitHub docs both ways are supported, however the 1st one looks more consistent visualy):
`COMMIT_SHA=${GITHUB_SHA}`

.github/workflows/build.yml
```diff
@@ -62,6 +62,18 @@ jobs:
             type=semver,pattern={{major}}
             type=sha
 
+      - name: Prepare args
```
kkovaletp, line 65: Please test if the proposed changes to the 'Build and push' step work. If they do, we actually don't need this step at all and it can be safely removed

kkovaletp (changes requested):
Here are a few optimizations and corrections)

.github/workflows/build.yml
googollee: Then `VERSION` has a prefix `v`, which doesn't currently.

.github/workflows/build.yml
googollee: Done.

.github/workflows/build.yml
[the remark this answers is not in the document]
kkovaletp: oh, I see(
I found the `${{ github.event.head_commit.timestamp }}`, but (according to the tooltip) it returns a Unix timestamp instead of a human-readable date-time with no built-in way to convert it without an additional step(
If that is true, then we still need this additional `run` step

.github/workflows/build.yml
kkovaletp: As it is only shown to users, I don't see an issue here.
However, as we still need the `run` step to handle build time, It is also fine if you'd prefer to keep the code for tag name parsing

kkovaletp (approved):
In the PR checks tab, I see that the current implementation works fine and all 3 build args are correctly resolved into readable values and passed to the build command.
Approving the PR.

.github/workflows/build.yml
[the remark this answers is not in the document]
googollee: I tried with `github.event.repository.updated_at`, but it's not the same as the time running the job.

image

In my opnion, we should use `date` to get the right time here.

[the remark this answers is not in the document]
googollee: Done.

.github/workflows/build.yml
googollee: Done with `github.ref_name`.

.github/workflows/build.yml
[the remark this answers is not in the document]
googollee: I don't know it from the beginning. I agreed that and I updated the build script to generate `BUILD_DATE`.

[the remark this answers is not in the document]
googollee: Done.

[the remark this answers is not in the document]
kkovaletp: Great, thanks)

kkovaletp:
After the latest changes (as of now) I don't have any other comments. As the PR is already approved, you can merge it

PR #1027, opened 2024-08-28
Fix the default value of BUILD_DATE.

googollee:


Dockerfile
```diff
@@ -29,7 +29,7 @@ COPY ui/package.json ui/package-lock.json /app/ui
 RUN npm ci
 
 COPY ui/ /app/ui
-RUN if [ "$BUILD_DATE" = "" ]; then \
+RUN if [ "${BUILD_DATE}" = "undefined" ]; then \
```
kkovaletp, line 32: Isn't it better to fix it in the lines 19 and 20 by replacing the `undefined` with the `$(date -u +'%Y-%m-%dT%H:%M:%SZ')` and then remove this IF section?

googollee: `ENV` command is not running in bash, but parses by Docker. It doesn't support bash scripts.

kkovaletp: Ok, I see.

kkovaletp (approved):
Approved

PR #1028, opened 2024-08-29
Update GitHub workflows with various improvements

kkovaletp:
- Updated DB images in tests to align with the PROD setup
- Try to upload coverage only if tests were executed (passed or failed) - not skipped or canceled
- Added ESLint execution after UI tests with results exported to artifact
- Shifted scheduled run time from 00 minutes according to GitHub recommendations, as the 00 minutes of each hour is the busiest time
- Dynamically extract a list of langs from the repo to use them for CodeQL analysis instead of hardcoded ones
- Updated versions of several outdated actions
- Added the Autobuild step before CodeQL analysis for GO
- Added the Anchore dependency scan job, reporting to the Security tab. I can add steps to manage PR comments with the results, but I need a token to be provided by @viktorstrate 
- Added the Hadolint Dockerfile scan job, reporting to the Security tab. I can add steps to manage PR comments with the results, but I need a token to be provided by @viktorstrate 
- Implemented weekly rebuild of images for the latest commit in the `master` branch and the latest released tag. It will recreate images with the recent base image and 3rd-party dependencies even if there were no new pushes for a long time
- Added the Dockle container analysis job to be run on `master` and tag and validate just pushed images, reporting to the Security tab
- Added `golangci-lint` config to the `/api` folder, as a starting point and for local usage
- Added 2 weekly jobs for Dependabot: 
-- Maintain dependencies for GitHub Actions
-- Maintain dependencies for Dockerfile


kkovaletp:
@googollee, please review the PR. I've done the initial implementation.
However, `CodeQL scanned 101 out of 130 Go files in this invocation.` - could you please double-check this flow and propose how to make it scan all 130 files? Or is it ok and the rest 29 files don't need to be scanned at all?

.github/workflows/build.yml
```diff
@@ -7,6 +7,8 @@ on:
     branches: [master]
     tags:
       - v*
+  schedule:
+    - cron: '18 1 * * 4'
```
googollee, line 11: Could you add a comment like `At 01:18 on Thursday`?

.github/workflows/build.yml
```diff
@@ -15,16 +17,61 @@ env:
...
+              git fetch --all
+              TAG=$(git describe --tags --abbrev=0 || exit 0)
+              if [ -z "$TAG" ]; then
+                echo 'tags=[{"tag": "${{ github.ref_name }}", "ref": "${{ github.ref }}"}]' >> $GITHUB_OUTPUT
+              else
+                echo 'tags=[{"tag": "${{ github.ref_name }}", "ref": "${{ github.ref }}"}, {"tag": "$TAG", "ref": "$(git show-ref --tags -d | grep "/$TAG$" | cut -d ' ' -f 2)"}]' >> $GITHUB_OUTPUT
+              fi
+              ;;
+            *)
+              echo "Run for '${{ github.event_name }}' is not expected"
```
googollee, line 48: If it's not expected, maybe should fail here?

.github/workflows/codeql-analysis.yml
```diff
@@ -7,34 +7,148 @@ on:
     # The branches below must be a subset of the branches above
     branches: [master]
   schedule:
-    - cron: '0 1 * * 4'
+    - cron: '37 1 * * 4'
```
googollee, line 10: Comment like `At 01:37 on Thursday`.

googollee (commented):
I'm not familiar with `codeql`, so I can't give advice here. Let's see how it runs.

googollee:
> `CodeQL scanned 101 out of 130 Go files in this invocation.`

Where can I find this?

.github/workflows/build.yml
kkovaletp: Well, 1st of all, the whole workflow will run only if triggered by 1 of the events, defined in the `on` section in line 3, so this is just a fallback section just in case. I let it run with the default ref (also added the default ref_name instead of "" to the tag now) instead of immediately failing the run, as I don't see a negative side of it - if there is something wrong, it will fail at some point later with the real issue. If not, it will run as is with no additional modifications - why not?) I mean, we have to intentionally fail the run only if it might do something dangerous, and I don't see that here. If you do - please share.

.github/workflows/build.yml
kkovaletp: Done

.github/workflows/codeql-analysis.yml
kkovaletp: Done

.github/workflows/build.yml
googollee: OK.

googollee:
> CodeQL scanned 101 out of 130 Go files in this invocation.

After checking the CodeQL logs, it says nothing about why it doesn't scan a Go file. According to [the doc](https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/fewer-lines-scanned-than-expected#about-analysis-of-compiled-languages), it only scans files that are built during the analysis. Could you add [`debug: true`](https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/fewer-lines-scanned-than-expected#inspect-the-copy-of-the-source-files-in-the-codeql-database) to get more logs about which files are scanned?

kkovaletp:
I see a warning for the extraction of the libheif package in the Autobuild step (lines 954-993). Don't know if it is related...
Also I see more info in the Analysis step, also related to the libheif

googollee:
I don't know where `Autobuild`/`Analysis` steps are. In `Docker builds`, layers are cached.

CodeQL seems doesn't run.

kkovaletp:
These are steps of Github Actions workflow `.github/workflows/codeql-analysis.yml` lines 47-64
Results of execution are in https://github.com/photoview/photoview/actions/runs/10629633898/job/29466895944?pr=1028

googollee:
From errors, it seems the version of libheif and go heif library don't match. I guess it's because of building inside `ubuntu-latest`, the version `1.12.0` in `ubuntul-latest `is not the version in go heif `1.17.6`.

.github/workflows/codeql-analysis.yml
```diff
@@ -7,34 +7,150 @@ on:
...
-    #   matrix:
-        # Override automatic language detection by changing the below list
-        # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
-        # language: ['go', 'javascript']
-        # Learn more...
-        # https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#overriding-automatic-language-detection
-
+    needs: create-matrix
+    if: ${{ needs.create-matrix.outputs.matrix != '[]' && github.repository == 'photoview/photoview' }}
+    runs-on: ubuntu-latest
```
googollee, line 30: Could you try to run it with `container: debian:testing` to match our image's environment?

.github/workflows/codeql-analysis.yml
```diff
@@ -7,34 +7,150 @@ on:
...
+        language: ${{ fromJSON(needs.create-matrix.outputs.matrix) }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
+
+      - name: Install GO build dependencies
+        if: ${{ matrix.language == 'go' }}
+        run: |
+          sudo apt-get update
```
googollee, line 42: Better to call `./scripts/install_build_dependencies.sh`.

kkovaletp: I tried a lot of things (check the history of recent commits here), but cannot install the dependencies: https://github.com/photoview/photoview/actions/runs/10633902870/job/29480008655?pr=1028
Do you have some ideas on how to solve this issue?

.github/workflows/codeql-analysis.yml
kkovaletp: Tried with `debian:testing` and `golang:1.22-bookworm`, but I got stuck with dependencies installation(
See the other thread.

.github/workflows/codeql-analysis.yml
googollee: Checking.

googollee: I think the problem is because we build API with `golang:1.22-bookworm` and add testing repos but here it's debian-testing only. Some packages are removed in the latest `debian-testing`.

You can find how API sets env [here](https://github.com/photoview/photoview/blob/master/Dockerfile#L51). One possible way is that using `container: go:1.22-bookwrom` and copying [these code](https://github.com/photoview/photoview/blob/master/Dockerfile#L51) to build.

I'm checking if there is a way to build api from `debian-testing` and install golang from `debian-testing`.

kkovaletp: > the problem is because we build API with `golang:1.22-bookworm` and add testing repos but here it's debian-testing only.

I tried both base images with the same result(

> You can find how API sets env [here](https://github.com/photoview/photoview/blob/master/Dockerfile#L51)

Yes, I used that flow as the source for implementing this one

> One possible way is that using `container: go:1.22-bookwrom` and copying [these code](https://github.com/photoview/photoview/blob/master/Dockerfile#L51) to build.

Tried that in previous commits with the same result(

I'll try removing that huge folder (to check the assumption of no space left) and if it won't help, try to build the image target 'api' and use it for this job

kkovaletp: @googollee I thought that there is something wrong with my Bash script to install deps into the image, but I tried to build locally the PROD image from the Dockerfile and got the same issue, as failed all recent runs (since y-day's evening): https://github.com/photoview/photoview/actions/runs/10643539201/job/29507209980?pr=1028
With this, it looks like the `debian:testing` image is actually testing for Debian team and they allow it to fail from time to time. For now, all other workflows complete successfully only because of cached steps.
Please double-check and confirm from your side.
If that is true, we cannot rely on it and have to move to another base image before the release (and probably before merging something else to the `master`)

googollee: Agreed. How about alpine? I was able to build with alpine before. Alpine releases stable every half year, and we could expect upgrading the alpine every half year. 

I don't know if jellyfin- ffmpeg works well with alpine, since it doesn't provide a build with musl.

kkovaletp: Sure, that was our 2nd option, so let's try.

googollee:
Let me know if there is any problem with the `alpine` image.

kkovaletp:
@googollee, I've merged the latest `master` with the Debian stable base image and managed to fix build errors, however, I see the same issue now: even after I manually installed all the dependencies, it still analyses not all the GO files:

https://github.com/photoview/photoview/actions/runs/10751803784/job/29819431878?pr=1028#step:7:961: 
`CodeQL scanned 101 out of 129 Go files in this invocation.`

Analyzing the workflow execution log I see the next warnings, which might be related to this issue:
- https://github.com/photoview/photoview/actions/runs/10751803784/job/29819431878?pr=1028#step:6:260
- https://github.com/photoview/photoview/actions/runs/10751803784/job/29819431878?pr=1028#step:6:773
- https://github.com/photoview/photoview/actions/runs/10751803784/job/29819431878?pr=1028#step:7:964

Please take a look at the execution results and the current workflow. Do you see some possible reasons?

.github/workflows/codeql-analysis.yml
```diff
@@ -7,34 +7,182 @@ on:
...
+          export GOPATH="/go"
+          export PATH="${GOPATH}/bin:${PATH}"
+          export CGO_ENABLED=1
+          # Comment out the last line with 'apt-get install -y reflex sqlite3'
+          sed -i '$ s/^/#/' ../scripts/install_build_dependencies.sh
+          chmod +x ../scripts/*.sh
+          ../scripts/install_build_dependencies.sh
+          apt-get install -y file
+          echo "<<< Dependencies installed >>>"
+          . ../scripts/set_compiler_env.sh
```
googollee, line 70: I think you should think about Code Analysis with amd64 only. Put this before `install_build_dependencies.sh` and only set `TARGETPLATFORM=linux/amd64`in the beginning. As it doesn't use `docker buildx` and `qemu`, it won't run with other arch.

googollee:
The error you listed is because we are using 1.23, and the analysis is using 1.22. The new standard library can't be compiled with 1.22.

```
  2024/09/07 13:33:22   /usr/local/go/src/slices/iter.go:50:17: cannot range over seq (variable of type iter.Seq[E])
  2024/09/07 13:33:22   -: This application uses version go1.22 of the source-processing packages but runs version go1.23 of 'go list'. It may fail to process source files that rely on newer language features. If so, rebuild the application using a newer version of Go.
```

And, there is no log about why it doesn't scan all go codes. So I can't say. What is this code analysis for?

kkovaletp:
> The error you listed is because we are using 1.23, and the analysis is using 1.22

I thought, the reason was the opposite, as I ran the workflow inside the GO 1.23 container, but it finds 1.22 declared in the `go.mod` - I'll try to modify it in my branch and check the result.
Hm, I've updated the feature level and I think there are no other ways, where the 1.22 version can be defined in the code or environment, but I still see the same warning about the 1.22 version in the new execution log: https://github.com/photoview/photoview/actions/runs/10753914524/job/29823905186#step:6:262
In fact, after updating the feature level in the `go.mod` I see this warning A LOT more times in the log, so it definitely needs to be aligned somehow with the code, not just 1 line change.
And still `scanned 101 out of 129 Go files in this invocation`.
So, I guess that I'll just accept the fact that a small part of the files are not scanned, and revert all this complex setup to the basic implementation, which runs faster with the same final result.

> What is this code analysis for?

It reports found issues to the https://github.com/photoview/photoview/security/code-scanning

.github/workflows/codeql-analysis.yml
kkovaletp: I set the platform in line 59.
I switched the order, but the result is the same - see the details in my message down the page

googollee:
> I thought, the reason was the opposite, as I ran the workflow inside the GO 1.23 container, but it finds 1.22 declared in the go.mod.

If so, I suggest not to use it. It doesn't follow [rules of Go](https://go.dev/ref/mod#go-mod-file-go). Since we didn't use any 1.23 semantics in `photoview`, it must not require to bump the version.

Also, after checking [results](https://github.com/photoview/photoview/security/code-scanning), they are blind matchs, without checking for correct semantics. [This one](https://github.com/photoview/photoview/security/code-scanning/107) validates the `rootPath` and should not limit any paths. [This one](https://github.com/photoview/photoview/security/code-scanning/104) is using the validating function to check `servePath`.

If you want to check Go code, I suggest using [`golangci-lint`](https://github.com/golangci/golangci-lint).

.github/workflows/codeql-analysis.yml
```diff
@@ -7,34 +7,182 @@ on:
...
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
+
+      - name: Install GO build dependencies
+        if: ${{ matrix.language == 'go' }}
+        working-directory: ./api
+        shell: bash
+        run: |
```
googollee, line 47: No need to set `working-directory` before. With that, `run` could be:

```
export TARGETPLATFORM="linux/amd64"

git config --global --add safe.directory `pwd`
# Comment out the last line with 'apt-get install -y reflex sqlite3'
sed -i '$ s/^/#/' ../scripts/install_build_dependencies.sh

source ./scripts/set_compiler_env.sh
./scripts/install_build_dependencies.sh
apt-get install -y file
echo "<<< Dependencies installed >>>"

go env
go mod download
# Patch go-face
sed -i 's/-march=native//g' ${GOPATH}/pkg/mod/github.com/!kagami/go-face*/face.go
# Build dependencies that use CGO
go install github.com/mattn/go-sqlite3 github.com/Kagami/go-face
```

kkovaletp: I thought that these GO commands should be executed from the repo's ./api folder, as it was originally implemented in the testing workflow (before the switch to the container). Are you sure that they will be executed correctly from the repo's root?

kkovaletp:
> I suggest using [`golangci-lint`](https://github.com/golangci/golangci-lint). It's based on AST, but not text matching

I'll investigate this tool and the way to integrate it here instead of GitHub's default code analysis tool

.github/workflows/codeql-analysis.yml
googollee: Those scripts are installing dependencies. We could change to `api` after setting env.

googollee:
Go team provides a tool [`govulncheck`](https://go.dev/blog/govulncheck) to check dependencies based on a database. It could also help.

kkovaletp:
I'll try it, but I think that Anchore does the same job and should produce better results, as they check many vuln-DBs. Also, it supports all our langs, so it is easier to maintain.

Also, there is a strange (as to me) solution in the [govulncheck](https://go.dev/blog/govulncheck) to filter out all vulns, which the tool thinks are not used by the code: 
- the tool might miss-detect some types of calls (I'm not sure if that is possible in GO, but in other langs, there are many ways to inject a lib or part of a lib into the code on condition or in real-time, which might not be detected correctly by code analyzers).
- even if a vulnerable code is not called by the product, it is still there and can be used by a malicious actor.

googollee:
> I think that Anchore does the same job and should produce better results, as they check many vuln-DBs.

I agree on that. But since they are a common tool, it can't support a special language very well, like problems we met. We could bump the go version in `go.mod` to fix one of them, but it's hard to tell why it doesn't scan all files since it doesn't provide any detail.

> there are many ways to inject a lib or part of a lib into the code on condition or in real-time, which might not be detected correctly by code analyzers

I don't think there is a way to inject code during compiling with Go. Go doesn't have macro and can't run any other external commands during `go build`.

> even if a vulnerable code is not called by the product, it is still there and can be used by a malicious actor.

I believe [govulncheck](https://go.dev/blog/govulncheck) checks all source code, but not only calling part.

kkovaletp:
@googollee, after some efforts of integrating the `golangci-lint` into our workflows, I decided to skip it for now and finish this PR 'as is': the `golangci-lint` requires too much effort for its correct configuration and optimal integration into our workflows. I've created #1059 to track this activity.

Please review the PR once more time

.github/workflows/build.yml
```diff
@@ -16,13 +19,61 @@ env:
...
+            push)
+              echo 'tags=[{"tag": "${{ github.ref_name }}", "ref": "${{ github.ref }}"}]' >> $GITHUB_OUTPUT
+              ;;
+            schedule)
+              git fetch --all
+              TAG=$(git describe --tags --abbrev=0 || exit 0)
+              if [ -z "$TAG" ]; then
+                echo 'tags=[{"tag": "${{ github.ref_name }}", "ref": "${{ github.ref }}"}]' >> $GITHUB_OUTPUT
+              else
+                echo 'tags=[{"tag": "${{ github.ref_name }}", "ref": "${{ github.ref }}"}, {"tag": "$TAG", "ref": "$(git show-ref --tags -d | grep "/$TAG$" | cut -d ' ' -f 2)"}]' >> $GITHUB_OUTPUT
```
googollee, line 46: If the workflow is triggered by the scheduler, and there is a tag at head, it will build image twice with the same code. Is that what you want? I think we could build it once and add mutiple tags to the image.

googollee (approved):

kkovaletp:
@googollee, I've just added a small change after your approval in the https://github.com/photoview/photoview/pull/1028/commits/7e58f19e774d2c31631c9536f3cc7b85d4163a72

Added artifact publishing steps for Hadolint and Anchore results, so that they are easily accessible/downloadable.
If that is fine, I'll merge the PR

googollee (approved):

googollee:
LGTM, approved.

Issue #1009, opened 2024-08-03
User id 999 is `systemd-journal` on Debian by default.


googollee:
The docker image of photoview creates [user `photoview` with id 999](https://github.com/photoview/photoview/blob/master/Dockerfile#L84). On Debian system, user ids `100-999` dynamically allocate to system users and groups. See [UID and GID glasses](https://www.debian.org/doc/debian-policy/ch-opersys.html#uid-and-gid-classes). We should use not use `999` here.

googollee:
Creating users in the image that does not exist in the host will cause many permission problems. The files created by the image have an invalid user/group as the owner, and the binary in the image can't read/write mounted files (except granting o+wr permission, which is another potential security issue) because of no such user on the host.

I propose to use `nobody:nogroup` in `Dockerfile` and guiding users to set a proper user/group when launching the image. `nobody:nogroup` is a de facto user without permission. If user doesn't provide user, it won't cause any security issues.

kkovaletp:
Good catch! Thanks for reporting.

The fix is technically very straightforward: just replace the UID and GID from the currently used `999` to some other ID > than 1024.
Your proposal to fix might be strategically better, but we need to investigate and research it deeper to make sure that it will work for us. 
However, this is a breaking change, as after the latest release some users already configured a user (and group) on their host with the same IDs to better manage the permissions inside the volumes. So, changing this now will break their setup and force them to urgently reconfigure the user on the host and fix all previously granted permissions of the content in the volumes.

googollee:
I hope not to create any user/group, but using `nobody:nogroup` by default, or `www-data:www-data`.

Docker doesn't isolate user id and group id between containers and host, if not [enabling user ns remapping](https://docs.docker.com/engine/security/userns-remap/). The user in the container shares the same permissions with the same id user on the host. Imaging two cases:

- User id `999` in the container and the host doesn't have any users with id `999`. 
  - The image creates caches on the host with 999 as the owner. When list those files, the owner is showed as id `999`.
  - The image can't read mounted files if they don't grant `o+r`.
    - The user can't change mounted files to the user/group with id 999, as there is no such user/group on the host.
 - User id `999` in the container and the host has another high-priority user with the same id `999`.
   - Suddenly the container runs with a high-priority user, as with the same id, the user in the container is the user on the host.
   - If photoview has any vulnerabilities, it'll be a chance to use it with more permissions.

[...]

PR #1033, opened 2024-09-01
Use alpine as base images.

### LINKED ISSUE #1031: `debian:testing` base image is too unstable for us, as it gets corrupted from time to time
**Describe the bug**
Recently we've migrated to the `debian:testing` base image, however, from yesterday evening it has been corrupted (this time the package repo for the `testing` distro's branch contains corrupted packages, which cannot be installed). This means that this branch is used for testing and we cannot rely on it.

**To Reproduce**
Try to build the PROD image locally without the cache while the Debian repo isn't fixed. Or check the log of the CI job: https://github.com/photoview/photoview/actions/runs/10643539201/job/29507209980?pr=1028

**Expected behavior**
We have to migrate to another base image, stable enough to rely on it. Maybe that will be Alpine, but we need to try it 1st.

 - Are you using Docker? yes 
 - Version `master`

**Additional context**
https://github.com/photoview/photoview/pull/1028

googollee:
Fix #1031 and #1009.


googollee:
It could provide `amd64,arm64,armv6,armv7` images, see https://github.com/googollee/photoview/pkgs/container/photoview/266580829?tag=build-api.

Dockerfile
```diff
@@ -36,54 +33,43 @@ RUN if [ "${BUILD_DATE}" = "undefined" ]; then \
...
 COPY scripts/install_runtime_dependencies.sh /app/scripts/
 RUN chmod +x /app/scripts/*.sh \
   && /app/scripts/install_build_dependencies.sh \
+  && mv $(go env GOPATH)/bin/reflex /usr/bin/ \
   && /app/scripts/install_runtime_dependencies.sh
 
 COPY api/go.mod api/go.sum /app/api/
-RUN source /app/scripts/set_compiler_env.sh \
-  && go env \
+RUN go env \
```
kkovaletp, line 54: Why don't we need to source the set env script?

Dockerfile
```diff
@@ -36,54 +33,43 @@ RUN if [ "${BUILD_DATE}" = "undefined" ]; then \
...
-# See for details: https://github.com/hadolint/hadolint/wiki/DL4006
-SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
-
 COPY scripts/install_runtime_dependencies.sh /app/scripts/
 RUN chmod +x /app/scripts/install_runtime_dependencies.sh \
   # Create a user to run Photoview server
-  && groupadd -g 999 photoview \
-  && useradd -r -u 999 -g photoview -m photoview \
+  && addgroup -g 9999 photoview \
+  && adduser -u 9999 -G photoview -D photoview \
```
kkovaletp, line 72: I'd agree that we have to fix this ID by setting it from the unprivileged range, but this will be a breaking change for users, who already created a user and group on their host to match the IDs in the container. Is it unavoidable (like, the 999 is busy)?

README.md
```diff
@@ -254,13 +254,7 @@ The site can now be accessed at [localhost:1234](http://localhost:1234).
...
     - `node` = 18
 
-In Debian/Ubuntu, install dependencies:
-
-```sh
-$ sudo apt update # Update the package list
-$ sudo apt install golang g++ libc-dev libheif-dev libdlib-dev libjpeg-dev libblas-dev libatlas-base-dev liblapack-dev # For API requirement
-$ sudo apt install reflex sqlite3 # For API optional tools
-```
+In Linux, please use [Docker Compose or Docker](https://github.com/googollee/photoview?tab=readme-ov-file#set-up-docker-development-environment).
```
kkovaletp, line 257: There are (were) contributors, who don't want/like to use Docker-based dev env. I guess that we can keep the old instructions for them in the Readme.
Same for lines 300-301

kkovaletp (commented):
Some comments and questions:

.github/workflows/build.yml
```diff
@@ -36,11 +35,14 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Docker Login
-        if: success() && github.event_name != 'pull_request' && github.repository == 'photoview/photoview'
+        if: ${{ success() && github.event_name != env.FROM_EVENT && github.repository == env.FROM_REPO }}
```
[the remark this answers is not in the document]
googollee: Not for testing. It changes variables for both strings and it's easy to change for testing. It's better to keep them.

Dockerfile
```diff
@@ -36,54 +33,43 @@ RUN if [ "${BUILD_DATE}" = "undefined" ]; then \
...
 ENV CGO_ENABLED=1
 
 # Download dependencies
-COPY scripts/apt/debian-testing.sources /etc/apt/sources.list.d/
-COPY scripts/set_compiler_env.sh /app/scripts/
 COPY scripts/install_build_dependencies.sh /app/scripts/
 COPY scripts/install_runtime_dependencies.sh /app/scripts/
 RUN chmod +x /app/scripts/*.sh \
   && /app/scripts/install_build_dependencies.sh \
+  && mv $(go env GOPATH)/bin/reflex /usr/bin/ \
```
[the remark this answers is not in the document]
googollee: For `reflex` here, it's safe as it's a pure go binary.

Dockerfile
googollee: The `set_env` script is used for cross-compiling. As we are building `arm` images in QEMU env, it does not require any cross-compiling. It's better to be removed to keep the workflow simple. Also, it's more stable to rely on the predefined env variables than to set them manually.

README.md
googollee: It's conflicted with [your reply](https://github.com/photoview/photoview/issues/1031#issuecomment-2323202255). As we are not building on Debian, these packages may not be installed in the future, and we can't verify it easily.

Since we list dependencies above, developers know which packages to be installed in their env.

Dockerfile
googollee: OK. Reverted.

.github/workflows/build.yml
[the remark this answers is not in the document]
googollee: Updated. Removed `success()` and using `IS_PUSHING_IMAGE` to control both login and push.

Dockerfile
googollee: Unfortunately, `alpine` reserves `999`.

```
 > [linux/amd64->arm/v6 release 3/7] RUN chmod +x /app/scripts/install_runtime_dependencies.sh   && addgroup -g 999 photoview   && adduser -u 999 -G photoview -D photoview   && /app/scripts/install_runtime_dependencies.sh:
0.130 addgroup: gid '999' in use
------
------
 > [linux/amd64->arm/v7 release 3/7] RUN chmod +x /app/scripts/install_runtime_dependencies.sh   && addgroup -g 999 photoview   && adduser -u 999 -G photoview -D photoview   && /app/scripts/install_runtime_dependencies.sh:
0.169 addgroup: gid '999' in use
------
------
 > [linux/amd64->arm64 release 3/7] RUN chmod +x /app/scripts/install_runtime_dependencies.sh   && addgroup -g 999 photoview   && adduser -u 999 -G photoview -D photoview   && /app/scripts/install_runtime_dependencies.sh:
0.129 addgroup: gid '999' in use
```

googollee:
@kkovaletp `Imagemagick` in `alpine` doesn't support any camera RAW format by default.

```
/app # magick identify -list format  | grep RAF
/app # magick identify -list format  | grep Fuji
/app # magick identify -list format  | grep Canon
/app # magick identify -list format  | grep NEF
/app # magick identify -list format  | grep CR2
/app # magick identify -list format  | grep Raw
    BAYER* BAYER     rw+   Raw mosaiced samples
   BAYERA* BAYER     rw+   Raw mosaiced and alpha samples
      BGR* BGR       rw+   Raw blue, green, and red samples
     BGRA* BGR       rw+   Raw blue, green, red, and alpha samples
     BGRO* BGR       rw+   Raw blue, green, red, and opacity samples
     CMYK* CMYK      rw+   Raw cyan, magenta, yellow, and black samples
    CMYKA* CMYK      rw+   Raw cyan, magenta, yellow, black, and alpha samples
     GRAY* GRAY      rw+   Raw gray samples
    GRAYA* GRAY      rw+   Raw gray and alpha samples
      M4V  VIDEO     rw+   Raw VIDEO-4 Video
     MONO* MONO      rw-   Raw bi-level bitmap
      RGB* RGB       rw+   Raw red, green, and blue samples
   RGB565* RGB       r--   Raw red, green, blue samples in 565 format
     RGBA* RGB       rw+   Raw red, green, blue, and alpha samples
     RGBO* RGB       rw+   Raw red, green, blue, and opacity samples
    YCBCR* YCBCR     rw+   Raw Y, Cb, and Cr samples
   YCBCRA* YCBCR     rw+   Raw Y, Cb, Cr, and alpha samples
```

I'm working on finding a way to solve it now.

googollee:
Solved [the raw problem](https://github.com/photoview/photoview/pull/1033#issuecomment-2323408915). Also verified that when a cached jpg is missing, rerunning the scan will try to generate a new one.

.github/workflows/build.yml
```diff
@@ -36,7 +34,7 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Docker Login
-        if: success() && github.event_name != 'pull_request' && github.repository == 'photoview/photoview'
+        if: ${{ env.IS_PUSHING_IMAGES == 'true' }}
```
kkovaletp, line 37: the `== 'true'` part is useless, as the variable already contains `true` or `false`

Dockerfile
kkovaletp: Ok, then this is the time to introduce this breaking change.
This means that the next version would be again major

README.md
kkovaletp: Ok, I see your point.
Well, for the Docker-based deployments, we provide a fully integrated package (image + PROD compose + DEV compose + optional Makefile). However, on the other side, we still support the direct host deployment scenario, and for that scenario, we have to provide the list of dependencies and deployment steps. Currently, we have the list of supported distros and the migration of our image to Alpine should increase this list, not decrease.
This doesn't mean that we have to provide installation scripts and other easy-install stuff (unless we want to), but we need to provide the list of prerequisites and requirements to make Photoview work on currently supported systems.

In short: in addition to recommending users to use Docker and warn them about the risk of these instructions becoming outdated, please keep the direct installation instructions as well.

> As we are not building on Debian, these packages may not be installed in the future, and we can't verify it easily.

This risk is acceptable for the dev env. You can mention in the warning above, that devs have to be ready to resolve some dependency-related issues and are expected to contribute updates to the readme with the tested corrections.
However, after some time we might see the same type of issues with our PROD host deployment dependency inconsistencies. Since both of us don't use host deployments, the best we can do is to react to users' reports and try to fix those issues. I don't think that this risk should stop us from improving the product.

Dockerfile
```diff
@@ -1,8 +1,8 @@
 ### Build UI ###
-FROM --platform=${BUILDPLATFORM:-linux/amd64} node:18 AS ui
+FROM --platform=${BUILDPLATFORM:-linux/amd64} node:18-alpine AS ui
 
 # See for details: https://github.com/hadolint/hadolint/wiki/DL4006
-SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
+SHELL ["/bin/sh", "-euo", "pipefail", "-c"]
```
kkovaletp, line 5: Does it work for SH shell in the same way? I thought this is Bash-specific modifier)

scripts/install_runtime_dependencies.sh
```diff
@@ -1,10 +1,11 @@
-#!/bin/bash
+#!/bin/sh
 set -euo pipefail
 
-apt-get update
-apt-get install -y curl libdlib19.2 ffmpeg exiftool libheif1 imagemagick
+apk update
+
+apk add curl libheif lapack cblas ffmpeg exiftool imagemagick imagemagick-heic imagemagick-jpeg imagemagick-raw imagemagick-tiff
```
kkovaletp, line 6: As ImageMagick in Alpine repo separates all its libs and plugins, please review the full list of available add-ons and install all image formats and additional libs, which might be useful for us. So far, I found:
- imagemagick-webp
- imagemagick-svg
- imagemagick-jxl
- imagemagick-libs - not sure what are these libs, but maybe they are useful (converters, codecs,...) - please investigate

kkovaletp (commented):
Few more comments

scripts/install_runtime_dependencies.sh
kkovaletp: Also, a similar case might be valid for `ffmpeg`: https://pkgs.alpinelinux.org/packages?name=ffmpeg*&branch=edge&repo=&arch=x86_64&maintainer=
However, as we plan to switch to the Jellyfin's ffmpeg, it might be not so critical, but still there might be the case, Jellyfin's ffmpeg is able to use and rely on those add-ons...

.github/workflows/build.yml
googollee: Nope. The ENV value is a `true` string, but not a boolean. `if` requires a boolean. So here we need the comparison.

I tried without the comparison before, and it fails.

scripts/install_runtime_dependencies.sh
googollee: Done.

I don't think we are able to use `jellyfin-ffmpeg` with `alpine`, because `jellyfin-ffmpeg` binaries are built with `glibc`. `ffmpeg` in `alpine` is compiled with `qsv` and `vaapi` codecs, which are hardware acceleration by Intel and (maybe) AMD CPU. We need to find a way to compile a `ffmpeg` with `cuda` ourselves if we want to support Nvidia.

kkovaletp: You've mentioned that you can try to build Jellyfin ffmpeg for Alpine from sources. Isn't that an option anymore?
Did you try to search whether someone already ported Jellyfin ffmpeg to Alpine (or tried at least)?

Yes, with QSV (Intel) and VAAPI (AMD), it is a good start, but it would be much better to have more HW and interfaces/protocols supported

googollee: I don't have any options yet. Will research solutions first after this PR.

README.md
googollee: Another reason of removing those Debian commands is that they are already broken in Debian stable and Ubuntu stable, because the library of go-heif `v1.17` won't be compiled with the `libheif` from the distribution. I can keep those instructions in README, but I guess it just confuses people.

googollee: Added those commands back.

.github/workflows/build.yml
```diff
@@ -1,16 +1,18 @@
 name: Docker builds
 
 on:
```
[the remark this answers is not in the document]
googollee: Reverted.

scripts/install_runtime_dependencies.sh
googollee: As a quick search, without [Nvidia supporting alpine](https://github.com/NVIDIA/nvidia-container-toolkit/issues/270) , we can't support Nvidia HW.

README.md
kkovaletp: > I can keep those instructions in README, but I guess it just confuses people.

Ok, let's see... At least, this will be the last working configuration, so devs could take it as a starting point and fix inconsistencies - that is easier than inventing it from scratch, I guess

scripts/install_runtime_dependencies.sh
kkovaletp: Sad news(
However, even without NVidia support, Jellyfin still provides a lot more HW support protocols: https://jellyfin.org/docs/general/administration/hardware-acceleration/

I mean, let's merge now with the default ffmpeg + plugins and then it looks to make sense to try replacing it with the Jellyfin's version

kkovaletp (approved):
Except for the open thread about the breaking change of UID and GID, I don't see any other topics to discuss before the merge, so please let me know what you think on that topic and make sure that everything works as expected (CI builds the image correctly with all the variables applied and product working; the local dev-env is built and running UI and API dev-servers fine with the ability to run tests)

Approving the PR, so once everything is done and checked, it can be merged

Dockerfile
googollee: Done.

PR #1035, opened 2024-09-02
Update to magick 7 in alpine.

googollee:
As we switched to `alpine`, we could update `magick` commands in the new way.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: [master]
+    branches: ['*']
   pull_request:
     branches: [master]
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: ['*']
+    branches: [master]
   pull_request:
     branches: [master]
 
--- a/api/scanner/media_encoding/executable_worker/executable_worker.go
+++ b/api/scanner/media_encoding/executable_worker/executable_worker.go
@@ -37,17 +37,17 @@ func newMagickWorker() *MagickWorker {
 		return nil
 	}
 
-	path, err := exec.LookPath("convert")
+	path, err := exec.LookPath("magick")
 	if err != nil {
-		log.Println("Executable worker not found: ImageMagick convert")
+		log.Println("Executable worker not found: magick")
 	} else {
-		version, err := exec.Command(path, "--version").Output()
+		version, err := exec.Command(path, "-version").Output()
 		if err != nil {
-			log.Printf("Error getting version of ImageMagick convert: %s\n", err)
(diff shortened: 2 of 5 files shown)


api/scanner/media_encoding/executable_worker/executable_worker.go
```diff
@@ -37,17 +37,17 @@ func newMagickWorker() *MagickWorker {
 		return nil
 	}
 
-	path, err := exec.LookPath("convert")
+	path, err := exec.LookPath("magick")
 	if err != nil {
-		log.Println("Executable worker not found: ImageMagick convert")
+		log.Println("Executable worker not found: magick")
```
kkovaletp, line 42: While the `magick` executable name better correlates with the 'ImageMagick', I'm not sure that it is intuitive enough for users to be mentioned in log messages as is...
This comment is more like a discussion starter than a change request...

.github/workflows/tests.yml
```diff
@@ -2,7 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: [master]
+    branches: ['*']
```
kkovaletp, line 5: A standard reminder for testing changes)

kkovaletp (commented):
Just 1 topic to discuss.
Apart from that, I don't have any other questions. This is a small PR)
Why wasn't the `magick` executable used in the Debian image?

kkovaletp:
Ah, 1 more question:
The PR is named as 'Update to magic 7', but I don't see the version change in the changed files... How does it work then?)

googollee:
Debian testing only provides ImageMagick 6. I believe we discussed this before. ImageMagick 6 -> 7 changes the way to call the command. With 7, all commands of ImageMagick are under `magick` command, e.g. `convert` -> `magick convert`.

Alpine provides ImageMagick 7, and I want to follow the new way here. I didn't do it in the previous PR because 7 also provides commands for the legacy way and I didn't want to change too many things in one PR.

api/scanner/media_encoding/executable_worker/executable_worker.go
googollee: Here `magick` is the main command of ImageMagick. It follows the way of `ffmpeg` below.

With ImageMagick 6, a command like `convert` isn't able to refer ImageMagick easily, so I use `ImageMagick convert`. Now `magick` is really identifiable and no need `ImageMagick` as a prefix.

kkovaletp:
Ok, I see: the ImageMagick's executable was updated together with the base image. Here you update only the way, it is called - from a legacy one to a modern one.

.github/workflows/tests.yml
kkovaletp: Revert the testing code and I'll approve the PR

googollee: Finished testing and reverted workflows.

kkovaletp (approved):
LGTM

PR #1040, opened 2024-09-04
Back to `bookworm`, rollback to 6c5c304

googollee:
It's not going well when moving to `debian:testing` then `alpine`.

For `debian:testing`, as upstream changes frequently, it may break our build.

For `alpine`, most of things goes well, but it doesn't support by Nvidia container runtime. We can't continue HW features with `alpine`.

I'd like to rollback these changes to previous `bookworm` image and try another way to catch up with upstream. Maybe build the latest `ImageMagick` ourselves.

Another problem is, current `arm` images are not compiled well. I'd like to see if `arm` works previously.

This PR reverts last 3 commits and cherrypicks useful changes from them. It doesn't change other codes. See [the comparison](https://github.com/photoview/photoview/compare/6c5c304...googollee:photoview:back-bookworm) between `back-bookworm` and commit 6c5c304.


kkovaletp:
> it doesn't support by Nvidia container runtime

We knew that before the switch and decided to go with it even without the support for Nvidia. Is something changed here?

> current `arm` images are not compiled well. I'd like to see if `arm` works previously.

What exactly doesn't work? 
You can pull some older tags from the Docker Hub to check previous builds - no need to roll back.

googollee:
> We knew that before the switch and decided to go with it even without the support for Nvidia. Is something changed here?

Now `h264_qsv` also doesn't work. I didn't verify it before, seeing it provided that codec.

> What exactly doesn't work?

Cross-compile doesn't work well and it generates `amd64` binaries for `armxx`. I can fix it but as `qsv` doesn't work too, I hope to check another way.

kkovaletp:
> Cross-compile doesn't work well and it generates `amd64` binaries for `armxx`

Can you check that by pulling an old image or building one here in the PR with no need to merge it?

Sad to read that we again did something, which doesn't meet all our needs, and have to reimplement it...
I don't, actually, want to get back to the Debian image, as we know its cons...
Let's review our options:
- originally we talked about Fedora, which doesn't provide images for 2 of our arch's
- [Oracle Linux](https://hub.docker.com/_/oraclelinux): Oracle Linux is another RHEL-compatible distribution that offers multiple architectures and frequent updates. It is known for its stability and enterprise-grade support.
- [Rocky Linux](https://hub.docker.com/_/rockylinux): Rocky Linux is a community-driven project that aims to be a downstream, binary-compatible release using the Red Hat Enterprise Linux (RHEL) source code. It supports multiple architectures and is designed to be stable and reliable
- [openSUSE Leap](https://hub.docker.com/r/opensuse/leap): openSUSE Leap is a stable distribution with regular updates and support for multiple architectures. It is known for its robustness and extensive package repository
- Anything else?
Unfortunately, all these registries don't provide (as far as I see) images for `arm/v6` and `arm/v7`, so if we'd choose 1 of them, we'll have to drop the support of those platforms

googollee:
I know it's back and forth a lot.

I don't know if other distributions provide wide platform support. Another problem is that `jellyfin-ffmpeg`.

I don't have too many experiences with Red Hat and SUSE. Do you know those distributions' release cycles?

I hope to go back to `debian:bookworm` because it worked previously. Since `ImageMagick` and `Ffmpeg` are important dependencies, it's worth building them ourselves to get more control. To control image formats well, we at least need to catch the latest versions of `libraw`/`libheif`. Other dependencies like `jpeg`/`png`/`tiff` are pretty stable and no need to catch the latest versions.

kkovaletp:
I use Fedora KDE Workstation 40 as 1 of my systems now. The other one is Kubuntu 24.04 LTS (and I regret that I chose it at the system installation time, while now after years of using it, I don't want to go through weeks of reinstalling and reconfiguring all the system and software there)
At work, I used Oracle Linux a lot, but it was ~5 years ago in a different project. Now I use Rocky Linux (as a successor of CentOS), but on a physical system and in a very hardened setup, so I don't do the same tasks, I'd do in the scope of our project.
Finally, I have pretty limited experience with SUSE (both proprietary and OpenSUSE) on VMs: it is a mature (maybe 1 of the most mature distro in the Linux family), but it is so different from all other Linux distros, so I had to google and learn almost everything, I had to do there...

As you see a major risk in switching to another distro out of the Debian family, I'd agree with you, as we already spent a lot of time trying to make a new image work, and I don't want us to throw yet another month into a trash for trying the same on a new base image, finding some new showstoppers there and getting back to Debian...
However, as you mentioned important dependencies, how would we solve the `libheif` version issue (it was 1 of your 1st tasks to update dependencies, and you found the unresolvable dependency to the base image, which doesn't allow to update the `libheif` version in the `bookworm` base image)?

I'd like at some point to migrate all our media processing to ImageMagick so that we don't have any code for that in our product and instead, rely completely on well-developed and maintained media processing tools - those projects have more resources and competence to do this job much better. We need to focus more on managing and representing the results of those tools in the best way.

googollee:
As I said, I hope to keep in Debian stable and build some packages (libraw, libheif, ImageMagick) based on Debian stable. I don't think other distributions have a good relationship with other images. `golang` and `node` only provide `debian` and `alpine` images. As `alpine` can't support HW acceleration well, we should keep with Debian stable.

kkovaletp:
I started the code review, but see a lot of unwanted changes. So please do the 1st round of the code review by yourself and fix those issues, then let me know when PR is ready.

googollee:
I didn't change any code. This PR revert the last 3 commits and it's the same as 6c5c304.

googollee:
See the comparison between `back-bookworm` and commit 6c5c304 [here](https://github.com/photoview/photoview/compare/6c5c304...googollee:photoview:back-bookworm).

kkovaletp:
> I didn't change any code. This PR revert the last 3 commits

Ok, so once it is merged, those 3 "reverted" commits will still be in the history before this one in the `master` branch - the branch HEAD won't be reset to a previous commit, right?
Then, why would we want to revert everything? Isn't it better to use this PR to restore only the things, which have to be restored, and keep those, we want to stay, as they are useful even for Debian image?

googollee:
I want to (squash) commit this as "Rollback to 6c5c304" and with one more PR to pick the updates. It's clear in history where it goes back.

kkovaletp:
Why to create 2 commits in `master` instead of 1? As far as I can see, the "revert" commit doesn't add a value.
In general, I wouldn't say I like the strategy of many small PRs to implement a feature, as in this case, we cannot focus enough on the business value, which we add to the product. Instead, I like (and try to follow on work) the strategy of 1 feature by 1 PR. If a feature is too big to be implemented by 1 developer or during a reasonable time (or the PR will be too big and complex for review), we create a feature branch and merge smaller PRs into that branch until we're happy with the result to merge it as a feature into `master`.
So, you already have a branch, reverted to the desired commit in `master`. Also, you opened the PR, so you see the difference between the current HEAD and the reverted state. Let's use this branch to restore changes, which we want to stay even after everything else is reverted.
What do you think?

googollee:
I don't insist on it. Cherrypicked useful changes from previous commits.

kkovaletp (approved):
LGTM

PR #1041, opened 2024-09-04
docs: Add reminder to url encode postgres password env variable

arauin:


kkovaletp:
@arauin, thanks for preparing the PR.
As the value for the password is set in the .env file. could you please move those notes (comments) from compose files to the .env example?
Also, as we have the same workflow for MariaDB/MySQL, could you please add the same comment there as well?
Finally, please update the Readme and add the same note/warning to the next sections:
- https://github.com/photoview/photoview?tab=readme-ov-file#getting-started--setup-with-docker
- https://github.com/photoview/photoview?tab=readme-ov-file#set-up-local-development-environment

kkovaletp (commented):
LGTM, but I'd approve it a bit later: we need to merge the #1040 1st.

googollee (approved):

PR #1042, opened 2024-09-05
Fix wrap nil error

bzzim:
In this case ReadVideoStreamMetadata return nil, nil because: 

```
func Wrap(err error, message string) error {
	if err == nil {
		return nil
	}
	.....
}
```

### DIFF
--- a/api/scanner/scanner_tasks/processing_tasks/process_video_task.go
+++ b/api/scanner/scanner_tasks/processing_tasks/process_video_task.go
@@ -225,7 +225,7 @@ func ReadVideoStreamMetadata(videoPath string) (*ffprobe.Stream, error) {
 
 	stream := data.FirstVideoStream()
 	if stream == nil {
-		return nil, errors.Wrapf(err, "could not get stream from file metadata (%s)", path.Base(videoPath))
+		return nil, errors.New(fmt.Sprintf("could not get stream from file metadata (%s)", path.Base(videoPath)))
 	}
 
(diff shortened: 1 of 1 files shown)


bzzim:
@kkovaletp Thanks for your quick response.
This is a bug that caused the app to crash every time it scanned a broken file (mp4 file without video - only sound). Because the scan starts over each time, it cannot be completed until the file is deleted.

https://github.com/photoview/photoview/blob/6e1e1d12ceffc9855cd05fd4d653fb7406e9fa19/api/scanner/scanner_tasks/processing_tasks/process_video_task.go#L103 
It is checked here if err != nil, but webMetadata = nil and err = nil

api/scanner/scanner_tasks/processing_tasks/process_video_task.go
```diff
@@ -225,7 +225,7 @@ func ReadVideoStreamMetadata(videoPath string) (*ffprobe.Stream, error) {
 
 	stream := data.FirstVideoStream()
 	if stream == nil {
-		return nil, errors.Wrapf(err, "could not get stream from file metadata (%s)", path.Base(videoPath))
+		return nil, errors.New(fmt.Sprintf("could not get stream from file metadata (%s)", path.Base(videoPath)))
```
googollee, line 228: Please use `fmt.Errorf(...)` here.

googollee (approved):

PR #1046, opened 2024-09-06
Bump to golang 1.23

### LINKED ISSUE #1045: Evaluate GO 1.23.1 for the project
As GO 1.23.1 is out, let's evaluate it for the project and if it is good enough for us, let's switch to this version from 1.22

googollee:
Fix #1045

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ RUN if [ "${BUILD_DATE}" = "undefined" ]; then \
   npm run build -- --base=$UI_PUBLIC_URL
 
 ### Build API ###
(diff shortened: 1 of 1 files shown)


kkovaletp:
I remember, you explained that the GO version in the `go.mod` is a feature level. Don't we want to update it as well now? Or. is it better to update it later, when some of 1.23 features will be used in our code?

Just checking to make sure we didn't forget about it)

googollee:
No need to update now. It could be done when we use new features.

kkovaletp (approved):
LGTM

PR #1049, opened 2024-09-12
Fix cache options.

googollee:
`type=local` requires providing a local path for caching. We don't need that. Using the default caching method.

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -11,6 +11,7 @@ on:
 
 env:
   IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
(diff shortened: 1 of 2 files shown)


kkovaletp:
Oh, as our goal here to not cache anything (and I didn't find `none` option), let's set the `/dev/null` path. In this case no resources are spent on caching and we're sure that it is not cached. 
What do you think?

googollee:
The empty strings here mean no sharing cache. Docker always caches layers locally. Without setting anything here, it uses Docker default caches. There is no cache actually as the workflow always runs in a fresh VM. It's what we want.

kkovaletp (approved):
Approving, as the result is the same, but you can simplify these expressions:
`${{ ( github.event_name == 'pull_request' && 'type=gha' ) || '' }}` --> `${{ github.event_name == 'pull_request' && 'type=gha' }}`

PR #1050, opened 2024-09-13
Fix cache flag in dependencies workflow.

googollee:
[Images](https://hub.docker.com/r/viktorstrate/dependencies/tags) are already pushed to Docker Hub. Pushing cache fails.

```
2024-09-13T09:58:20.6939814Z #84 exporting cache to registry
2024-09-13T09:58:20.6940171Z #84 preparing build cache for export
2024-09-13T09:58:20.8546297Z #84 writing layer sha256:01a842ec8147ad659fa8934536853b16bbc46c0a3128c25d18a2b49f40e520b5
2024-09-13T09:58:21.0471441Z #84 preparing build cache for export 0.4s done
2024-09-13T09:58:21.0472641Z #84 writing layer sha256:01a842ec8147ad659fa8934536853b16bbc46c0a3128c25d18a2b49f40e520b5 0.2s done
2024-09-13T09:58:21.0474733Z #84 ERROR: error writing layer blob: push access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
2024-09-13T09:58:21.0476149Z 
```

As the ENV is a string, it can't use `!` expression. Change to a condition.

Also add `latest` tag for images.

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -49,6 +49,7 @@ jobs:
           images: ${{ env.DOCKER_IMAGE }}
           # Docker tags based on the following events/attributes
           tags: |
+            type=raw,value=latest
             type=sha
 
       - name: Build and push
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -62,7 +62,7 @@ jobs:
           labels: ${{ steps.docker_meta.outputs.labels }}
           cache-from: ${{ env.IS_CACHING == 'true' && 'type=gha' }}
           cache-to: ${{ env.IS_CACHING == 'true' && 'type=gha,mode=max' }}
-          no-cache: ${{ ! env.IS_CACHING }}
+          no-cache: ${{ env.IS_CACHING != 'true' }}
           sbom: true
           provenance: mode=max
           annotations: ${{ steps.docker_meta.outputs.annotations }}


kkovaletp (approved):
Approved, but I'm not sure this is the final fix...
Maybe this is the time to migrate to the https://hub.docker.com/r/photoview/photoview

googollee:
I'll open a discussion about the repo.

kkovaletp:
> I'll open a discussion about the repo.

I'm afraid, we don't have other option, as the current registry won't allow us to push 1 more image.
We already have `photoview` repo in the new org registry. Also, I've created the `dependencies` repo. Both should be writable with the existing @viktorstrate credentials, as he is the owner of the new org.
He added me there at the creation time ~6 months ago, but I have only editor rights there and cannot add you to the org.

googollee:
The image could be pushed to the repo, but it still fails while [pushing cache](https://github.com/photoview/photoview/actions/runs/10848466431/job/30105631920#step:7:35270). I have no idea now.

@kkovaletp In my option, I prefer ghcr.io than Docker Hub. It works better with GitHub repo. Any idea?

kkovaletp:
https://github.com/photoview/photoview/actions/runs/10848466431/job/30105631920#step:7:17 - here I can see that it resolves the expression to 'false' - maybe this is an illegal value and the empty string is needed (my suggestion about simplification of the expression might be wrong)?
Let's comment out both caching lines for testing purposes. If it will be pushed to the registry, then we could play with those expressions to make them work.

> I prefer ghcr.io than Docker Hub. It works better with GitHub repo

Sure, but we don't have access to the Photoview's GitHub registry(

googollee:
https://github.com/photoview/photoview/pull/1052 to fix.

PR #1053, opened 2024-09-15
Use self-building dependencies and update the code with imagemagick.

googollee:
- Use self-building dependencies, including `libraw`/`libheif`/`imagemagick`/`jellyfin-ffmpeg`.
- Update go dependencies to the latest version, including `go-libheif`.
- Drop building `arm/v6` because of https://github.com/photoview/photoview/discussions/1051.
- Update the building scripts in favor of `dpkg-archtecture`, like how we build dependencies.
- Use `ImageMagick 7` instead of `darktable`.

Custom codec with `jellyfin-ffmpeg` will be in the following PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,16 +2,16 @@ name: Docker builds
 
 on:
   push:
-    branches: ['*']
+    branches: [master]
     tags:
       - v*
   pull_request:
     branches: [master]
 
 env:
-  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'googollee/photoview' }}
+  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
   DOCKER_USERNAME: viktorstrate
-  DOCKER_IMAGE: ghcr.io/googollee/photoview
+  DOCKER_IMAGE: viktorstrate/photoview
   DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
   PLATFORMS: linux/amd64,linux/arm64,linux/arm/v7
 
@@ -37,11 +37,8 @@ jobs:
         if: ${{ env.IS_PUSHING_IMAGES == 'true' }}
         uses: docker/login-action@v3
         with:
-          registry: ghcr.io
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-          # username: ${{ env.DOCKER_USERNAME }}
-          # password: ${{ env.DOCKER_PASSWORD }}
+          username: ${{ env.DOCKER_USERNAME }}
+          password: ${{ env.DOCKER_PASSWORD }}
 
       - name: Docker meta
         id: docker_meta
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,18 +2,18 @@ name: Docker builds
 
 on:
   push:
-    branches: [master]
+    branches: ['*']
     tags:
       - v*
   pull_request:
     branches: [master]
 
 env:
-  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
+  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'googollee/photoview' }}
   DOCKER_USERNAME: viktorstrate
-  DOCKER_IMAGE: viktorstrate/photoview
+  DOCKER_IMAGE: ghcr.io/googollee/photoview
   DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
-  PLATFORMS: linux/amd64,linux/arm64,linux/arm/v7,linux/arm/v6
+  PLATFORMS: linux/amd64,linux/arm64,linux/arm/v7
 
 jobs:
   build:
@@ -37,8 +37,11 @@ jobs:
         if: ${{ env.IS_PUSHING_IMAGES == 'true' }}
         uses: docker/login-action@v3
         with:
-          username: ${{ env.DOCKER_USERNAME }}
-          password: ${{ env.DOCKER_PASSWORD }}
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          # username: ${{ env.DOCKER_USERNAME }}
+          # password: ${{ env.DOCKER_PASSWORD }}
 
       - name: Docker meta
         id: docker_meta
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: [master]
+    branches: ['*']
   pull_request:
     branches: [master]
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: ['*']
+    branches: [master]
   pull_request:
     branches: [master]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,13 +49,28 @@ ENV PATH="${GOPATH}/bin:${PATH}"
 ENV CGO_ENABLED=1
 
 # Download dependencies
-COPY scripts/*.sh /app/scripts/
+COPY scripts/set_compiler_env.sh /app/scripts/
 RUN chmod +x /app/scripts/*.sh \
+  && source /app/scripts/set_compiler_env.sh
+
+COPY --from=viktorstrate/dependencies /artifacts.tar.gz /dependencies/
+RUN export $(cat /env) \
+  && cd /dependencies/ \
+  && tar xfv artifacts.tar.gz \
+  && cp -a include/* /usr/local/include/ \
+  && cp -a pkgconfig/* ${PKG_CONFIG_PATH} \
+  && cp -a lib/* /usr/local/lib/ \
+  && cp -a bin/* /usr/local/bin/ \
+  && apt-get install -y ./deb/jellyfin-ffmpeg.deb
+
+COPY scripts/install_*.sh /app/scripts/
+RUN chmod +x /app/scripts/*.sh \
+  && export $(cat /env) \
   && /app/scripts/install_build_dependencies.sh \
   && /app/scripts/install_runtime_dependencies.sh
 
 COPY api/go.mod api/go.sum /app/api/
-RUN source /app/scripts/set_compiler_env.sh \
+RUN export $(cat /env) \
   && go env \
   && go mod download \
   # Patch go-face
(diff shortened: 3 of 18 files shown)


googollee:
https://github.com/googollee/photoview/pkgs/container/photoview/273926913?tag=dep-magick is built with this branch. You can try it with the image.

api/scanner/media_encoding/executable_worker/magick_cli.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/photoview/photoview/api/utils"
+)
+
+type MagickCli struct {
+	path string
+}
+
+func newMagickCli() *MagickCli {
+	if utils.EnvDisableRawProcessing.GetBool() {
+		log.Printf("Executable worker disabled (%s=1): ImageMagick\n", utils.EnvDisableRawProcessing.GetName())
```
kkovaletp, line 18: Assuming that we log only the variable name. Don't we need to also include the boolean value as well?

dependencies/prepare.sh
```diff
@@ -11,7 +11,7 @@ DEBIAN_ARCH=$TARGETARCH
 if [ "$TARGETARCH" = "arm" ]
 then
   DEBIAN_ARCH=armhf
-  if [ "$TARGETVARIANT" = "v5" ]
+  if [ "$TARGETVARIANT" = "v5" || "$TARGETVARIANT" = "v6" ]
```
kkovaletp, line 14: As both arch's are dropped, don't we need to remove this `if`?

scripts/install_build_dependencies.sh
```diff
@@ -1,42 +1,12 @@
...
-  dpkg --add-architecture amd64
-  DEBIAN_ARCH='amd64'
-fi
+: ${DEB_HOST_ARCH=`dpkg --print-architecture`}
+echo Arch: ${DEB_HOST_ARCH}
 
-apt-get update
-
-# Install G++/GCC cross compilers
-if [ "$DEBIAN_ARCH" == "arm64" ]; then
```
kkovaletp, line 18: Just curious: why we don't need these 2 deps at all?

kkovaletp (commented):
Some questions and comments

Dockerfile
```diff
@@ -66,30 +81,44 @@ RUN source /app/scripts/set_compiler_env.sh \
...
-FROM --platform=${BUILDPLATFORM:-linux/amd64} debian:bookworm-slim AS release
+FROM debian:bookworm-slim AS release
 ARG TARGETPLATFORM
 
 # See for details: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
 COPY scripts/install_runtime_dependencies.sh /app/scripts/
-RUN chmod +x /app/scripts/install_runtime_dependencies.sh \
+RUN --mount=type=bind,from=api,source=/dependencies/,target=/dependencies/ \
```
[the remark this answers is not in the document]
googollee: `COPY --form` creates another layer, while `RUN --mount` won't. Since there is no need to keep the `deb` file after installation, I prefer not adding another layer.

Dockerfile
```diff
@@ -66,30 +81,44 @@ RUN source /app/scripts/set_compiler_env.sh \
...
-  && /app/scripts/install_runtime_dependencies.sh
-
-WORKDIR /home/photoview
+  # Install required dependencies
+  && /app/scripts/install_runtime_dependencies.sh \
+  # Install self-building libs
+  && cd /dependencies \
+  && cp -a lib/* /usr/local/lib/ \
+  && cp -a bin/* /usr/local/bin/ \
+  && ldconfig \
```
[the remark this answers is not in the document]
googollee: Not necessary, but no harm if we run it.

`ldconfig` updates LD configures with new `so` files, which the runtime requires. During compiling around line 64, the compiler reads `a`/`h` files with `pkg-config` but not `so`.

I added `ldconfig` in line 64 for the completeness.

api/scanner/media_encoding/executable_worker/magick_cli.go
googollee: Updated.

dependencies/prepare.sh
googollee: I hope to cover all archs here since our code can be compiled with other arm archs.

Updated to emphasize `v7`.

scripts/install_runtime_dependencies.sh
```diff
@@ -1,22 +1,13 @@
...
-rm -rf /var/lib/apt/lists/*
+apt-get install -y curl exiftool
+
+# libheif dependencies: no dependency
+# libraw dependencies
+apt-get install -y libjpeg62-turbo liblcms2-2
+# ImageMagick dependencies
+apt-get install -y libjbig0 libtiff6 libfreetype6 libjxl0.7 liblqr-1-0 libpng16-16 libdjvulibre21 libwebpmux3 libwebpdemux2 libwebp7 libopenexr-3-1-30 libopenjp2-7 libjpeg62-turbo liblcms2-2 libxml2 libx11-6 libgomp1
+# go-face dependencies
+apt-get install -y libdlib19.1 libblas3 liblapack3 libjpeg62-turbo
```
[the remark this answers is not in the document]
googollee: I want to separate dependencies for different libraries. It won't reinstall the package if it's already done, so calling multiple times won't increase much time.

A multiline command is sometimes used to avoid a line too long. I tried that before and I found using multiple `apt-get` is clearer than that.

All packages listed here are the direct-dependencies of each library. If we remove/switch any packages in the future, just removing the correspond line.

scripts/set_compiler_env.sh
```diff
@@ -1,84 +1,54 @@
...
+set -eu
 
-# Script to configure environment variables for Go compiler
-# to allow cross compilation
-
-: ${TARGETPLATFORM=}
-: ${TARGETOS=}
-: ${TARGETARCH=}
-: ${TARGETVARIANT=}
+# Script to configure environment variables for  compiler to cross compilation
```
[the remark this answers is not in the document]
googollee: Update.

scripts/set_compiler_env.sh
```diff
@@ -1,84 +1,54 @@
...
 
-if [ ! -z "$TARGETPLATFORM" ]; then
-  TARGETOS="$(echo $TARGETPLATFORM | cut -d"/" -f1)"
-  TARGETARCH="$(echo $TARGETPLATFORM | cut -d"/" -f2)"
-  TARGETVARIANT="$(echo $TARGETPLATFORM | cut -d"/" -f3)"
+DEBIAN_ARCH=$TARGETARCH
+if [ "$TARGETARCH" = "arm" ]
+then
+  DEBIAN_ARCH=armhf
+  if [ "$TARGETVARIANT" = "v5" || "$TARGETVARIANT" = "v6" ]
```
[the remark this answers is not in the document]
googollee: Updated, the same as `dependencies/prepare.sh`.

scripts/set_compiler_env.sh
```diff
@@ -1,84 +1,54 @@
...
+TARGETOS="$(echo $TARGETPLATFORM | cut -d"/" -f1)"
+TARGETARCH="$(echo $TARGETPLATFORM | cut -d"/" -f2)"
+TARGETVARIANT="$(echo $TARGETPLATFORM | cut -d"/" -f3)"
 
-if [ ! -z "$TARGETPLATFORM" ]; then
-  TARGETOS="$(echo $TARGETPLATFORM | cut -d"/" -f1)"
-  TARGETARCH="$(echo $TARGETPLATFORM | cut -d"/" -f2)"
-  TARGETVARIANT="$(echo $TARGETPLATFORM | cut -d"/" -f3)"
+DEBIAN_ARCH=$TARGETARCH
+if [ "$TARGETARCH" = "arm" ]
```
kkovaletp, line 13: I'm not sure that this `if` is correct, as currently, we support only `arm/v7`. `arm64` won't go into this `if`. So, from my PoV the logic should be as simple as if `$TARGETARCH` is `arm` and `$TARGETVARIANT` is `v7`, then `DEBIAN_ARCH=armhf`.
What do you think?

scripts/set_compiler_env.sh
```diff
@@ -1,84 +1,54 @@
...
 
-if [ "$CGO_ENABLED" = "1" ]; then
-  case "$GOARCH" in
-  "amd64")
-    export COMPILER_ARCH="x86_64-linux-gnu"
-    ;;
-  "ppc64le")
-    export COMPILER_ARCH="powerpc64le-linux-gnu"
+GOARM="7"
+if [ "$TARGETARCH" = "arm" && ! -z "$TARGETVARIANT" ]; then
```
kkovaletp, line 34: I guess that you forgot to update this code and remove unsupported arch's

scripts/install_runtime_dependencies.sh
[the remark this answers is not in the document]
kkovaletp: Ok, so I guess that this form still meets your goals:
```bash
apt-get install -y \
  curl exiftool \
  # libheif dependencies: no dependency
  # libraw dependencies
  libjpeg62-turbo liblcms2-2 \
  # ImageMagick dependencies
  libjbig0 libtiff6 libfreetype6 libjxl0.7 liblqr-1-0 libpng16-16 libdjvulibre21 libwebpmux3 libwebpdemux2 libwebp7 \
  libopenexr-3-1-30 libopenjp2-7 libjpeg62-turbo liblcms2-2 libxml2 libx11-6 libgomp1 \
  # go-face dependencies
  libdlib19.1 libblas3 liblapack3 libjpeg62-turbo
```
What do you think?

dependencies/prepare.sh
kkovaletp: Hmm, I thought that we drop the support of `v6` completely from our product. Would you like to keep it for the direct host deployment scenario?
I'm concerned about the possible complexity of our code and scripts in the case of different lists of arch's for different deployments.

kkovaletp (commented):
Couple more comments

scripts/install_runtime_dependencies.sh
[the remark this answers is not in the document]
googollee: Well, why not. Updated.

dependencies/prepare.sh
googollee: I want to follow the original script and cover other archs in the script, at least covering what Debian supports. We drop `arm/v6` because of dependencies. The code still works with `arm/v6` aka `armel`. [The original script](https://github.com/photoview/photoview/blob/a6cbfc76f3d4acbf9dd3e5767a34ba319fc4cd56/scripts/set_compiler_env.sh#L56) supports `ppc` and `s390x` even we don't say we support them.

scripts/set_compiler_env.sh
googollee: See [the previous thread](https://github.com/photoview/photoview/pull/1053#discussion_r1761715575).

scripts/set_compiler_env.sh
googollee: See [the previous thread](https://github.com/photoview/photoview/pull/1053#discussion_r1761715575).

scripts/install_runtime_dependencies.sh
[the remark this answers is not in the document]
googollee: Now we know that command doesn't work, see [here](https://github.com/photoview/photoview/actions/runs/10899558813/job/30245223797?pr=1053#step:5:1677).

Reverted.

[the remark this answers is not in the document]
googollee: `shell`/`bash` doesn't support mix commands and comments in a multi-line command.

It works with `Dockerfile` because Docker parses the file in a wrong way. Docker removes all comment lines first, then calls the `RUN` command with a shell.

dependencies/prepare.sh
kkovaletp: But you've created this original script, as far as I remember, right?
Ok, as there is no negative impact of having these arch's there, I'm ok if you'd like to keep them. Just think that this is a 'dead code'. We could get back to it later when doing some other changes in that place.

scripts/install_runtime_dependencies.sh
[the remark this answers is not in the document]
kkovaletp: Oh, ok. Then, sure, let's keep them as is...
Very strange... I was sure that it works, as I used it many times in Dockerfiles and CI YAML files...

kkovaletp:
As all the questions now answered, please finalize the PR and revert the testing code. Then I'll approve it

dependencies/prepare.sh
googollee: It renamed from [`go_wrapper.sh`](https://github.com/photoview/photoview/blob/6693920b61b1afda29786566b418ba7df9ffd628/docker/go_wrapper.sh).

If we drop other archs in the script, I'd like to fail the script with non-supported archs. How about that?

scripts/install_runtime_dependencies.sh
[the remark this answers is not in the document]
googollee: Both Dockerfile and YAML files are parsed with two steps, first is Dockerfile or YAML format, which erases all comments in the file, then put commands to a shell. It's not the same way as shell parsing scripts.

dependencies/prepare.sh
kkovaletp: Ah, I see.
Let's keep it as is for now, as there is no negative impact. We always could remove those archs later when we are faced with an issue

.github/workflows/build.yml
```diff
@@ -2,18 +2,18 @@ name: Docker builds
 
 on:
   push:
-    branches: [master]
+    branches: ['*']
```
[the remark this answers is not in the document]
googollee: Done.

.github/workflows/tests.yml
```diff
@@ -2,7 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: [master]
+    branches: ['*']
```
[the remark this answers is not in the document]
googollee: Reverted.

kkovaletp (approved):
Approved

PR #1058, opened 2024-09-20
Fix #677: Support RAW image with JPEG version without installing Magick

### LINKED ISSUE #677: When RAW processing is disabled, already existing JPG versions are also ignored
**Describe the bug**
When RAW processing is disabled with `PHOTOVIEW_DISABLE_RAW_PROCESSING=1`, Photoview will not scan and display already existing JPG versions of the RAW files.

**To Reproduce**
1. disable RAW processing
2. scan a folder that has RAW+JPEG images

**Expected behavior**
I would expect Photoview to still display photos that already have an existing JPG version, and simply not process additional RAW files _without_ an existing JPG companion. 

**Your environment:**

 - Are you using Docker? **yes**
 - Database **mariadb 10.5**
 - Browser **any**
 - Version `2.3.12 ([0b9d04ef](https://github.com/photoview/photoview/commit/0b9d04eff2b6f896a70e8d081b6706dc4068f732))`

WindLi001:
This PR fix issue "When RAW processing is disabled, already existing JPG versions are also ignored #677".

According to code in EncodeHighRes(), when a RAW image has a counterpart JPEG file, there is no need to encode it to JPEG with Magick. That is, Magick isn't necessary in this situation. 

But because the installation of Magick is checked in IsSupported() when a RAW image is processed, the RAW image with counterpart JPEG file won't be displayed if Magick isn't installed.

This PR deletes the check of the installation of Magick in IsSupported() to fix it.

### DIFF
--- a/api/scanner/media_type/media_type.go
+++ b/api/scanner/media_type/media_type.go
@@ -260,7 +260,7 @@ func (imgType *MediaType) IsSupported() bool {
 		return true
 	}
 
-	if executable_worker.Magick.IsInstalled() && imgType.IsRaw() {
+	if imgType.IsRaw() {
 		return true
 	}
 


kkovaletp:
@WindLi001, thanks for the PR. The fix looks very simple) 
Would it be possible to add a unit test to cover the changed line and pass the coverage check, also increasing overall code coverage?

@googollee, when you have time, could you please review it to make sure that the fix correctly handles all possible cases and doesn't introduce any new bugs?

WindLi001:
> @WindLi001, thanks for the PR. The fix looks very simple) Would it be possible to add a unit test to cover the changed line and pass the coverage check, also increasing overall code coverage?

I only change the code of IsSupported(), and I find there is no unit test function of it. Do you mean writing a unit test function of IsSupported()? OK, I'll try it.

googollee:
Sorry, I don't agree on this fix. This changes the function of `MediaType.IsSupported()`. From `EncodeMediaData.EncodeHighRes()` and `ProcessPhotoTask.ProcessMedia()`, it should be possible to convert the side-car JPEG file to a high-resolution caching file. I need to do some test and see how it works.

googollee:
I propose the change of https://github.com/photoview/photoview/pull/1096.

kkovaletp:
The #1096 looks better from my PoV, so once it merged, I'd close this PR if there are no objections

PR #1066, opened 2024-09-30
Include all heif plugins

### LINKED ISSUE #1062: Update libheif to fix issues with heif/heic photos taken in iOS 18
**Describe the bug**
libheif has problems with parsing photos taken in iOS 18.
It has been fixed in v1.18.0, as stated in this issue: https://github.com/strukturag/libheif/issues/1190

**To Reproduce**
Try to scan a .HEIC file that was created in iOS 18.
Example: https://www.dropbox.com/scl/fi/zv78otopss4hdhko2iy2w/IMG_0928.HEIC?rlkey=llqr6erkwbccse94iokklu22t&st=1rdf0i3h&dl=0

**Expected behavior**
The file gets scanned correctly

**Actual behavior**
ERROR: image decoding: failed to decode HEIF image (/photos/iPhone/IMG_0928.HEIC): Error while loading plugin: No decoding plugin installed for this compression format

googollee:
Fix #1062

### DIFF
--- a/dependencies/build_libheif.sh
+++ b/dependencies/build_libheif.sh
@@ -21,6 +21,8 @@ cd ..
 mkdir -p /output/bin /output/lib /output/include /output/pkgconfig
 cp -a /usr/local/bin/heif-* /output/bin/
 cp -a /usr/local/lib/libheif* /output/lib/
+cp -a /usr/lib/${DEB_HOST_MULTIARCH}/libheif/plugins/libheif-*.so /output/lib
+cp -a /usr/local/lib/${DEB_HOST_MULTIARCH}/gdk-pixbuf-*/*/loaders/libpixbufloader-heif.so /output/lib
 cp -a /usr/local/lib/pkgconfig/libheif* /output/pkgconfig/
 cp -a /usr/local/include/libheif /output/include/
 file /usr/local/lib/libheif.so*
--- a/dependencies/build_libheif.sh
+++ b/dependencies/build_libheif.sh
@@ -21,8 +21,6 @@ cd ..
 mkdir -p /output/bin /output/lib /output/include /output/pkgconfig
 cp -a /usr/local/bin/heif-* /output/bin/
 cp -a /usr/local/lib/libheif* /output/lib/
-cp -a /usr/lib/${DEB_HOST_MULTIARCH}/libheif/plugins/libheif-*.so /output/lib
-cp -a /usr/local/lib/${DEB_HOST_MULTIARCH}/gdk-pixbuf-*/*/loaders/libpixbufloader-heif.so /output/lib
 cp -a /usr/local/lib/pkgconfig/libheif* /output/pkgconfig/
 cp -a /usr/local/include/libheif /output/include/
 file /usr/local/lib/libheif.so*
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,16 +2,16 @@ name: Docker builds
 
 on:
   push:
-    branches: ['*']
+    branches: [master]
     tags:
       - v*
   pull_request:
     branches: [master]
 
 env:
-  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'googollee/photoview' }}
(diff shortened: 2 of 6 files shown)


dependencies/build_libheif.sh
```diff
@@ -20,6 +20,8 @@ cd ..
 
 mkdir -p /output/lib /output/include /output/pkgconfig
 cp -a /usr/local/lib/libheif* /output/lib/
+cp -a /usr/lib/${DEB_HOST_MULTIARCH}/libheif/plugins/libheif-*.so /output/lib
+cp -a /usr/local/lib/${DEB_HOST_MULTIARCH}/gdk-pixbuf-*/*/loaders/libpixbufloader-heif.so /output/lib
```
kkovaletp, line 24: So, it looks like the plugins are installed during the build, but we forgot to copy them to the final image.
Is there a chance, anything else is forgotten?

BTW, I see that these `cp` commands don't save the original location paths of copied *.so files. Are you sure that it is fine and the lib will find them on the final image even if they are located not on the same paths as were installed during the build?

googollee: There is no other files for `libraw`. For `imagemagick`, I see some configuration xml which should be included, like `quantization-table.xml`.

I updated to include those files, even we don't use them now.

`ldconfig` updates all so files. So if all libraries follow the `ld`, they should not depend on the file path.

googollee: I also include other binaries building from dependencies. It's useful for debug.

kkovaletp:
It looks good in general, but there are build errors for `dependencies` and `photoview` images. I guess that you need to solve the error for `dependencies, push the new image, and only then the `photoview` build should run fine, as it pulls `dependencies` from the repo async

googollee:
@kkovaletp I updated the script to build all plugins into `libheif.so` now. It doesn't have any other plugin files. It's easy to maintain.

Also updated the runtime dependencies.

googollee:
> but there are build errors for `dependencies` and `photoview` images.

`photoview` building fails because it requires to copy `output/etc` for ImageMagick 7 configurations now. As current `dependencies` image doesn't have those files, it fails.

googollee:
Now the final image is increased to 1.1g. Should we optimize it now or after fixing iOS images?

kkovaletp:
> As current `dependencies` image doesn't have those files, it fails

I guess, it is better to prepare a separate PR with just changes for that image: after it gets merged, this one can succeed. If combine all changes in 1 PR (this one), then it will fail even in `master`, as these pipelines are independent and the `build` pipeline will pull older `dependencies` image, while the current one is building

kkovaletp:
> Now the final image is increased to 1.1g. Should we optimize it now or after fixing iOS images?

Let's fix the issue here and finish other pending high-prio tasks. I hope that once we move non-RAW image processing to ImageMagick and ensure all the processing is done by external workers, we can remove many unneeded dependencies, which will release some space. Anyway, we could optimize the final image before the release

googollee:
I did some cleanup and the image size drops to 800MiB. Most of the increase (~400MiB) is `jellyfin-ffmpeg` and its dependencies.

kkovaletp:
> I did some cleanup

Are you sure that we don't need those deps? Won't we get a similar issue because of missing deps?

googollee:
> Are you sure that we don't need those deps? Won't we get a similar issue because of missing deps?

- Removed x11 support in imagemagick
- Removed gtk pixbuf support in libheif
- Use dav1d for libheif avif decoding
- Use aom for libheif avif encoding, instead of all choices AOM, rav1e, svt-av1
- Use libde265 for libheif heic decoding, instead of all choices libde265, ffmpeg (this removes a lot in size)
- Use x265 for libheif heic encoding

So yes, I think it doesn't remove any format supports. [libheif](https://github.com/strukturag/libheif?tab=readme-ov-file#supported-features) provides different packages for one codec. We uses only one for each codec to save spaces.

kkovaletp:
Looks good. Revert debug-related changes when you finish the testing, and I'll approve

kkovaletp (approved):
LGTM

PR #1067, opened 2024-09-30
Build libheif with all dependencies in one so file.

googollee:
Related with #1062


kkovaletp (approved):
LGTM

googollee:
It also cleans dependencies. Now `libheif` only has one dep for heic and one dep for avif. See https://github.com/strukturag/libheif?tab=readme-ov-file#supported-features.

googollee:
`magick` works well now with the pic in #1062.

PR #1065, opened 2024-09-29
Refactoring part 2: Long func refactoring

kkovaletp:
Split long functions into smaller parts to decrease cognitive complexity

This PR is related to the #953 and depends on the #1064 and should be merged after it

PR #1073, opened 2024-10-03
Refactoring part 2.1: Long func refactoring in `database`

kkovaletp:
PR is part of the #1065

### DIFF
--- a/api/database/migration_exif.go
+++ b/api/database/migration_exif.go
@@ -11,6 +11,42 @@ import (
 	"gorm.io/gorm"
 )
 
+type exifModel struct {
+	ID       int `gorm:"primarykey"`
+	Exposure *string
+	Flash    *string
+}
+
+var flashDescriptions = map[int]string{
+	0x0:  "No Flash",
+	0x1:  "Fired",
+	0x5:  "Fired, Return not detected",
+	0x7:  "Fired, Return detected",
+	0x8:  "On, Did not fire",
+	0x9:  "On, Fired",
+	0xD:  "On, Return not detected",
+	0xF:  "On, Return detected",
+	0x10: "Off, Did not fire",
+	0x14: "Off, Did not fire, Return not detected",
+	0x18: "Auto, Did not fire",
+	0x19: "Auto, Fired",
+	0x1D: "Auto, Fired, Return not detected",
+	0x1F: "Auto, Fired, Return detected",
+	0x20: "No flash function",
+	0x30: "Off, No flash function",
+	0x41: "Fired, Red-eye reduction",
+	0x45: "Fired, Red-eye reduction, Return not detected",
+	0x47: "Fired, Red-eye reduction, Return detected",
+	0x49: "On, Red-eye reduction",
+	0x4D: "On, Red-eye reduction, Return not detected",
+	0x4F: "On, Red-eye reduction, Return detected",
+	0x50: "Off, Red-eye reduction",
+	0x58: "Auto, Did not fire, Red-eye reduction",
+	0x59: "Auto, Fired, Red-eye reduction",
+	0x5D: "Auto, Fired, Red-eye reduction, Return not detected",
+	0x5F: "Auto, Fired, Red-eye reduction, Return detected",
+}
+
 // Migrate MediaExif fields "exposure" and "flash" from strings to integers
 func migrateExifFields(db *gorm.DB) error {
 	mediaExifColumns, err := db.Migrator().ColumnTypes(&models.MediaEXIF{})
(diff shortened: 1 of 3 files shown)


api/database/migration_exif.go
```diff
@@ -51,11 +71,35 @@ func migrateExifFields(db *gorm.DB) error {
 
 		return nil
 	})
+}
 
-	if err != nil {
-		return err
+func processFlash(exifCol gorm.ColumnType, db *gorm.DB) error {
```
googollee, line 76: `normalizeFlash`?

api/database/migration_exif.go
```diff
@@ -51,11 +71,35 @@ func migrateExifFields(db *gorm.DB) error {
...
+			// do migration
+			if err := migrateExifFieldsFlash(db); err != nil {
+				return err
+			}
+		}
 	}
+	return nil
+}
 
+func processExposure(exifCol gorm.ColumnType, db *gorm.DB) error {
```
googollee, line 91: `normalizeExposure`?

api/database/migration_exif.go
```diff
@@ -68,13 +112,21 @@ func migrateExifFieldsExposure(db *gorm.DB) error {
...
-			Exposure *string
-		}
 		var results []exifModel
 
-		return tx.Model(&exifModel{}).Table("media_exif").Where("exposure LIKE '%/%'").FindInBatches(&results, 100, func(tx *gorm.DB, batch int) error {
+		return calculateExposure(tx, results)
+	})
+
+	if err != nil {
+		return errors.Wrap(err, "migrating `media_exif.exposure` failed")
```
googollee, line 121: Since we are changing this line, it's better to follow best practices of error.

```
return fmt.Errorf("migrating `media_exif.exposure` failed: %w", err)
```

api/database/migration_exif.go
```diff
@@ -131,43 +179,21 @@ func migrateExifFieldsFlash(db *gorm.DB) error {
...
-			0x58: "Auto, Did not fire, Red-eye reduction",
-			0x59: "Auto, Fired, Red-eye reduction",
-			0x5D: "Auto, Fired, Red-eye reduction, Return not detected",
-			0x5F: "Auto, Fired, Red-eye reduction, Return detected",
-		}
+		return replaceFlashValues(tx, results)
+	})
+
+	if err != nil {
+		return errors.Wrap(err, "migrating `media_exif.flash` failed")
```
googollee, line 188: Same here to use `fmt.Errorf()`.

googollee (approved):
Agree on this change. Adding some suggestions.

api/database/migration_exif.go
kkovaletp: regarding the "process" part of the names here and later: what do you think about "convert" or "migrate" alternatives? I'm not a native speaker) but "normalize" doesn't look accurate in this case, as it means (to me) to align something with a "normal" reference value...

api/database/migration_exif.go
kkovaletp: Thanks, I'll change it, but could you please explain the difference between these 2 expressions and when is it better to use which?
I'll then go through other PRs in advance, as we have such calls across all the project.

api/database/migration_exif.go
googollee: How about `parse`? I see the function `migrateExifFieldsFlash` converts the flash value from a string to a int, according to [the map](https://github.com/photoview/photoview/pull/1073/files/43199f4bace36bc9eb13de68424bbd5acd61abb1#diff-9b219c58eb0d2e05ccb3c24f2ee2924e670ca7526eabb158521272e2dfe034f0R20).

To me,  `convert` is a general word, no special meaning. `migrate` means converting a valid data from one pattern to another patten. `parse` means converting a general data to a structed data.

api/database/migration_exif.go
googollee: [`fmt.Errorf()`](https://pkg.go.dev/fmt#Errorf) is an official verison of `errors.Warpf()`. They do the same thing, to attach a context info to an error. `fmt.Errorf()` works like `fmt.Printf()` and returns an error. It introduces a new placeholder `%w` to wrap an error, and put the string returned by `error.Error()`.

In the other hand, https://github.com/pkg/errors is already archived.

api/database/migration_exif.go
kkovaletp: done

api/database/migration_exif.go
kkovaletp: done

api/database/migration_exif.go
kkovaletp: Done in all the module, please review

api/database/migration_exif.go
kkovaletp: done

googollee (approved):

googollee:
Approved.

PR #1074, opened 2024-10-03
Refactoring part 2.2: Long func refactoring in `dataloader`

kkovaletp:
PR is part of the #1065

### DIFF
--- a/api/dataloader/mediaURLLoader.go
+++ b/api/dataloader/mediaURLLoader.go
@@ -56,7 +56,9 @@ func NewHighresMediaURLLoader(db *gorm.DB) *MediaURLLoader {
 		maxBatch: 100,
 		wait:     5 * time.Millisecond,
 		fetch: makeMediaURLLoader(db, func(query *gorm.DB) *gorm.DB {
-			return query.Where("purpose = ? OR (purpose = ? AND content_type IN ?)", models.PhotoHighRes, models.MediaOriginal, media_type.WebMimetypes)
+			return query.
+				Where("purpose = ? OR (purpose = ? AND content_type IN ?)",
+					models.PhotoHighRes, models.MediaOriginal, media_type.WebMimetypes)
 		}),
 	}
 }
--- a/api/dataloader/userFavoriteLoader.go
+++ b/api/dataloader/userFavoriteLoader.go
@@ -19,7 +19,6 @@ func NewUserFavoriteLoader(db *gorm.DB) *UserFavoritesLoader {
 				userIDMap[key.UserID] = struct{}{}
 				mediaIDMap[key.MediaID] = struct{}{}
 			}
-			
 
 			uniqueUserIDs := make([]int, len(userIDMap))
 			uniqueMediaIDs := make([]int, len(mediaIDMap))
@@ -37,24 +36,33 @@ func NewUserFavoriteLoader(db *gorm.DB) *UserFavoritesLoader {
 			}
 
 			var userMediaFavorites []*models.UserMediaData
-			err := db.Where("user_id IN (?)", uniqueUserIDs).Where("media_id IN (?)", uniqueMediaIDs).Where("favorite = TRUE").Find(&userMediaFavorites).Error
+			err := db.
+				Where("user_id IN (?)", uniqueUserIDs).
+				Where("media_id IN (?)", uniqueMediaIDs).
+				Where("favorite = TRUE").
+				Find(&userMediaFavorites).Error
 			if err != nil {
 				return nil, []error{err}
 			}
 
 			result := make([]bool, len(keys))
-			for i, key := range keys {
-				favorite := false
-				for _, fav := range userMediaFavorites {
-					if fav.UserID == key.UserID && fav.MediaID == key.MediaID {
-						favorite = true
-						break
-					}
-				}
-				result[i] = favorite
-			}
+			result = iterateFavorites(keys, userMediaFavorites, result)
 
 			return result, nil
 		},
 	}
+}
+
+func iterateFavorites(keys []*models.UserMediaData, userMediaFavorites []*models.UserMediaData, result []bool) []bool {
+	for i, key := range keys {
+		favorite := false
+		for _, fav := range userMediaFavorites {
+			if fav.UserID == key.UserID && fav.MediaID == key.MediaID {
+				favorite = true
+				break
+			}
+		}
+		result[i] = favorite
+	}
+	return result
 }
--- a/api/dataloader/loaders.go
+++ b/api/dataloader/loaders.go
@@ -8,7 +8,9 @@ import (
 	"gorm.io/gorm"
 )
 
-const loadersKey = "dataloaders"
+type contextKey string
+
+const loadersKey contextKey = "dataloaders"
 
 type Loaders struct {
 	MediaThumbnail      *MediaURLLoader
--- a/api/dataloader/userLoader.go
+++ b/api/dataloader/userLoader.go
@@ -19,7 +19,10 @@ func NewUserLoaderByToken(db *gorm.DB) *UserLoader {
 				return nil, []error{err}
 			}
 
-			rows, err := db.Table("access_tokens").Select("distinct user_id").Where("expire > ?", time.Now()).Where("value IN (?)", tokens).Rows()
+			rows, err := db.Table("access_tokens").
+				Select("distinct user_id").
+				Where("expire > ?", time.Now()).Where("value IN (?)", tokens).
+				Rows()
 			if err != nil {
 				return nil, []error{err}
 			}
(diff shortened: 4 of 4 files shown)


api/dataloader/mediaURLLoader.go
```diff
@@ -56,7 +56,9 @@ func NewHighresMediaURLLoader(db *gorm.DB) *MediaURLLoader {
 		maxBatch: 100,
 		wait:     5 * time.Millisecond,
 		fetch: makeMediaURLLoader(db, func(query *gorm.DB) *gorm.DB {
-			return query.Where("purpose = ? OR (purpose = ? AND content_type IN ?)", models.PhotoHighRes, models.MediaOriginal, media_type.WebMimetypes)
+			return query.
```
googollee, line 59: The format of this line is weird. Do you use `gofmt`?

In the other hand, I don't see good benefit to limit the size of a line. Morden editor can handle a long line well.

api/dataloader/userFavoriteLoader.go
```diff
@@ -37,24 +36,33 @@ func NewUserFavoriteLoader(db *gorm.DB) *UserFavoritesLoader {
...
-				result[i] = favorite
-			}
+			result = iterateFavorites(keys, userMediaFavorites, result)
 
 			return result, nil
 		},
 	}
 }
+
+func iterateFavorites(keys []*models.UserMediaData, userMediaFavorites []*models.UserMediaData, result []bool) []bool {
```
googollee, line 56: The purpose of breaking up long functions is to facilitate testing. I want to:

- Add tests to (at least) new functions when breaking them up.
- Alternatively, keep the functions until we have added tests, then break them up

api/dataloader/mediaURLLoader.go
kkovaletp: Well, maybe I'm too old) But to me, long unsplit lines look not easily readable. Even with IDE's word wrapping.

I started my programming journey with Pascal/Delphi, then switched to Java, then to Python. Now getting familiar with Go. I used to see a visual code structure and even when dealing with long OOP calls or Java stream processing, I usually put each item of the chain to its own line with indention, representing the dependency level. It helps me to easily see the code structure.

BTW, I executed `gofmt ./api/dataloader/mediaURLLoader.go` and it returned me exactly the same formatting:
```
func NewHighresMediaURLLoader(db *gorm.DB) *MediaURLLoader {
        return &MediaURLLoader{
                maxBatch: 100,
                wait:     5 * time.Millisecond,
                fetch: makeMediaURLLoader(db, func(query *gorm.DB) *gorm.DB {
                        return query.
                                Where("purpose = ? OR (purpose = ? AND content_type IN ?)",
                                        models.PhotoHighRes, models.MediaOriginal, media_type.WebMimetypes)
                }),
        }
}
```

api/dataloader/userFavoriteLoader.go
kkovaletp: Testing is the important part of the refactoring, but I don't want to write tests in the scope of this round, as it will take forever to cover all the changed code in all the PRs. Also, I'm not so good in Go to write unit tests without using asserting lib (which, you've mentioned earlier, isn't so stable in Go).

But, writing tests is not the only purpose. Extracting nested code and decreasing the function's cognitive complexity helps to easily read the code and better understand the logic of the function without the need to keep unnecessary details in mind. I think that it is enough reason for these changes.

Of course, later everybody is more than welcome to write unit and integration tests for the code, rewrite the code in a more optimal way, etc.

api/dataloader/mediaURLLoader.go
googollee: Oh, ok. I misreaded it as spaces. LGTM.

api/dataloader/userFavoriteLoader.go
googollee: I'd like to test codes before refactoring. Without testing, it's hard to say the refactoring is right.

Decreasing the function's cognitive complexity doesn't produce the easy-reading code directly. For good code, it requires to analyze concepts from the business before coding, but not give a proper name to a group of existing code.

For example, the name `iterateFavorites` doesn't tell anything. What does the function do when iterating? How to control the itration? It still needs to go through code to understand the function that it returns favours statuses of given `keys`. A proper signature may be `func returnFavoritesStatus(input []*models.UserMediaData, favourData []*models.UserMediaData) []bool`. Even without testing, it's not a good function. It passes a `result` and returns another result, and they are refered to the same instance. The function should create `[]bool` inside and return that value, instead of creating the instance before the function calling.

And for this function, I think a better way is querying favours status through sql directly, instead of grabbing all raw data and calculating them. It reduces the memory and io.

My point is, just decreasing the function's cognitive complexity can't generate good code. It may introduce bugs without tests, which are hard to be found during review. For existing code, if lint generates warnings of the complexity, just keep them and improve the code when we change them.

kkovaletp: > I'd like to test codes before refactoring. Without testing, it's hard to say the refactoring is right.

Sure, I agree with you. But, as I mentioned, this time I didn't change the code's logic and just extracted the nested code to separate private (local) functions. Besides that, I'll spend a lot of time writing tests for all the changed code, as I'm not so good in Go, while most of the code was touched.

> For good code, it requires analyzing concepts from the business before coding, but not giving a proper name to a group of existing code.

Sure, but we already have the legacy code. I just re-formatted it to formally compline with the linter and look a bit more structured.

> For example, the name `iterateFavorites` doesn't tell anything. What does the function do when iterating? How to control the itration?

This is the issue, I mentioned in the source PR's comment: I wasn't always able to understand the logic of the extracted code. I'm open to your naming proposals and code optimizations.

> Even without testing, it's not a good function. It passes a `result` and returns another result, and they are refered to the same instance. The function should create `[]bool` inside and return that value, instead of creating the instance before the function calling.

Thanks, I'll check what I can do to make it better. In general, when I did automatic code extraction by IDE, I then verified manually which variables this function got, which changed, and which had to be reused later in the code after the new function call. So, I added all reusable variables to the return values to make sure that nothing is lost and the package works exactly as before with this extraction being fully transparent.

> And for this function, I think a better way is querying favours status through sql directly, instead of grabbing all raw data and calculating them. It reduces the memory and io

I'm not sure that my skills let me implement such a change. As I mentioned, my main focus during this round was to reformat the code, but not change the logic. 

> My point is, just decreasing the function's cognitive complexity can't generate good code. It may introduce bugs without tests, which are hard to be found during review. For existing code, if lint generates warnings of the complexity, just keep them and improve the code when we change them.

Hmm, does it mean, taking into account all my previous answers in this comment, that I should close these refactoring PRs? If yes, can you please quickly go through them and add a comment to those, you'd agree to merge - there are some of them pretty small with almost no code extraction (like the DB-related one). I'll then update them with the `errors.*` package replacement and try to address your other suggestions while I can.

googollee: I know you've spent a lot of time on this set of PRs. Personally, I hope to continue with these PRs. I hope to analyze these functions one by one. This will take more time than just splitting the functions, and will also delay the progress of other features.

I go deeper into `NewUserFavoriteLoader()`. This function takes `keys` as input, checks if users favor media, and returns `[]bool` with `true` as `favor`/ `false` as not favor in the same order as `keys`. The function `iterateFavorites()` is the core of `NewUserFavoriteLoader()`, so splitting them out is not a good idea. To reduce the complexity, there are two things in my mind:

- Line 16 to 36 are preparing `userIDs` and `mediaIDs`. We could create a function like `uniqueIDs(keys []*models.UserMediaData) (userIDs []int, mediaIDs []int)` to move them separately.
- Line 47 to 54 are a complex logic. We could create a function like `checkFavorite(source []*models.UserMediaData, userID, mediaID int) bool` to give it a better name.

> does it mean, taking into account all my previous answers in this comment, that I should close these refactoring PRs?

To be frank, I don't want to spend too much time on refactoring only, since they are not such high-priority. You can see how I analyze this function and how slow it could be. It's not about what you wrote and I hope to focus on high-priority things. If you agree on this, I'll go refactoring issues and close them if they are just splitting functions but not a better logic.

kkovaletp: > This will take more time than just splitting the functions, and will also delay the progress of other features.

Please don't invest too much time into this refactoring stuff. New features and issues, marked with a "High prio" label and having the status "Before the next release" are the most important things, we have on the list. This refactoring is important in strategic scope, but not the tactical priority. I initiated it, as I thought, it could be easily reviewed as far as I don't change the logic of the code.

> If you agree on this, I'll go refactoring issues and close them if they are just splitting functions but not a better logic.

Sure, just go through those PRs quickly and let me know if you find some of them as simple and easy-to-review, as the DB-refactoring one (which you agreed to merge with some small suggestions). Everything else can be closed and deferred later (feel free to close them as you go).

kkovaletp:
Closing, as it is too complex for review. The code is going to be refactored in parallel with the development inside of these modules

PR #1076, opened 2024-10-03
Refactoring part 2.4: Long func refactoring in `graphql/resolvers`

kkovaletp:
PR is part of the https://github.com/photoview/photoview/pull/1065


api/graphql/resolvers/album.go
```diff
@@ -20,7 +22,9 @@ func (r *queryResolver) MyAlbums(ctx context.Context, order *models.Ordering, pa
 	return actions.MyAlbums(r.DB(ctx), user, order, paginate, onlyRoot, showEmpty, onlyWithFavorites)
 }
 
-func (r *queryResolver) Album(ctx context.Context, id int, tokenCredentials *models.ShareTokenCredentials) (*models.Album, error) {
+func (r *queryResolver) Album(ctx context.Context, id int, tokenCredentials *models.ShareTokenCredentials) (*models.Album,
+	error) {
```
googollee, line 26: This line breaks weirly. Could we keep this in one line?

api/graphql/resolvers/media.go
```diff
@@ -13,7 +13,9 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (r *queryResolver) MyMedia(ctx context.Context, order *models.Ordering, paginate *models.Pagination) ([]*models.Media, error) {
+func (r *queryResolver) MyMedia(ctx context.Context, order *models.Ordering, paginate *models.Pagination) ([]*models.Media,
+	error) {
```
googollee, line 17: Same here.

api/graphql/resolvers/media.go
```diff
@@ -22,7 +24,9 @@ func (r *queryResolver) MyMedia(ctx context.Context, order *models.Ordering, pag
 	return actions.MyMedia(r.DB(ctx), user, order, paginate)
 }
 
-func (r *queryResolver) Media(ctx context.Context, id int, tokenCredentials *models.ShareTokenCredentials) (*models.Media, error) {
+func (r *queryResolver) Media(ctx context.Context, id int, tokenCredentials *models.ShareTokenCredentials) (*models.Media,
+	error) {
```
googollee, line 28: Same here.

api/graphql/resolvers/search.go
```diff
@@ -9,7 +9,9 @@ import (
 	"github.com/photoview/photoview/api/graphql/models"
 )
 
-func (r *Resolver) Search(ctx context.Context, query string, limitMedia *int, limitAlbums *int) (*models.SearchResult, error) {
+func (r *Resolver) Search(ctx context.Context, query string, limitMedia *int, limitAlbums *int) (*models.SearchResult,
+	error) {
```
googollee, line 13: Same here.

api/graphql/resolvers/share_token.go
```diff
@@ -40,7 +40,8 @@ func (r *shareTokenResolver) HasPassword(ctx context.Context, obj *models.ShareT
 	return hasPassword, nil
 }
 
-func (r *queryResolver) ShareToken(ctx context.Context, credentials models.ShareTokenCredentials) (*models.ShareToken, error) {
+func (r *queryResolver) ShareToken(ctx context.Context, credentials models.ShareTokenCredentials) (*models.ShareToken,
+	error) {
```
googollee, line 44: Same here?

api/graphql/resolvers/share_token.go
```diff
@@ -64,7 +65,9 @@ func (r *queryResolver) ShareToken(ctx context.Context, credentials models.Share
 	return &token, nil
 }
 
-func (r *queryResolver) ShareTokenValidatePassword(ctx context.Context, credentials models.ShareTokenCredentials) (bool, error) {
+func (r *queryResolver) ShareTokenValidatePassword(ctx context.Context, credentials models.ShareTokenCredentials) (bool,
+	error) {
```
googollee, line 69: Same here.

api/graphql/resolvers/share_token.go
```diff
@@ -120,7 +127,9 @@ func (r *mutationResolver) DeleteShareToken(ctx context.Context, tokenValue stri
 	return actions.DeleteShareToken(r.DB(ctx), user.ID, tokenValue)
 }
 
-func (r *mutationResolver) ProtectShareToken(ctx context.Context, tokenValue string, password *string) (*models.ShareToken, error) {
+func (r *mutationResolver) ProtectShareToken(ctx context.Context, tokenValue string, password *string) (*models.ShareToken,
+	error) {
```
googollee, line 131: Same here.

api/graphql/resolvers/user.go
```diff
@@ -316,46 +321,61 @@ func (r *mutationResolver) UserRemoveRootAlbum(ctx context.Context, userID int,
...
 	}
 
+	if err := clearCacheAndReloadFaces(deletedAlbumIDs, db); err != nil {
+		return nil, err
+	}
+
+	return &album, nil
+}
+
+func cleanup(tx *gorm.DB, albumID int, deletedAlbumIDs []int, childAlbumIDs []int) ([]int, error) {
```
googollee, line 343: I can't read this function quickly. What's the difference between `albumID` and `deletedAlbumIDs`?

api/graphql/resolvers/user.go
```diff
@@ -316,46 +321,61 @@ func (r *mutationResolver) UserRemoveRootAlbum(ctx context.Context, userID int,
...
+		// Delete albums from database
+		if err := tx.Delete(&models.Album{}, "id IN (?)", deletedAlbumIDs).Error; err != nil {
+			deletedAlbumIDs = nil
+			return nil, err
+		}
+	}
+	return deletedAlbumIDs, nil
+}
+
+func clearCacheAndReloadFaces(deletedAlbumIDs []int, db *gorm.DB) error {
```
googollee, line 363: it's better to put `db` in the first parameter.

api/graphql/resolvers/user.go
kkovaletp: Hmm, maybe I don't understand some aspects of your question, but it looks pretty straightforward:
`albumID int` - passed as a param to the parent function; represents the current album, we're working on; is appended to the array of `deletedAlbumIDs` if `userAlbumCount == 0`
`deletedAlbumIDs []int` - a local var in the parent function, representing an array of deleted albums; passed as a param to this function, where updated and returned back to the parent function; in the scope of this function it can grow, appending the `albumID` if its `userAlbumCount == 0`

BTW, probably line 356 `deletedAlbumIDs = nil` can be safely removed, as it assigns `nil` to the variable, which would be never used. What do you think?

api/graphql/resolvers/album.go
kkovaletp: done

api/graphql/resolvers/media.go
kkovaletp: done

api/graphql/resolvers/search.go
kkovaletp: done

api/graphql/resolvers/share_token.go
kkovaletp: done

api/graphql/resolvers/share_token.go
kkovaletp: done

api/graphql/resolvers/user.go
kkovaletp: done

api/graphql/resolvers/user.go
googollee: `deletedAlbumIDs, err = cleanup(tx, albumID, deletedAlbumIDs, childAlbumIDs)` deletedAlbumIDs is both input and output parameter. It's not a good design.

I guess `deletedAlbumIDs` should be created by this function and returned to the caller.

kkovaletp: You're right, thanks. The `deletedAlbumIDs` is declared but never updated before calling the `cleanup()` in the caller function, so it doesn't make sense to pass an empty var from the caller to the function.
Updated the code.

api/graphql/resolvers/user.go
```diff
@@ -290,7 +295,6 @@ func (r *mutationResolver) UserRemoveRootAlbum(ctx context.Context, userID int,
 	}
 
 	var deletedAlbumIDs []int = nil
```
googollee, line 297: It's better to remove this line, and creates the variable in line 323 as `deletedAlbumIDs, err := cleanup()`.

kkovaletp: The line 323 is inside the transaction. From the code logic I see that if the transaction fails, the function is terminated before the variable is used, however, will your proposal work (be compiled and accepted by linters) and is it ok from coding practice in Go (as I remember from Java that variables, used after code blocks, should be declared before these blocks, not inside them)?
If you confirm that this is ok for Go and will be accepted by the compiler and linters, I'll make this change. I just want to make sure)

googollee: It's OK.

The transaction is not a part of the Go language but from the SQL engine. Only the result of `tx` is passed to the engine. Variables in the callback function are kept inside the process of the service binary.

kkovaletp: Unfortunately, it doesn't work: once I removed the declaration and modified the line 323 as proposed, here is what IDE shows to me
SCR-20241011-iauv

googollee: Ah, OK. I did't notice the using of this variable is outside of the transaction.

googollee (approved):

PR #1088, opened 2024-10-04
Set `$NODE_ENV` according to the image purpose during Docker build

### LINKED ISSUE #1087: Build UI with `$NODE_ENV=production`
Currently, we build UI inside the Dockerfile with no ENV definition for the `npm`, while having Dev dependencies. This means that those Dev dependencies also got installed and might impact or be included in the built files.

That is not what we want, so we have to set `$NODE_ENV=production` at the UI build stage (at the beginning).
However, in this case, it won't be possible to use this stage as a target for building the testing image for CI and local dev env. So we need to find out the way, how to make both workflows work:
- Is it time to create a development Dockerfile?
- Are there some other options?

More info in:
- https://stackabuse.com/the-ultimate-guide-to-configuring-npm/
- https://www.knowledgehut.com/blog/web-development/npm-install-dev-dependencies

kkovaletp:
Fix #1087 

Defines the NODE_ENV variable at the UI build stage in Dockerfile with the default value "production".
Sets the correct value for this variable at the Docker build time in:
- CI build production image --> "production"
- CI build testing image --> "testing"
- `dev-compose.yml` for local development --> "development"

Note: the intentions, added in the `dev-compose.yml` are made by my IDE auto-formatting, as previous formatting wasn't fully compliant with YAML standard (while still working fine for Docker).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -124,6 +124,7 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
           build-args: |
+            NODE_ENV=production
             VERSION=${{ github.ref_name }}
             COMMIT_SHA=${{ github.sha }}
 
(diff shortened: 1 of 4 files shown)


googollee (approved):
~~Do you know the difference between these envs?~~

I see the reason in the issue. Thanks.

PR #1090, opened 2024-10-08
Cleanup deps.

googollee:
Remove deps:

- libexif-dev: It's a old library without changes in years. ImageMagick doesn't use it.
- libwmf-dev: WMF is a Windows Document format. There is no need to support document formats for `photoview`.
  - Also removed runtime dep `libfreetype6` for WMF.
- libaom-dev: It doesn't work as av1 encoder. Instead with `librav1e-dev` for av1 encoder.

### DIFF
--- a/dependencies/build_imagemagick.sh
+++ b/dependencies/build_imagemagick.sh
@@ -6,14 +6,14 @@ set -eu
 
 echo Compiler: ${DEB_HOST_MULTIARCH} Arch: ${DEB_HOST_ARCH}
 
-apt-get install -y libjxl-dev:${DEB_HOST_ARCH} liblcms2-dev:${DEB_HOST_ARCH} liblqr-1-0-dev:${DEB_HOST_ARCH} libdjvulibre-dev:${DEB_HOST_ARCH} libjpeg62-turbo-dev:${DEB_HOST_ARCH} libopenjp2-7-dev:${DEB_HOST_ARCH} libopenexr-dev:${DEB_HOST_ARCH} libpng-dev:${DEB_HOST_ARCH} libtiff-dev:${DEB_HOST_ARCH} libwebp-dev:${DEB_HOST_ARCH} libxml2-dev:${DEB_HOST_ARCH} libfftw3-dev:${DEB_HOST_ARCH} zlib1g-dev:${DEB_HOST_ARCH} liblzma-dev:${DEB_HOST_ARCH} libbz2-dev:${DEB_HOST_ARCH} libexif-dev:${DEB_HOST_ARCH} libwmf-dev:${DEB_HOST_ARCH}
+apt-get install -y libjxl-dev:${DEB_HOST_ARCH} liblcms2-dev:${DEB_HOST_ARCH} liblqr-1-0-dev:${DEB_HOST_ARCH} libdjvulibre-dev:${DEB_HOST_ARCH} libjpeg62-turbo-dev:${DEB_HOST_ARCH} libopenjp2-7-dev:${DEB_HOST_ARCH} libopenexr-dev:${DEB_HOST_ARCH} libpng-dev:${DEB_HOST_ARCH} libtiff-dev:${DEB_HOST_ARCH} libwebp-dev:${DEB_HOST_ARCH} libxml2-dev:${DEB_HOST_ARCH} libfftw3-dev:${DEB_HOST_ARCH} zlib1g-dev:${DEB_HOST_ARCH} liblzma-dev:${DEB_HOST_ARCH} libbz2-dev:${DEB_HOST_ARCH}
 URL=$(curl -s https://api.github.com/repos/ImageMagick/ImageMagick/releases/latest | grep "tarball_url" | cut -d : -f 2,3 | tr -d ' ,"')
 echo download ImageMagick from $URL
 curl -L -o ./magick.tar.gz "$URL"
 
 tar xfv ./magick.tar.gz
 cd ImageMagick-*
-./configure --enable-static --enable-delegate-build --disable-shared --with-x=no --host=${DEB_HOST_MULTIARCH}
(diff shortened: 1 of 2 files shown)


kkovaletp:
> * libwmf-dev: WMF is a Windows Document format.

Actually, this is an alternative to SVG: https://en.wikipedia.org/wiki/Windows_Metafile. I think that we need to support it until we move image processing to the ImageMagick.

> * libaom-dev: It doesn't work as av1 encoder. Instead with `librav1e-dev` for av1 encoder.

Why do we even need the `av1` encoder? We encode all videos using `H264` encoders. Did you mean 'decoder'?

kkovaletp:
A general question: isn't it better to implement the image processing migration to the ImageMagick 1st? In this case, we could drop many more unused dependencies together with not used anymore code.

dependencies/build_imagemagick.sh
```diff
@@ -6,14 +6,14 @@ set -eu
...
+apt-get install -y libjxl-dev:${DEB_HOST_ARCH} liblcms2-dev:${DEB_HOST_ARCH} liblqr-1-0-dev:${DEB_HOST_ARCH} libdjvulibre-dev:${DEB_HOST_ARCH} libjpeg62-turbo-dev:${DEB_HOST_ARCH} libopenjp2-7-dev:${DEB_HOST_ARCH} libopenexr-dev:${DEB_HOST_ARCH} libpng-dev:${DEB_HOST_ARCH} libtiff-dev:${DEB_HOST_ARCH} libwebp-dev:${DEB_HOST_ARCH} libxml2-dev:${DEB_HOST_ARCH} libfftw3-dev:${DEB_HOST_ARCH} zlib1g-dev:${DEB_HOST_ARCH} liblzma-dev:${DEB_HOST_ARCH} libbz2-dev:${DEB_HOST_ARCH}
 URL=$(curl -s https://api.github.com/repos/ImageMagick/ImageMagick/releases/latest | grep "tarball_url" | cut -d : -f 2,3 | tr -d ' ,"')
 echo download ImageMagick from $URL
 curl -L -o ./magick.tar.gz "$URL"
 
 tar xfv ./magick.tar.gz
 cd ImageMagick-*
-./configure --enable-static --enable-delegate-build --disable-shared --with-x=no --host=${DEB_HOST_MULTIARCH}
+./configure --enable-64bit-channel-masks --enable-static --enable-delegate-build --disable-shared --with-x=no --with-magick-plus-plus=no --host=${DEB_HOST_MULTIARCH}
```
kkovaletp, line 16: Could you please explain these parameters and their impact? Did you review all possible parameters and set the optimal ones?

googollee:
> Actually, this is an alternative to SVG: https://en.wikipedia.org/wiki/Windows_Metafile. I think that we need to support it until we move image processing to the ImageMagick.

I don't think SVG files belong to photos. I don't want to support a platform-special format. I removed it because it has a lot of dependencies, like x11 and FreeType font.

It's an ImageMagick dependency. If we want to process it with ImageMagick, we should keep it. But, as I said above, I don't want to support a platform-special format.

> Why do we even need the av1 encoder? We encode all videos using H264 encoders. Did you mean 'decoder'?

`decoder` is `libdav1d`. If we don't want to convert files to heif, we could remove all `encoder` dependencies. Do you agree on that?

> A general question: isn't it better to implement the image processing migration to the ImageMagick 1st? In this case, we could drop many more unused dependencies together with not used anymore code.

We can't drop any here. ImageMagick delegates some formats to different dependencies. Here are all those dependencies. I hope to make a clear dependency before moving everything to ImageMagick.

dependencies/build_imagemagick.sh
googollee: It enables ImageMagcik working with more channels, see https://imagemagick.org/script/multispectral-imagery.php.

kkovaletp:
> I don't want to support a platform-special format.

Well, there are pros and cons of removing this format support from ImageMagick:
Pros:
- drop many dependencies, which makes the image smaller, and more secure (less deps > less chances to get vulns from them)
- the format is really old and rarely used (mostly by legacy software)
Cons:
- this format is still a picture format and our product is used not only as a photo gallery but also as a picture gallery (I know that from submitted issues and discussions in Discord)

From my personal perspective, chances are low that someone would want to use this format in practice. So, I'm fine with the removal, however, I'd like to document somewhere, what a user should do to add this format support back (of course, it is possible for a direct deployment scenario). I think that a new Discussion is the best place for this, as a readme is overloaded and this case is going to be used in rare setups.

> If we don't want to convert files to heif, we could remove all `encoder` dependencies.

Please double-check it from the code perspective, but as far as I know, we convert images to JPG only and videos to MP4 with H264 (different encoders) and AAC. If that is true, then we don't need other encoders, as they have never been used. If in the future we'd like to add/switch to another format, we'll add the needed encoders back.

dependencies/build_imagemagick.sh
kkovaletp: Thanks. What about other parameters? Are there some other useful ones, which we could add or modify to optimize and enhance ImageMagick for our use case?

googollee: I didn't find other parameters relating to image format. I believe this 64bit channel masks is not a popular feature, but it doesn't increase the binary size.

googollee:
> our product is used not only as a photo gallery but also as a picture gallery

Even as a picture gallery, WMF is not a popular format. It's easy to add it back. I'd like to see how it works without WMF support first. If necessary, we could add the support later.

> we convert images to JPG only and videos to MP4 with H264 (different encoders) and AAC

Agree. I'll remove encoders from `libheif` and update the pr. `ImageMagick` doesn't split encoders and decoders with packages, so it can't do the similar thing.

kkovaletp:
> I'd like to see how it works without WMF

I've searched for WMF in the `scanner/media_type/media_type.go` and it is not there, meaning that we don't support it now. With this, it is safe to drop its support from ImageMagick, as there is no degradation from the Photoview side.
Also, there is no need to document anything about this format, as we never supported it.

Let's wait for feature requests for that format and decide later based on real use cases.

kkovaletp (approved):
LGTM

PR #1091, opened 2024-10-08
Update dep

googollee:
Follow up with https://github.com/photoview/photoview/pull/1090


kkovaletp:
I didn't get why do we need this PR to be split from the #1090?

googollee:
> I didn't get why do we need this PR to be split from the https://github.com/photoview/photoview/pull/1090?

It changes the build dependencies of `libheif`. If `dependencies` image doesn't update, `photoview` image won't pass the build. That's why the test and build fail now.

kkovaletp (approved):
Approving, but please merge it in a sequence with #1090, once it is approved too

PR #1092, opened 2024-10-08
Refactoring part 3: get the `github.com/pkg/errors` out from the code (and a few more improvements)

kkovaletp:
This PR affects many source code files, but most of the changes are simple and similar, so I hope this wouldn't be a problem. The changes are:
- replace "errors" and  "github.com/pkg/errors" with "fmt"; 
- replace deprecated "github.com/99designs/gqlgen/handler" with "github.com/99designs/gqlgen/graphql/handler/transport" in `api/graphql/auth/auth.go`; 
- replace deprecated `strings.Title()` with `cases.Title().String()` from "golang.org/x/text/cases" in `api/graphql/resolvers/media.go`


kkovaletp:
@googollee, while working on this PR, I found the next issue in my IDE:
- package `utils` all in red because of the error: 'case-insensitive file name collision: "Throttle.go" and "throttle.go"'
- the `utils/trottle.go` also shows the next errors:
- line 5 `type Throttle struct {`: 'Throttle redeclared in this block (see details)'
- line 10 `func NewThrottle(interval time.Duration) Throttle {`: 'NewThrottle redeclared in this block (see details)'

It looks like a naming issue, however, as the same name is used for different objects, I'm not sure that the built-in IDE's rename functionality will correctly rename them to solve the problem. The fact that the type and its function are capitalized (public), doesn't make things simpler.
The naming would be another pain point, as I'm not sure that I understand all the cases and purposes, this code serves.

Could you please take a look and advise on how to fix this issue? Which other code files are affected?

googollee:
I didn't want to refactor so many files at the same time. I will do it package by package to limit the potential error.

IIRC, I renamed `Throttle.go` to `throttle.go` because using lower cases in file names is common. So I guess when you rebase your branch to the latest master, you keep both files by mistake. It'd be easy to find if the change is limited to `api/utils` only, or not a problem if the change doesn't touch this package.

kkovaletp:
This PR was part of a series of my refactoring PRs. We decided to close it, as you'd like to move to the `fmt` package-by-package at the time, you touch that code.
I just was busy with other things and got to this PR only this week, extracted a smaller part of refactoring to a separate PR, and after making sure that nothing else is lost here, I closed this one, as was discussed.

PR #1096, opened 2024-10-14
Process jpgs if raw processing is disabled.

googollee:
Fix https://github.com/photoview/photoview/issues/677.

Sure.

- Assuming there is only file: `raw.raf`.

If `PHOTOVIEW_DISABLE_RAW_PROCESSING` is set (`1` or `true`), `photoview` will not process `raw.raf`. No media item in this case.

If `PHOTOVIEW_DISABLE_RAW_PROCESSING` is not set (other values), `photoview` will process `raw.raf`. The original media item is `raw.raf` and stacks with thumbnails generated from `raw.raf`.

- Assuming there are files: `raw.raf` and `raw.jpg`.

If `PHOTOVIEW_DISABLE_RAW_PROCESSING` is set (`1` or `true`), `photoview` will not process `raw.raf` and handle `raw.jpg` as a standalone photo. The original media item is `raw.jpg` and stacks with thumbnails generated from `raw.jpg`.

If `PHOTOVIEW_DISABLE_RAW_PROCESSING` is not set (other values), `photoview` will process `raw.raf` and treat `raw.jpg` as a conter-part file of `raw.raf`. The original media item is `raw.raf` and stacks with `raw.jpg` and thumbnails generated from `raw.jpg`.

- If Changing `PHOTOVIEW_DISABLE_RAW_PROCESSING` from `1` -> `0` or `0` -> `1`.

After a new scanning cycle, there will be only one item with the right stacked files.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches: [master]
   push:
-    branches: ['*']
+    branches: [master]
     tags:
       - v*
   schedule:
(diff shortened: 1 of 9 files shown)


kkovaletp (commented):
The change is small and straightforward. I like it.

As you touch this package/file, can you please also add unit tests?
For the package - is better, but at least for the file (it should be easy as the file contains several straightforward functions)

googollee:
Yes, I'll add unit tests. I hope to confirm this direction first. Working on unit tests.

googollee:
@kkovaletp I finished the test. `codecov` still complains about the unreachable codes. Those codes are in `test_utils` and are hard to produce. So they could be ignored.

kkovaletp:
Looks good. Please revert the testing code and I'll approve

googollee:
Done. Also move all test_data together.

kkovaletp (approved):
LGTM

PR #990, opened 2024-07-06
add option to download all pictures as zip from share

derqurps:
This PR adds the ability to show the sidebar in the album view, so that the whole album can be downloaded (for anonymous share users).

This should close https://github.com/photoview/photoview/issues/613

In [AlbumSidebar.tsx](https://github.com/photoview/photoview/compare/master...derqurps:photoview:share_download_pictures?expand=1#diff-4e4ae8cd9a347342034080292fc5fdd3a7f6b65a18a4e172dc87b365776b3330), i removed the graphql request, as it was used only for fetching the album title. And the title is available in the calling file [AlbumTitle.tsx](ui/src/components/album/AlbumTitle.tsx), so i passed it along from there.

PR #1097, opened 2024-10-16
Add ability to allow share albums to be downloaded

snowe2010:
Shares are currently not downloadable in full, you must download each image individually. This commit adds the ability to set a share as
downloadable, and it adds a download button to the share.

This commit does not include the generated schemas as it seems like they were generating incorrectly. They will be in the next commit. Shares currently are not able to be downloaded in full, unless you are logged in. This PR adds the ability to set a Share to allow downloading the full album rather than downloading each image individually. 

Some restrictions. I couldn't figure out a way to only show the download button if the share was allowed to be downloaded. This is due to the way that the share tokens work on the frontend. So trying to download if you do not have permission gives an error.

- **Add ability to allow share albums to be downloaded**
- **Generated code for allowing shared album downloads**

# Demo

updated sidebar:
image
image


kkovaletp:
@snowe2010, thanks for your contribution, but it looks like this PR is a duplicate of the https://github.com/photoview/photoview/pull/990 right?
Besides that, what is the reason for allowing users to download shared content? By default, if content is loaded and shown on the page, it can be saved from that page using browser functionality, so, if you share an album and don't allow it to be downloaded, actually users are still able to download it, but using different (not so convenient) way.

snowe2010:
> That's why it is always better to discuss your contribution plans in advance in the issue or in our Discord ;-)

haha I messaged and waited a day! I needed the feature pronto though so I just went for it. 

> That makes sense. This is the reason, why this feature is valuable in the product. But my question was more about the "Allow downloading" checkbox: what is the use case, when it might be useful to share an album, but doesn't allow a bulk download function?

oh no clue. The software didn't allow the download of the full album already so I thought that was a desired feature. If you are ok with just making it available no matter what I don't think there's a problem with that. I mean most people probably wouldn't complain, but less code to maintain is never a bad thing.

kkovaletp:
Did you have a chance to compare your PR with #990?
What do you think?

snowe2010:
I like the other PR more, if you're fine with not having the checkbox. It will work more consistently and doesn't require db updates and stuff. One thing I want the other PR to change is that the cog should just be a download icon on the share page, like I did in this PR. Else users are going to have no clue they can even download (I already pulled the pr, built it for my server and deployed it and have people downloading files from that and I had to explain the button lol). https://github.com/snowe2010/photoview/blob/allow-share-album-download/ui/src/components/album/icons/download.svg https://github.com/photoview/photoview/pull/1097/files#diff-dfb80e14435167147e2bff2366d0e2b67a4e729cba19b6bd92128ba58a4d2cd0R136

kkovaletp:
Great! Thanks for letting us know!
I agree with the button location improvement. It would be good if you could create another PR for the button improvement only. I guess that it should depend on the #990, right? If so, you can prepare a draft PR with the button and later rebase it on `master` once the #990 is merged (hope that it will happen soon).

With this, I'm closing this PR as not planned.

snowe2010:
@kkovaletp I think the AlbumTitle is shared by both the share page and the album page, so it would take some update. Probably best to wait for #990 to be merged.

Issue #942, opened 2024-04-30
Regenerate generated files in CI workflow and notify if the code is changed


jordy2254:
There's a lot of generated files in version control both from the api (gql gen) and ui (apollo client) Both of these cause several issues which could be alievated from removing them from tracking
- Bloating of the code base and reduced searchability via the basic tools available on github
- High likley hood/guarenteed conflicts when users are working on api changes or similar front end features

To mitigate this utilising husky (Which we already have) And implementing a hook to generate the generated files would provide some benefit. Development flow, docker builds & tests would all have to change a little to factor in this change but overall I believe this would be beneficial to reduce the number of conflicts as work progresses. 
@viktorstrate Can you think of any reason this shouldn't be done? 

UPD by @kkovaletp: **Conclusion**:
https://github.com/photoview/photoview/issues/942#issuecomment-2423700495

viktorstrate:
I think one reason for committing generated files is simplicity.

I haven't really had merge conflicts because of them, since they only change when the Graphql API changes.

Not having to worry about generating them is another reason. Eg. if I checkout another branch will just work.

Last reason is that the generated files _are_ source code ie. Go and Typescript, and actually even quite readable. I see it as a plus to be able to read them and search in them. I often look at the generated Typescript types to understand the code. I actually think GitHub can use them to analyse the code and give type hints.

But I also see that there are benefits for removing them.

jordy2254:
@viktorstrate  Sounds like you find it useful then, I haven't personally looked at the generated source on github at all only using it locally to check some interfaces. However, if you find it useful and feel it should stay we can close this issue. I personally have already experienced merge conflicts on the 2 issues I've done but that's because they both touch the api. We could do the below if we keep the generated files in vcs, so I think this issue can just change scope rather than content.

- Ensure generated code is committed in a separate commit (This is only to make the rebuild process easy)
- Run code generation as part of the code build to ensure all generated code is correct

Let me know your thoughts for going forward.

kkovaletp:
> * Run code generation as part of the code build to ensure all generated code is correct

[...]

PR #1099, opened 2024-10-17
Fix gqlgen code

### LINKED ISSUE #1098: Fix `gqlgen`.
Currently, running `go generate` aka `gqlgen` returns a lot of errors. It blocks `photoview` to regenerate codes with the latest `gqlgen`, as well as add new APIs.

We need to fix it.

googollee:
Fix #1098.

- Split `schema.graphql` into small pieces according to the go file.
- Rerun `go generate ./...` under `api/` to generate go files with the latest `gqlgen`.
  - Most of go files has the same content, just the order of functions are changed.
  - `*.utils.go` contains functions/structs that can't be generated from `*.graphql` files.
- Also removed old `errors` library in those go files.


kkovaletp:
This PR is so huge and complex (although, GitHub shows most of the changes as complete removal and implementation), so I cannot review it from the code perspective.
So, I'd ask you to run a complete manual regression testing for the product (as all the APIs seem to be affected by the change) to make sure that everything works as before and all API endpoints are ok.

googollee:
I agree that this is huge. I renamed generated files to original file names. It's helpful to review them with diff now. @kkovaletp let me know if it helps reviewing. If it's not enough, I can split PR to each graphql files one by one.

I run the local test instance, but it's hard to say I can cover all API endpoints.

kkovaletp:
> I renamed generated files to original file names.

Thanks. It might help, but taking into account amount of changes and my limited experience in Go, it doesn't help much...

> I can split PR to each graphql files one by one.

I still can review them file-by-file in the scope of this PR, so I don't see a benefit in splitting - the same amount of work (more than I can do during reasonable time)

> For two `generated.go`, all content are generated. Even they break lint, I suggest to keep them

Yes, I agree that all the code has to be kept in the repo. There was a discussion (and there is issue #942 about that), but I think that all the code should be available for traceability and easier investigation.
BTW, when does this code have to be regenerated? Does it make sense to add pre-commit hooks or CI steps for that to make sure the code is always in sync with the generated parts?

> I run the local test instance, but it's hard to say I can cover all API endpoints.

When we had similar cases in my previous project on the work, we solved them by combining at least 2 of the next solutions:
- Good code coverage by automated tests (unit, integration, system levels)
- Code review online session: a webinar (or a sequence of webinars) with the team, where the author goes over the code and explains what exactly it does and why it is implemented in that way, answering questions from the team members
- Deep manual testing of the impacted functionality, including corner and negative cases

However, I don't think that the 2nd option is suitable for our project. It would be great if you could cover changed files by unit tests with a good coverage level, but I understand that this is a lot of extra work. So, the only option left is to run the code and try to call all the affected APIs in all possible cases from the product frontend to make sure that it still works as expected. I know that it is also a lot of work, but less than covering it with good-quality unit tests, achieving good coverage (I think).
Sure, the best would be to combine both unit and manual tests)

googollee:
> Does it make sense to add pre-commit hooks or CI steps for that to make sure the code is always in sync with the generated parts?

Yes. I plan to submit a script to run that and returning non-zero if files are updated.

I also updated the description of what this PR does.

I ran local tests as:

- Clean all database
- Start a new instance
- Go through start wizard
  - Media path has some jpeg, some raw standalone and some raw with jpeg.
- Run scanning
- Check album
- Check faces
- Check map

It works well with that test.

kkovaletp:
> I plan to submit a script to run that and returning non-zero if files are updated.

Great. Then we could add it to the 'Code analysis' workflow.
I, personally, don't like git hooks, as they require local setup, don't always work, and make commit/push actions much longer. So, I'd prefer to avoid them as long as possible)

> I ran local tests

Great starting point, but I see the next gaps:
- Please add a video to your testing gallery
- Check Timeline:
  - check pagination
  - check Date dropdown
  - check the media info side panel for images and videos with and without EXIF, location, people
  - set as an album cover
  - share the media
  - mark as favorite
  - show only favorites
  - open an image and a video (and play it) in the viewer, navigate back and forth in the viewer
- Open previously shared link in an incognito window
- Check Albums
  - I don't know what you checked, so list the functionality (here and for the other sections):
  - Sub-albums
  - Pagination
  - Viewer
  - Sorting
  - Favorites
  - Settings (sharing, cover, downloads)
- Map
  - Show all images with the location
  - Regroup images based on zoom
  - Viewer
- People
  - List all recognized face groups
  - Recognize unlabeled faces
  - Inside a group: viewer, side panel, favorites, 4 buttons to manage the group
- Settings
  - Create a non-admin user with a different path and re-login as that user
  - Change language and jump over the pages
  - Set a theme and jump over the pages
  - Run a scan for all and per user manually
  - Change scanner options and find them in the DB
  - Set different thumb generation method and rescan with a clear cache before that 
- Search

googollee:
After testing with the image for a while, I went through all of the functions in https://github.com/photoview/photoview/pull/1099#issuecomment-2421539043. All are working well.

googollee:
`ghcr.io/googollee/photoview:fix-gqlgen` is the image I'm using. You can test it also.

kkovaletp:
Cool) I'll test it tomorrow.
But I trust your results as well)

kkovaletp:
I tested several cases and didn't find anything except of the #1103 (I double-checked it with the `master` image and it is reproduced there as well, so it is not introduced here)

So, I'm ready to approve the PR once you revert the testing changes.

googollee:
Reverted.

kkovaletp (approved):
Approve based on manual testing

kkovaletp:
> I plan to submit a script to run that and returning non-zero if files are updated.

@googollee, do you want me to create an issue for that (to not forget about it)?

googollee:
> do you want me to create an issue for that (to not forget about it)?

I'm working on it. But I found this PR missing a lot of commits and it's not fully fixed `gqlgen`. I need to fix this first and create a new PR.

kkovaletp:
Ok, sure. Then waiting for it

googollee:
I made mistakes when rolling back workflows. It rolled back more commits instead of cherry-picking the workflow commit.

kkovaletp:
That's why I prefer to not use such advanced git features if there is an option to achieve the same result simpler (for example, by making 1 more commit)

googollee:
It did create a new commit with the Cherrypick.

In the other hand, I'd like to (re)open the discussion about enabling workflows for forked repos. It's better to avoid those changes when working a PR.

kkovaletp:
> I'd like to (re)open the discussion about enabling workflows for forked repos

Ok, please describe what exactly you would like to change, taking into account that there are different purposes for forking a repo

googollee:
We can discuss based on PR #1105.

PR #1101, opened 2024-10-19
Add the ability to re-detect faces for selected photo

### LINKED ISSUE #1071: Add the ability to re-detect all face for all image
As far as I know, detection only occurs during image processing, it would be nice to be able to do it whenever you want.
What is needed:

- clear face_group table
- clear image_faces table
- re-detect all image using already implemented functionality
- add button to front-end (i think it would be find add it near SCAN in settings)

This can help with the error in case of failed detection (see #1063)

vsakhnenko:
https://github.com/user-attachments/assets/01d88ca5-c7f7-49fa-8fd1-8cc40700f27f


kkovaletp:
@vsakhnenko, thanks for the demo video. So far it looks good. Let's wait for the code review from @googollee on the backend part.
But before that, can you please resolve the conflict with GraphQL?

PR #1108, opened 2024-10-26
Push `dependencies` image to the new registry

kkovaletp:
This is the 1st of 2 PRs to migrate to the new Docker registry, where we have more control over the images and are able to manage them.
The new registry is https://hub.docker.com/u/photoview

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -10,10 +10,11 @@ on:
     - cron: '18 23 * * 3'
 
 env:
-  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
+  #IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
+  IS_PUSHING_IMAGES: true
   IS_CACHING: ${{ github.event_name == 'pull_request' }}
   DOCKER_USERNAME: viktorstrate
-  DOCKER_IMAGE: viktorstrate/dependencies
+  DOCKER_IMAGE: photoview/dependencies
   DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
   PLATFORMS: linux/amd64,linux/arm64,linux/arm/v7
 
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -38,8 +38,10 @@ jobs:
         if: ${{ env.IS_PUSHING_IMAGES == 'true' }}
         uses: docker/login-action@v3
         with:
-          username: ${{ env.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_PASSWORD }}
+          username: kkoval
+          password: dckr_pat_cqK5D6f9c2FylKI0Xhcp9LKoWlE
+          #username: ${{ env.DOCKER_USERNAME }}
+          #password: ${{ secrets.DOCKER_PASSWORD }}
 
       - name: Docker meta
         id: docker_meta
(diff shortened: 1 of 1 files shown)


.github/workflows/dependencies.yml
```diff
@@ -10,11 +10,10 @@ on:
...
 
 env:
-  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
+  #IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
+  IS_PUSHING_IMAGES: true
   IS_CACHING: ${{ github.event_name == 'pull_request' }}
-  DOCKER_USERNAME: viktorstrate
-  DOCKER_IMAGE: viktorstrate/dependencies
-  DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+  DOCKER_IMAGE: ghcr.io/photoview/dependencies
```
googollee, line 16: I guess it should be `ghcr.io/photoview/photoview/dependencies`, as it's a image of `photoview/photoview` repo.

googollee (approved):
We can try this workflow first. If it fails, we could update the image name.

.github/workflows/dependencies.yml
kkovaletp: I've tried this change: https://github.com/photoview/photoview/actions/runs/11559527551/job/32174376849?pr=1108#step:7:23138
But it also doesn't work, as this is not a regular repo, but part of the `photoview` org. GitHub registry for images and packages is on the org level, not on the repo level for org repos, so the initial address was correct.
Anyway, I wanted to double-check it

googollee: Did you grant the writing permission to workflows in Settings? 
image

kkovaletp: I don't have admin permissions in this org(
@viktorstrate, this question to you

googollee: I'd like to try the simple one first: `ghcr.io/photoview/photoview:dependencies` or even `ghcr.io/photoview/photoview`. If all settings are right, these should work well. After that, we can check how a sub-image works.

kkovaletp:
```
Error: buildx failed with: ERROR: failed to solve: failed to push ghcr.io/photoview/photoview/photoview:latest: unexpected status from POST request to https://ghcr.io/v2/photoview/photoview/photoview/blobs/uploads/: 403 Forbidden
```
So, basically, we cannot use the GHCR of the Photoview org.

googollee: I see you change the image name to `ghcr.io/photoview/photoview/photoview`. Is that right? I think it should be 'ghcr.io/photoview/photoview`.

kkovaletp: I tried that earlier with the `dependencies` image. If GHCR accepts login and allows pushing, it should be a different error message

kkovaletp: I've successfully pushed the image to the Docker hub org registry with my temporary PAT used. I've asked @viktorstrate in the Discord to double-check what is inside the current secret (PAT or password), as only the PAT is fully-supported.

viktorstrate (approved):
Let's try and see if this will work

kkovaletp:
@googollee, @viktorstrate, after today's investigation I've found that the reason, why the secret doesn't work, is that this is an intentional implementation of GitHub Actions to prevent stealing secrets by submitting malicious PRs.
https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/
Long story short:
> Due to the dangers inherent to automatic processing of PRs, GitHub’s standard pull_request workflow trigger by default prevents write permissions and secrets access to the target repository.

So, there is only the way to review the PR and merge it with the hope that it will work in the `master` branch.

googollee (approved):

PR #1109, opened 2024-10-26
Switch all images to the new registry

kkovaletp:
This is the 2nd of 2 PRs to migrate to the new Docker registry, where we have more control over the images and are able to manage them.
The new registry is https://hub.docker.com/u/photoview

**Make sure** that the https://github.com/photoview/photoview/pull/1108 is merged and corresponding CI jobs successfully finished before merging this PR, as it expected the `dependencies` image to be available in the new registry.


Dockerfile
```diff
@@ -64,7 +64,7 @@ RUN chmod +x /app/scripts/*.sh \
   && /app/scripts/install_build_dependencies.sh \
   && /app/scripts/install_runtime_dependencies.sh
 
-COPY --from=viktorstrate/dependencies /artifacts.tar.gz /dependencies/
+COPY --from=photoview/dependencies /artifacts.tar.gz /dependencies/
```
googollee, line 67: It should be `ghcr.io/photoview/dependencies`. (or `ghcr.io/photoview/photoview/dependencies` depending on how the workflow works)

kkovaletp: The `dependencies` imege is always pushed to the new Docker org registry, so there is no need to put a condition here. Anyway, it should be fine to refer to the default location here even if it is a PR or local branch in a forked repo.

kkovaletp:
@googollee, @viktorstrate, please review the 2nd part of the 'switch to the new registry' PR. It also fixes the issue in the `master` branch with incorrect condition resolving.

viktorstrate (approved):

googollee (approved):

PR #1113, opened 2024-11-02
Ignore all generated go files in codecov.yml

viktorstrate:


kkovaletp:
@googollee, please review this small improvement to make sure it is in sync with the #1105 and nothing is missing.

codecov.yml
```diff
@@ -1,3 +1,4 @@
 ignore:
+  - "api/**/*.generated.go"
```
googollee, line 2: The file `api/graphql/resolvers/*.go` (except `api/graphqlresolvers/*.util.go`) are not fully generated go files. `gqlgen` generates functions without implementation, see [the example](https://github.com/photoview/photoview/blob/master/api/graphql/resolvers/album.go#L67) (This is also not implemented in [the original file](https://github.com/photoview/photoview/blob/c151bdee963c97c6f0c7561eda32974845625bb5/api/graphql/resolvers/album.go#L124)). We added implementation. `gqlgen` will move the implementation to the new file when regenerating them.

I still think we should cover those files, especially they are manipulating data. I'd like to move those db methods to a database layer, but it's not an easy work.

kkovaletp:
@viktorstrate, @googollee, I see that this PR has been stuck for a week after the last review comment.
@viktorstrate, what would be your decision on it? Are you ok to close it or do you have an alternative vision and would like to continue the discussion?

viktorstrate:
I think it's fine to close it. I was just saw an old commit where some generated files were not ignored. But it seems like this will not be a problem after #1105.

PR #1116, opened 2024-11-08
Fix scheduled docker build for the latest tagged commit

kkovaletp:
Now scheduled runs for the Build workflow run only for the latest commit in the `master`. This fix is to make it catch the latest tag and run also for it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-tags: 'true'
+          fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


kkovaletp:
Here is the proof that it works: https://github.com/photoview/photoview/actions/runs/11742486064/job/32713296118?pr=1116#step:3:1
I need to remove this debug data, so the final commit doesn't have any differences in the workflow run in the scope of the PR

kkovaletp:
@googollee, please review this small PR to make the scheduled runs to rebuild the image for the latest tag.

PR #1119, opened 2024-11-11
Update Language Label for Polish in Language Preferences

DontAskForAnything:
This PR updates the language label for Polish from `"polski"` to `"Polski"` in the `languagePreferences` array.

### Reason for Change
In order to maintain consistency with other language labels in the app, which all use capitalization (e.g., `"English"`, `"Deutsch"`, `"Français"`), the Polish label was updated to `"Polski"`. Using an uppercase first letter aligns with UI conventions and enhances readability. This minor change provides a more polished and professional appearance for users selecting languages.

### DIFF
--- a/ui/src/Pages/SettingsPage/UserPreferences.tsx
+++ b/ui/src/Pages/SettingsPage/UserPreferences.tsx
@@ -25,7 +25,7 @@ const languagePreferences = [
   { key: 3, label: 'Svenska', flag: 'se', value: LanguageTranslation.Swedish },
   { key: 4, label: 'Dansk', flag: 'dk', value: LanguageTranslation.Danish },
   { key: 5, label: 'Español', flag: 'es', value: LanguageTranslation.Spanish },
-  { key: 6, label: 'polski', flag: 'pl', value: LanguageTranslation.Polish },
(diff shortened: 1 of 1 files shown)


kkovaletp:
@googollee, can you please take a look at the failed Dependencies image build job?
It fails on the `RUN export $(cat /env) && ./download_jellyfin-ffmpeg.sh`
And here is the only output, it shows: `Compiler: x86_64-linux-gnu Arch: amd64`

Was something changed on the jellyfin-ffmpeg repo level?

googollee:
@kkovaletp I re-ran the build and it succeeded.

I guess the failure is caused by GitHub blocking the download. I saw that sometimes before and re-ran the command working well. Docker Build doesn't log the real command generating the error, but the one before. So the error command is just after the one producing `Compiler: x86_64-linux-gnu Arch: amd64`, see [the code](https://github.com/photoview/photoview/blob/3cf63e7e2b7fac271d5ebd547d16db165dd853aa/dependencies/download_jellyfin-ffmpeg.sh#L9).

kkovaletp:
Thanks, @googollee.
I rerun it 3 times and it was failing. It looks like it was fixed on the upstream repo side...

kkovaletp (approved):

PR #1125, opened 2024-11-17
Fix imagemagick doesn't have any dep when cross-compiling (arm64, arm/v*).

googollee:
Also add openh264 for libheif avc decoder.


kkovaletp (approved):
LGTM

PR #1128, opened 2024-11-19
Refactor `auth.go` and `media.go` to remove deprecated dependencies

kkovaletp:
This change is extracted from the PR #1092 before closing it.

@googollee, I have a question regarding the changes in the `media.go`: it has a comment that it was auto-generated, so making changes in the code doesn't make sense, as it will be overwritten the next time, the code is regenerated. Is there a way to instruct the `gqlgen` to generate the code in this way, including the fixes?
I committed the changes in the `media.go` to show the desired result of `gqlgen` run. If there is no way to implement this fix to be replicated by `gqlgen`, I'll revert this file and keep only the changes in the `auth.go`.


googollee:
See #1129. Currently test workflow doesn't fail when test is failing. I fixed it in this PR.

googollee:
> I have a question regarding the changes in the media.go: it has a comment that it was auto-generated, so making changes in the code doesn't make sense

`gqlgen` moves the implementation of functions defined in the schema when it generates new codes. So it makes sense to update the code in it. Be aware that it doesn't move any functions not defined in the schema, like util functions. So put those functions in `xxx.util.go` if necessary.

kkovaletp:
> `gqlgen` moves the implementation of functions defined in the schema when it generates new codes. So it makes sense to update the code in it

Do you mean the `graphql/resolvers/media.graphql`? I see only type definitions and no code. Where should I make these changes exactly?

googollee:
For example, in schema, it defines [`Query.myMedia(order: Ordering, paginate: Pagination): [Media!]! @isAuthorized`](https://github.com/photoview/photoview/blob/772129dbeb234eabce3d8603132876a45fb209a4/api/graphql/resolvers/media.graphql#L107), which generates the function [`func (r *queryResolver) MyMedia(ctx context.Context, order *models.Ordering, paginate *models.Pagination) ([]*models.Media, error)`](https://github.com/photoview/photoview/blob/772129dbeb234eabce3d8603132876a45fb209a4/api/graphql/resolvers/media.go#L160) in `media.go`.

If that function doesn't exist in `media.go` file (or the file doesn't exist), `gqlgen` generates an unimplemented function (one example is [here](https://github.com/photoview/photoview/blob/772129dbeb234eabce3d8603132876a45fb209a4/api/graphql/resolvers/album.go#L67)) in new `media.go`. If the function already exists in `media.go`, it copies the implementation to the new generated `media.go`.

So if changes are inside `MyMedia()` function, keep them in `media.go`. If split new functions from `MyMedia` like previous refactor to reduce the complex, put new functions in `media.util.go` and update `MyMedia()` to use them.

kkovaletp:
Great, thanks for the explanation! Now I think I understand the concept.
So, actually, it doesn't generate the code, but only an empty code structure, which then has to be populated with the actual code, doing the job.
I thought the generator was smarter than it is))

kkovaletp:
@googollee, I set this PR as ready. please review it once you have time

PR #1129, opened 2024-11-19
Remove `continue-on-error: true` because it hides errors.

googollee:
I added an always-failing test and it doesn't break the test workflow, see [here](https://github.com/googollee/photoview/actions/runs/11916842602/job/33210448452#step:6:427).

After some investing, it's because of `continue-on-error: true`. Removing that and rerunning the test breaks the workflow, see [here](https://github.com/googollee/photoview/actions/runs/11916984621/job/33210936450#step:6:427).

So, we should remove `continue-on-error: true` to catch errors with testing workflow.

Also fix all failing tests.


kkovaletp (approved):
The main idea looks good to me, however, in this case, we don't have coverage info if the test run fails. It is not a big problem, as fixing failed tests has a higher prio over analysing test coverage, but I still wanted to mention this side effect.

I see that you regenerated the auto-generated code, but the `media.go` still contains the same code, including deprecated calls (only the comment was changed). What should I notice here in the scope of the #1128 PR?

googollee:
> we don't have coverage info if the test run fails.

It's fine. If the test fails, it may not run all tests and the coverage is unusable. It should always pass tests first before checking the coverage.

> I see that you regenerated the auto-generated code.

I just regenerated codes to fix the testing. Let's do the other thing with https://github.com/photoview/photoview/pull/1128.

PR #1130, opened 2024-11-20
Add `log` package for better logging level.

googollee:
It's splitted from #1124.

We should use `log` package to do better logging (level, context). There is not much logic in `log` now. In the future, we could add more functions to this package.

### DIFF
--- a/api/log/default.go
+++ b/api/log/default.go
@@ -1,0 +1,57 @@
+package log
+
+import (
+	"context"
+	"log/slog"
+)
+
+var defaultLogger *slog.Logger
+
+func init() {
+	defaultLogger = slog.Default()
+}
+
+// Debug calls [Logger.Debug] on the default logger.
+func Debug(msg string, args ...any) {
+	defaultLogger.Debug(msg, args...)
+}
+
+// DebugContext calls [Logger.DebugContext] on the default logger.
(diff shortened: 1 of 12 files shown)


viktorstrate (approved):

viktorstrate:
Awesome, looks good!

kkovaletp:
@googollee, great improvement!
It would be good to have an env vars to: 
- control the log level 
- define the log output stream destination (STDOUT, file, a log collection service, etc.)

Another thing (more documentation) is how to use and when to call which log methods. I guess, it should be fine to add it as comments to the corresponding methods.

googollee:
> It would be good to have an env vars to some features.

I'm not decided to do it through ENV vars, or introduce the config file. I see `photoview` has more and more ENV vars and it's hard to organize them in a structural way. So I think it's time to have a config file.

> Another thing (more documentation) is how to use and when to call which log methods.

I'm also thinking some other features of logging, like span logger. I'll do document things at that time.

kkovaletp:
I like the idea of having a config file with better readability and visual structure than the current `.env` file.
What do you think about YAML format? We already use it for docker-compose files, so we won't introduce any new complexity to users. Also, YAML is powerful and visually organized in a good and clear way.

PR #1131, opened 2024-11-21
Include a group of small but real media files for testing.

googollee:
Split from #1124.

Some tests scans all files under `api/scanner/test_data`. If I put more image/video files under test_data, it will break tests as changing the number of image files.

I plan to include a group of [small but real media files](https://github.com/photoview/photoview/pull/1131/files#diff-9b606754754bb4d38f97fdbc6171911023e27ecb99413d85d72f7d2293d26ece) for testing. Since those files are used by different packages (at least by `media_type` and `scanner_tasks`), `api/scanner/test_data` is the only reasonable place to put those media files. To avoid breaking tests, I move current files under `api/scanner/test_data` to `api/scanner/test_data/scanner`.


kkovaletp:
@googollee, it's great that you touched this complex topic. Currently, we have a bit of a mess with the test data, which is located in several folders and used by different tests with no common rules. I'd propose moving all currently existing test data to its own folder structure under `/api/test_data`, where we can replicate the module structure so that all the testing data is located in one folder branch. At the same time, it is clear which test uses which files.
As a next step, I'd suggest adding these new small real media files under `/api/test_data/generic` (or `common`, `shared`, etc. - naming is always a difficult thing) to indicate that they are used by many different tests and expandable in the future. Then we can iteratively refactor the existing tests to use those generic test data for the cases, where it is suitable (keeping corner and negative cases being handled by the test-specific data from their own folders).

What do you think?

googollee:
> I'd propose moving all currently existing test data to its own folder structure under `/api/test_data`

I thought about it. Currently, I find only `scanner` operates with real files. Other packages should use `scanner` package instead of touching files directly. Ideally, I'd like `scanner` provides a facade to isolate file operations. So I put all those files under `scanner/test_data`. All other packages should use some mocking interfaces to test.

I don't want to use those small files with integration tests. Files used by integration tests should be at the same level of `api`/`ui`.

kkovaletp:
> I'd like `scanner` provides a facade to isolate file operations. So I put all those files under `scanner/test_data`

Makes sense.

I went through the repo structure and see that currently we have only these testing data paths:
- `scanner/media_encoding/executable_worker/testdata`
- `scanner/test_data`
I'd like both to be merged under the same `scanner/test_data` folder, so please move also the 1st one to the `scanner/test_data/executable_worker`. 
I don't see the `scanner/test_data/fake_media` in your PR. Did you forget it?
Also, the name `scanner/test_data/scanner` doesn't look self-descriptive. Can you please name it according to the purpose of its usage?

Do I miss something else?)

googollee:
`scanner/media_encoding/executable_worker/testdata` contains files only used by `scanner/media_encoding/executable_worker`. If files are only used by one package, it should be in that package but not export to others.

I plan to remove `scanner/test_data/fake_media` with https://github.com/photoview/photoview/pull/1124, so I didn't move them to `scanner/test_data/scanner`. They are still used by current code, which is going to be updated by #1124. The `media_type` in #1124 will use `libmagic` to check file types instead of ext names. That doesn't work with any fake media files.

I don't have a good idea for `scanner/test_data/scanner`. Files in it are used by `scanner` package and `cleanup_tasks` package. Since those files are simulating a real library, how about `photo_library`?

kkovaletp:
> `scanner/media_encoding/executable_worker/testdata/bin/*` are a group of mocking scripts.

Hmm, I see... Maybe it makes sense to rename the folder from `testdata/bin` to `test_mocks`. What do you think?

> I only want to share medias between packages, because they are relatively independent from the code and it's reasonable to share them between packages.

I'm thinking, is there a possibility in the future test development, that we would need some other test data except of the media files? Maybe some DB presets in SQL to test functionality, which requires exact data in the DB as a prerequisite (some state transitions in the scanner and face recognition)... In this case, it still makes sense to have a more generic folder name

kkovaletp:
> If we need such files, we should create an `api/test_data` at that time.

Ok. Looks like having many `test_data` folders across the project is unavoidable...

I see, there are failed tests. I'll review the PR once they are fixed, assuming more changes to be done.

kkovaletp (approved):
LGTM

PR #1136, opened 2024-11-23
Enhance Codecov integration

kkovaletp:
- Define sub-projects and set coverage threshold for failure in the `codecov.yml`: https://docs.codecov.com/docs/commit-status
- Make both test execution engines to generate also JUnit reports 
- Upload these JUnit reports to Codecov for analysis: https://docs.codecov.com/docs/test-analytics

### DIFF
--- a/scripts/test_api_coverage.sh
+++ b/scripts/test_api_coverage.sh
@@ -1,5 +1,5 @@
-#!/bin/sh
+#!/bin/bash
 set -eu
 
 cd $(dirname $0)/../api
-go test ./... -v -database -filesystem -p 1 -coverprofile=coverage.txt -covermode=atomic
+go test ./... -v -database -filesystem -p 1 -coverprofile=coverage.txt -covermode=atomic 2>&1 | tee >(go-junit-report >test-api-coverage-report.xml)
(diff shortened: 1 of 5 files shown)


scripts/test_api_coverage.sh
```diff
@@ -1,5 +1,5 @@
-#!/bin/sh
+#!/bin/bash
```
googollee, line 1: I used `/bin/sh` before because I hope to limit scripts with POSIX shell for better compatibility and portability.

I know `bash` provides more features. We could migrate to `bash` since it's already widely provided across Linux distributions. If so, please change the 2nd line to `set -euo pipeline`, or even `set -euxo pipeline` for better debugging. (`sh` doesn't support `set -o pipeline`)

kkovaletp: I had to switch to Bash because the `tee >(go-junit-report >test-api-coverage-report.xml)` construction doesn't work correctly in SH
I'll add the `set -euxo pipeline`

kkovaletp: https://github.com/photoview/photoview/actions/runs/11998663626/job/33445679242?pr=1136#step:6:18 - `set -euxo pipeline` causes error. Please suggest how to fix it. For now I'll revert it back.

googollee: Sorry, my fault. It should be `set -euxo pipefail`.

kkovaletp: Done

googollee (approved):

googollee:
Approved.

PR #1139, opened 2024-12-07
Try to fix Codecov upload test results error

kkovaletp:


kkovaletp:
Both tries are unsuccessful.
I've submitted the bug in the action's repo, let's see what they suggest:
https://github.com/codecov/test-results-action/issues/104

kkovaletp:
There is no need to keep this PR anymore

PR #1157, opened 2025-01-11
Add support for external HTTP authentication

emersion:
Allow administrators to delegate password checks to another HTTP server. This is useful when multiple services are offered to users (Photoview being one of them) and passwords are handled by a separate centralized authentication server.

## Summary by CodeRabbit

- **New Features**
	- Added support for external HTTP-based user authentication
	- Introduced a new configuration option for HTTP user authentication endpoint

- **Configuration**
	- Added `PHOTOVIEW_HTTP_USER_AUTH` environment variable to enable optional external authentication

### DIFF
--- a/api/example.env
+++ b/api/example.env
@@ -37,3 +37,6 @@ PHOTOVIEW_DEVELOPMENT_MODE=1
 # Support `qsv`, `vaapi`, `nvenc`.
 # Only `qsv` is verified with `/dev/dri//dev/dri` devices.
 # PHOTOVIEW_VIDEO_HARDWARE_ACCELERATION=
+
+# Set an HTTP endpoint to use for user authentication
+#PHOTOVIEW_HTTP_USER_AUTH=http://localhost:8080
--- a/api/graphql/models/user.go
+++ b/api/graphql/models/user.go
@@ -3,8 +3,10 @@ package models
 import (
 	"crypto/rand"
 	"fmt"
+	"net/http"
 	"time"
 
+	"github.com/photoview/photoview/api/utils"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/bcrypt"
 	"gorm.io/gorm"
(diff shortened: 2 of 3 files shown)


kkovaletp:
@emersion, thanks for contributing the PR!
Actually, we already have the #624 to track and discuss the solution. Please read it if you haven't seen it before and align your solution with at least some of the mentioned requirements and/or solutions. Feel free to comment on that issue if you'd like to discuss this topic more before the implementation.

@jordy2254 was working on this epic before. Jordy, if you are still available, feel free to provide your opinion on this PR.
@googollee, I'll assign you as a reviewer. Please also let us know what you think.

kkovaletp:
@emersion, as far as I understand, you'd like to have an authentication service in front of all your self-hosted services and once users pass it, let them go to those services with no additional logins, right? So, this is the SSO approach.
In this case there are several standard SSO protocols between the app and the auth service: Oauth, SAML are the most commonly used and the Oauth is the most secure one.
If we're going to implement an SSO auth, I'd vote for the Oauth protocol to be implemented.

BTW, I see no changes on the UI side, while the client's browser needs to set session cookies. I think that this might be missing...

emersion:
No, this isn't what this PR is doing. This PR uses an external server for checking usernames/passwords, instead of using the database. No UI change is needed.

OAuth support would be nice, but is a much more involved change and is orthogonal to this PR.

kkovaletp:
> given the direction this project is going (e.g. AI nonsense)

Face recognition was implemented long time ago and not changed since that. The recent changes are focused on a better media processing flow. What exactly don't you like?
Anyway, you're free to fork, as this is an open-source project.

emersion:
> What exactly don't you like?

I was referring to CodeRabbit comments.

googollee:
[NGINX's auth_request directive](https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-subrequest-authentication/) is not a common feature, since it's not enabled by default and requires compiling with `--with-http_auth_request_module`.

> Having an auth server in front of photoview is pretty weird in my opinion, I've never seen such as setup except for single-user deployments, and UX-wise presenting HTTP basic auth directly to users isn't great.

I believe it's a common pattern, since most clusters have a frontend/api-gateway to terminate https, do auth, and route traffic. This gateway may show a friendly page to users to authenticate, and proxy traffic if passed.

> It also begs a few questions, e.g. how to communicate the verified credential to photoview in a standard way (there's isn't a standard for this AFAIK).

Yes. I read the feature as dedicating the user authentication to an external server. I know OAuth and ldap are common solutions for this feature. I didn't mean it's wrong. Could you describe more about how to set up your server? For example, how do you create each user in photoview without authenticating with photoview?

PR #1176, opened 2025-02-25
Add coverage of all sub-packages of Go unit tests.

googollee:
## Summary by CodeRabbit

- **Tests**
	- Enhanced automated test coverage to include all modules, ensuring a more comprehensive validation process during testing.

### DIFF
--- a/scripts/test_api_coverage.sh
+++ b/scripts/test_api_coverage.sh
@@ -2,4 +2,4 @@
 set -euxo pipefail
 
 cd "$(dirname $0)/../api"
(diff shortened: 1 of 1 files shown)


googollee:
@kkovaletp If I read codecov report right, this change increases the coverage rate by 15.73%. Please double check it.

Without this change, unit tests only calculate the coverage within the package. This change enables the coverage of all packages in the project, if unit tests call them.

kkovaletp (approved):
Good catch! Thanks!
Codecov report's indirect changes page shows a lot of coverage improvements)

However, this impacts (fixes) only the measurements. Real level of coverage is the same as before.

googollee:
> However, this impacts (fixes) only the measurements. Real level of coverage is the same as before.

Yes. But now we can use codecov to trace the coverage of some packages when updating the code.

PR #1194, opened 2025-04-19
Reduce security vulnerabilities via dependency upgrades

PeterDaveHello:
Upgraded Babel, esbuild, rollup, micromatch, nanoid, vue-template-compiler and related packages to address npm audit issues and lower the number of vulnerabilities.

Vulnerabilities reduced from 22 to 17 (moderate: 13→9, high: 5→4, critical: 4 unchanged).

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -233,11 +233,13 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.24.2",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.24.2.tgz",
-      "integrity": "sha512-y5+tLQyV8pg3fsiln67BVLD1P13Eg4lh5RW9mF0zUuvLrv9uIQ4MCL+CRT+FTsBlBjcIan6PGsLcBN0m3ClUyQ==",
+      "version": "7.26.2",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.26.2.tgz",
+      "integrity": "sha512-RJlIHRueQgwWitWgF8OdFYGZX328Ax5BCemNGlqHfplnRT9ESi8JkFlvaVYbS+UubVY6dpv87Fs2u5M29iNFVQ==",
+      "license": "MIT",
       "dependencies": {
-        "@babel/highlight": "^7.24.2",
+        "@babel/helper-validator-identifier": "^7.25.9",
+        "js-tokens": "^4.0.0",
         "picocolors": "^1.0.0"
       },
       "engines": {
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
@CodiumAI-Agent /review

PeterDaveHello:
@coderabbitai review

PeterDaveHello:
@coderabbitai summary

kkovaletp:
@PeterDaveHello, thanks, but:
- There is already https://github.com/photoview/photoview/pull/1016 with many UI version updates, waiting for the final review and merge
- When you create a PR with version updates for UI, you need to include at least `package.lock` and `package.json` files. Adding only the lock file is not enough.

Anyway, I'm closing this PR as not planned.

PeterDaveHello:
@kkovaletp With all due respect, I think there might be a misunderstanding regarding the purpose and effect of this PR.

Updating only the `package-lock.json` is actually a standard and recommended way to address security vulnerabilities in transitive dependencies, as documented in the official npm documentation. When running `npm audit fix`, it's common that only the lock file gets changed, especially if the patched versions are already within the semver range specified in `package.json`, or if the vulnerabilities are in indirect dependencies.

Addressing these security issues by updating the lock file does **not** affect application functionality, but it does reduce risk and improve the project's security posture immediately. In many projects, such lock file updates are regularly merged ahead of broader dependency upgrades for exactly this reason.

Would you please reconsider handling security-related lock file fixes separately and with priority, since they are safe and do not impact the direct dependencies or application behavior?

Thank you!

kkovaletp:
@PeterDaveHello, thanks for the clarification. I didn't get it from your description that you update only indirect dependency versions, and your `package.json` is not changed, which is why it wasn't included.
However, merging this PR still does not make much sense without updating the main dependencies altogether. Also, it is not so safe for so old versions, as there might be mistakenly not fixed dependencies in those old versions, so blindly updating indirect dependencies might cause unexpected version incompatibility.

I think that it is better to wait for a UI maintainer and do the complex version update.

PeterDaveHello:
@kkovaletp, Thank you for your reply. Security updates for indirect dependencies are applied separately from major version upgrades to address known vulnerabilities while maintaining compatibility within existing constraints.

This PR updates only transitive dependencies within allowed version ranges, preserving the project’s stability and functionality. The changes were generated by `npm audit fix`, which followed npm’s recommended security practices to ensure the project remained protected against published vulnerabilities.

Considering these points, prioritizing the security fixes could help maintain the project's security posture effectively.

kkovaletp:
@PeterDaveHello, I understand your point.
My point is that the current set of dependencies was frozen back in 2021 or 2022. It is now consistent: all main and transitive dependencies' versions were tested for compatibility.
The concern is: when you update transitive dependencies without updating the corresponding main dependencies, there is a high chance of getting some partially-incompatible versions. This is because many libs and packages specify their own dependencies as "newer than" instead of an "exact" match. Later, new versions with breaking changes were released, and not every library maintains breaking changes carefully.
I saw this issue several times in other projects, this is not a theoretical probability. We cannot blindly trust `npm audit fix`, based on such dependency declaration trends.

PeterDaveHello:
@kkovaletp Thank you for elaborating. I understand your concern regarding potential incompatibilities, it's a valid point, especially with older dependency sets.

The detailed `package-lock.json` diff offers transparency here, allowing us to review the specific version changes directly and assess potential compatibility risks together.

I appreciate this discussion, as it helps clarify these crucial aspects for everyone involved. Ultimately, the decision on how to proceed rests with you, and I respect your judgment.

Thank you again for your time and consideration on this matter.

PR #1197, opened 2025-05-01
Bump GQLgen, GQLparser and GORM versions in API

kkovaletp:
## Summary by CodeRabbit

- **Chores**
  - Updated internal dependencies to the latest versions for improved stability and performance.


kkovaletp:
@googollee, can you please take a look into the docker build failure? I cannot find the real issue in the log...
Is the new GORM version incompatible with our project? What should be changed/fixed to make it compatible? Does it need a big effort?

googollee (approved):

kkovaletp:
@googollee, thanks for approving, but did you have a chance to look into the build error issue, caused by the new GORM version?
Any suggestions on how to fix it?
Or, would you prefer to keep its previous version for now and let you handle its version update later?

googollee:
It's complex now. I don't know if it was my mistake or Docker Hub did some changes. Now the platform is `arm32v7` instead of `armv7`, see https://hub.docker.com/r/arm32v7/debian/tags. Since the error is in `armv7`, I can't reproduce it locally as Docker Hub doesn't support it.

I'll create a new PR to fix the name `arm32v7` first.

googollee:
I confirmed this error only happens in `arm/v7`. I'm checking now.

googollee:
From the log:

```
2025-05-03T07:01:37.2027213Z #70 7.658 gorm.io/gorm/internal/stmt_store
2025-05-03T07:01:37.2027673Z #70 7.684 # gorm.io/gorm/internal/stmt_store
2025-05-03T07:01:37.2039229Z #70 7.684 /go/pkg/mod/gorm.io/gorm@v1.26.0/internal/stmt_store/stmt_store.go:99:10: cannot use defaultMaxSize (untyped int constant 9223372036854775807) as int value in assignment (overflows)
```

When I rollback `gorm` to `v1.25.12`, it works well. So it's a issue in the upstream. Could you try to use that version first>

googollee:
The real problem is fixed with https://github.com/go-gorm/gorm/pull/7439, which is in the master branch. We can update `gorm` next time.

PR #1199, opened 2025-05-03
More granular `media-cache` folder permissions

### LINKED ISSUE #1181: Default HOST_PHOTOVIEW_LOCATION doesn't setup with correct permissions
Just testing this out on my local machine (macos, docker desktop). Following the [docker setup](https://github.com/photoview/photoview?tab=readme-ov-file#getting-started--setup-with-docker) results in an error from the UI when trying to scan photos:

`Scanner error Error scanning media for album (1) file (/photos/***.jpg): cache directory error (/photos/***.jpg): could not make album image cache directory: mkdir /home/photoview/media-cache/1: permission denied`

The `HOST_PHOTOVIEW_LOCATION` in this case is set to the default `/opt/photoview`. 

After running `sudo make all` the permissions on `/opt/photoview` on my host are like so:
```
drwxr-xr-x  3 root  wheel  96 Mar 17 00:20 database
drwxrwxrwx  2 root  wheel  64 Mar 17 00:20 storage
```
I double checked that `other` permissions include execute for each directory up the tree all the way to root.
I had to use `sudo` for `make all` because my user does not have write access to `/opt` to create the directories.

If I change `HOST_PHOTOVIEW_LOCATION` to somewhere under my user home folder, everything seems to work fine. That's fine but.. what gives? I don't understand why the default location doesn't seem to work. `other` has write permission to `/opt/photoview/storage`, which is mapped to `home/photoview/media-cache` on the container, and `other` also has execute permission on `/opt/photoview` and `/opt`.

kkovaletp:
Fixes #1181 

Add a preparation service to assign the correct permissions to the media cache folder instead of making it writable by anyone.

## Summary by CodeRabbit

- **New Features**
  - Added a preparation step to Docker Compose setups to ensure correct permissions for the media cache directory before starting the main service.

- **Documentation**
  - Improved comments in configuration files to clarify how to update media cache paths and related settings.

- **Chores**
  - Adjusted Makefile and Docker Compose examples to manage directory permissions externally rather than setting them directly.

### DIFF
--- a/docker-compose example/docker-compose.example.yml
+++ b/docker-compose example/docker-compose.example.yml
@@ -1,4 +1,20 @@
 services:
+  ## Makes sure that the media cache folder is created with the correct permissions
+  photoview-prepare:
+    image: photoview/photoview:2
+    hostname: photoview-prepare
+    container_name: photoview-prepare
+    user: root
+    entrypoint: []
+    command: /bin/bash -c "sleep 1 && chown -R photoview:photoview /home/photoview/media-cache"
+    security_opt:
+      - seccomp:unconfined
+      - apparmor:unconfined
+    volumes:
+      - "/etc/localtime:/etc/localtime:ro" ## use local time from host
+      - "/etc/timezone:/etc/timezone:ro"   ## use timezone from host
+      - "${HOST_PHOTOVIEW_LOCATION}/storage:/home/photoview/media-cache"
+
   photoview:
     image: photoview/photoview:2
     labels:
@@ -13,6 +29,8 @@ services:
     ## Comment out the entire `depends_on` section if PHOTOVIEW_DATABASE_DRIVER is set to `sqlite` in the .env
     ## Or comment out the `mariadb:` and uncomment the `postgres:` if PHOTOVIEW_DATABASE_DRIVER is set to `postgres`
     depends_on:
+      photoview-prepare:
+        condition: service_completed_successfully
       # postgres:
       mariadb:
         condition: service_healthy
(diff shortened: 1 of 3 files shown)


kkovaletp:
@googollee, please review this small security fix

docker-compose example/docker-compose.minimal.example.yml
```diff
@@ -1,4 +1,18 @@
 services:
+  ## Makes sure that the media cache folder is created with the correct permissions
+  photoview-prepare:
+    image: photoview/photoview:2
+    hostname: photoview-prepare
+    container_name: photoview-prepare
+    user: root
+    entrypoint: []
+    command: /bin/bash -c "sleep 1 && chown -R photoview:photoview /home/photoview/media-cache"
+    cap_add: [CHOWN]
```
googollee, line 10: Suggest to use:

```
cap_add:
  - CHOWN
```

It's easy to add more cap in the future.

docker-compose example/docker-compose.example.yml
```diff
@@ -1,4 +1,18 @@
 services:
+  ## Makes sure that the media cache folder is created with the correct permissions
+  photoview-prepare:
+    image: photoview/photoview:2
+    hostname: photoview-prepare
+    container_name: photoview-prepare
+    user: root
+    entrypoint: []
+    command: /bin/bash -c "sleep 1 && chown -R photoview:photoview /home/photoview/media-cache"
+    cap_add: [CHOWN]
```
googollee, line 10: Suggest to use:

```
cap_add:
  - CHOWN
```

It's easy to add more cap in the future.

googollee (approved):

docker-compose example/docker-compose.minimal.example.yml
kkovaletp: Sure, I've expanded the list

docker-compose example/docker-compose.example.yml
kkovaletp: Sure, I've expanded the list

PR #1202, opened 2025-05-09
Refactor API video route

kkovaletp:
This PR was originally initiated by the issue, detected by Coverity in my forked repo:

> CID 1588385: Explicit null dereferenced (FORWARD_NULL)
> var_deref_model: Passing null pointer media to authenticateMedia, which dereferences it

But then I did more code refactoring and added tests

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of expired or invalid share tokens, ensuring unauthorized access is correctly denied with appropriate error messages.
- **New Features**
	- Enhanced video route to support multiple video sources with improved error handling during video processing and caching.
	- Added mock video routes for testing with configurable cache paths.
- **Tests**
	- Added comprehensive tests for video routes covering authentication, cache handling, multiple media URLs, error scenarios, and context cancellation.
	- Expanded authentication tests to cover additional share token edge cases.
- **Refactor**
	- Modularized video route logic for improved maintainability and testability.
	- Updated media processing to support context-aware cancellation.
	- Replaced standard logging with structured, context-aware logging across routes.
- **Chores**
	- Improved thread safety for test cache path configuration.


eyestoreoye:
Great implementation of this feature. I particularly like the approach to error handling.

eyestorefivemstore:
The performance optimizations here are quite clever. Nice engineering work!

api/routes/videos.go
```diff
@@ -1,72 +1,134 @@
...
+	db *gorm.DB,
+	mediaName string,
+	authenticateFn func(*models.Media, *gorm.DB, *http.Request) (bool, string, int, error),
+	getCachePathFn func(albumID, mediaID int, filename string) string,
+) {
+	var mediaURLs []models.MediaURL
+	if err := db.Model(&models.MediaURL{}).
+		Select("media_urls.*, Media.*").
+		Preload("Media").
+		Joins("Media").
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, thanks for your analysis. I noticed that you also suggest dropping the `Select("media_urls.*, Media.*").` part from the query. Are you sure that it is correct and that the results will be exactly the same (so the code won't see any difference, continuing to work as before)?

api/routes/videos_test.go
```diff
@@ -0,0 +1,261 @@
...
+// setTestCachePath temporarily sets a different media cache path for testing
+// and returns a function to restore the original state
+func setTestCachePath(tempPath string) func() {
+	// Set the test value
+	utils.ConfigureTestCache(tempPath)
+	// Return function to clear the override for this goroutine
+	return func() {
+		utils.ConfigureTestCache("")
+	}
+}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, it was you who suggested removing the `original` variable from the package-level function to make it thread-safe. Now you suggest reverting this change because of the same reason. Please carefully analyze the way backend unit tests are configured and executed in this project on 3 supported DBMSs. Then, please find the best way to make this function thread safe: let this test code run in parallel for 3 DBMSs and run all test cases in parallel for each DBMS.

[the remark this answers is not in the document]
kkovaletp: @coderabbitai, great, thanks! Could you please also review the other code in the `video_test.go` and suggest the needed changes to make sure it is thread-safe? Do we need to set file names, depending on the thread to make them different? Anything else?
Finally, please suggest how to modify the code to run the file's test cases in parallel?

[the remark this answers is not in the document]
kkovaletp: @coderabbitai, I see some unverified assumptions in your suggested code in the block 5 "complete restructured example". Also, some currently existing code is missing there. Finally, as you've moved all package-level declarations inside the function after the `testCases` code, a lot of undefined variables are reported by VScode. 
Please carefully analyze these issues and the existing code, and provide me the complete solution with all issues fixed and all assumptions verified according to the product and testing code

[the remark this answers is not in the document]
kkovaletp: @coderabbitai, after executing the new implementation of the test code, I've got the next errors:
```
=== NAME  TestVideoRoutes/Video_file_not_in_cache,_needs_processing
    videos_test.go:307:
        	Error Trace:	/app/api/routes/videos_test.go:307
        	Error:      	Received unexpected error:
        	            	bcrypt: password length exceeds 72 bytes
        	            	failed to generate hash for share password
        	            	github.com/photoview/photoview/api/graphql/models/actions.hashSharePassword
        	            		/app/api/graphql/models/actions/share_token_actions.go:143
        	            	github.com/photoview/photoview/api/graphql/models/actions.AddMediaShare
        	            		/app/api/graphql/models/actions/share_token_actions.go:40
        	            	github.com/photoview/photoview/api/routes.TestVideoRoutes.func12
        	            		/app/api/routes/videos_test.go:306
        	            	testing.tRunner
        	            		/usr/local/go/src/testing/testing.go:1792
        	            	runtime.goexit
        	            		/usr/local/go/src/runtime/asm_amd64.s:1700
        	Test:       	TestVideoRoutes/Video_file_not_in_cache,_needs_processing

=== NAME  TestVideoRoutes/Valid_video_retrieval
    videos_test.go:262:
        	Error Trace:	/app/api/routes/videos_test.go:262
        	Error:      	Received unexpected error:
        	            	FOREIGN KEY constraint failed
        	Test:       	TestVideoRoutes/Valid_video_retrieval

=== NAME  TestVideoRoutes/Multiple_media_URLs_with_same_name
    videos_test.go:307:
        	Error Trace:	/app/api/routes/videos_test.go:307
        	Error:      	Received unexpected error:
        	            	FOREIGN KEY constraint failed
        	            	failed to insert new share token into database
        	            	github.com/photoview/photoview/api/graphql/models/actions.AddMediaShare
        	            		/app/api/graphql/models/actions/share_token_actions.go:55
        	            	github.com/photoview/photoview/api/routes.TestVideoRoutes.func12
        	            		/app/api/routes/videos_test.go:306
        	            	testing.tRunner
        	            		/usr/local/go/src/testing/testing.go:1792
        	            	runtime.goexit
        	            		/usr/local/go/src/runtime/asm_amd64.s:1700
        	Test:       	TestVideoRoutes/Multiple_media_URLs_with_same_name

=== NAME  TestVideoRoutes/Authentication_with_share_token
    videos_test.go:307:
        	Error Trace:	/app/api/routes/videos_test.go:307
        	Error:      	Received unexpected error:
        	            	FOREIGN KEY constraint failed
        	            	failed to insert new share token into database
        	            	github.com/photoview/photoview/api/graphql/models/actions.AddMediaShare
        	            		/app/api/graphql/models/actions/share_token_actions.go:55
        	            	github.com/photoview/photoview/api/routes.TestVideoRoutes.func12
        	            		/app/api/routes/videos_test.go:306
        	            	testing.tRunner
        	            		/usr/local/go/src/testing/testing.go:1792
        	            	runtime.goexit
        	            		/usr/local/go/src/runtime/asm_amd64.s:1700
        	Test:       	TestVideoRoutes/Authentication_with_share_token
```
Please find the root cause and fix it consistently. It looks like the new test value is incompatible with the product code constraints and limitations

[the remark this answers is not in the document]
kkovaletp: The tests are still failing because of constraints and many 404 errors. I don't want to spend more time on this, taking into account that other tests are running sequentially. Making all tests thread-safe and running in parallel will be another global task out of this PR scope.

api/routes/videos_test.go
```diff
@@ -0,0 +1,260 @@
...
+				os.Remove(videoPath)
+
+				return httptest.NewRecorder()
+			},
+			validateFunc: func(t *testing.T, rr *httptest.ResponseRecorder) {
+				// This will likely fail in tests because ProcessSingleMedia can't fully work in tests
+				// When ProcessSingleMedia fails, we expect 500
+				assert.Contains(t, []int{http.StatusOK, http.StatusInternalServerError}, rr.Code)
+			},
+			cleanupFunc: nil,
```
kkovaletp, line 229: @coderabbitai, I don't like this test case, as it doesn't verify anything: it is always passed, even if an error occurred during the code execution. It is because it expects both success and failure.
I understand that the reason for implementing it this way was my request to keep things simple and easily maintainable. But having a test, which is always passed no matter what - is not acceptable.
So, please carefully analyze the project code, the testing framework, and the testing code of this file and related files. Once done, please find the best way to make this test work according the unit tests best practices, while keeping it maintainable and reliable.

kkovaletp: @coderabbitai, when I added your mocked function, I see the next errors in VScode for the lines with assignments to the `scanner.ProcessSingleMedia`:
- "use of package scanner not in selector compiler[InvalidPkgUse](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#InvalidPkgUse)"
- "cannot assign to scanner.ProcessSingleMedia (neither addressable nor a map index expression) compiler[UnassignableOperand](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#UnassignableOperand)"

Please find the best solution for these errors.

api/routes/videos_test.go
```diff
@@ -0,0 +1,329 @@
...
+			}
+
+			tc.validateFunc(t, rr)
+
+			if tc.cleanupFunc != nil {
+				tc.cleanupFunc(t)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, thanks for this suggestion, but isn't this new code redefining the default testing framework's main function? If yes, I suspect it can affect other tests execution - those executed after this file. I want to make sure that whatever happens inside this file during the test suite execution, other test files are not impacted.
Can we use `defer` instead?

api/routes/videos_test.go
```diff
@@ -0,0 +1,329 @@
...
+	expiry := time.Now().Add(24 * time.Hour)
+	shareToken, err := actions.AddMediaShare(db, user, media.ID, &expiry, &tokenPassword)
+	require.NoError(t, err)
+
+	// Create two routers - one with real handler for auth tests, one with mock handler for other tests
+	realRouter := mux.NewRouter()
+	RegisterVideoRoutes(db, realRouter)
+
+	mockRouter := mux.NewRouter()
+	registerMockVideoRoutesForTesting(db, mockRouter, tempCachePath)
```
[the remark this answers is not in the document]
kkovaletp: No need. IDE will clearly show the implementation location

api/routes/videos_test.go
```diff
@@ -0,0 +1,329 @@
...
+	utils.ConfigureTestCache(tempPath)
+	return func() {
+		utils.ConfigureTestCache(original)
+	}
+}
+
+// mockProcessSingleMedia replaces scanner.ProcessSingleMedia with a mock function during tests
+// and returns a function to restore the original implementation
+var originalProcessSingleMedia = processSingleMediaFn
+
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, please double-verify your comment. I successfully executed this code locally and on CI with no issues. Also, it is declared in the same package (in `videos.go`), so it is accessible from there.

kkovaletp:
@googollee, please review this PR. It is the second part of my fixes triggered by Coverity findings.
Here is another PR from my forked repo for the same branch, where the bot provided many other suggestions: https://github.com/kkovaletp/photoview/pull/449 - just in case, you want to look at them)

kkovaletp:
@googollee, this is the last part of my refactoring cycle, which is still waiting for your review ;-)

api/utils/media_cache.go
```diff
@@ -37,9 +38,20 @@ func CachePathForMedia(albumID int, mediaID int) (string, error) {
 	return photoCachePath, nil
 }
 
-var testCachePath string = ""
+var (
+	testCachePath string = ""
```
googollee: Please protect all usage of `testCacchePath`, e.g. in `MediaCachePath()`.

api/routes/authenticate_routes.go
```diff
@@ -72,9 +73,16 @@ func shareTokenFromRequest(db *gorm.DB, r *http.Request, mediaID *int, albumID *
 	var shareToken models.ShareToken
 
 	if err := db.Where("value = ?", token).First(&shareToken).Error; err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return false, "unauthorized", http.StatusForbidden, errors.New("share token not found")
```
googollee: I suggest the error message is "invalid share token".

api/routes/authenticate_routes.go
```diff
@@ -72,9 +73,16 @@ func shareTokenFromRequest(db *gorm.DB, r *http.Request, mediaID *int, albumID *
...
+		}
 		return false, internalServerError, http.StatusInternalServerError, err
 	}
 
+	if shareToken.Expire != nil && time.Now().UTC().After(shareToken.Expire.UTC()) {
+		return false, "unauthorized", http.StatusForbidden, errors.New("share token expired")
```
googollee: `share token expired` exposes a infomation that the token was valid. I suggest the error message is "invalid share token". Since a expired token equals to an invalid token, it's better not exposing that info.

api/routes/videos.go
```diff
@@ -1,72 +1,179 @@
...
+			w, r, db, mediaName,
+			authenticateMedia, // Real authentication
+			func(albumID, mediaID int, filename string) string {
+				return path.Join(utils.MediaCachePath(), strconv.Itoa(albumID), strconv.Itoa(mediaID), filename)
+			},
+		)
+	})
+}
+
+func registerMockVideoRoutesForTesting(db *gorm.DB, router *mux.Router, tempCachePath string) {
```
googollee: Put this function in `_test.go` file.

api/routes/videos.go
```diff
@@ -1,72 +1,179 @@
...
+
+func RegisterVideoRoutes(db *gorm.DB, router *mux.Router) {
+
+	router.HandleFunc("/{name}", func(w http.ResponseWriter, r *http.Request) {
+		mediaName := mux.Vars(r)["name"]
+
+		handleVideoRequest(
+			w, r, db, mediaName,
+			authenticateMedia, // Real authentication
+			func(albumID, mediaID int, filename string) string {
```
googollee: A function called `GetCacheFilename` is better.

api/routes/videos.go
```diff
@@ -1,72 +1,179 @@
...
-			w.WriteHeader(http.StatusNotFound)
-			w.Write([]byte("404"))
-			return
+		w.WriteHeader(http.StatusNotFound)
+		w.Write([]byte("not found"))
+		return
+	}
+
+	if len(mediaURLs) > 1 {
+		sanitizedMediaName := strings.Map(func(r rune) rune {
```
googollee: Did we see any non-printable name?

api/routes/videos.go
```diff
@@ -1,72 +1,179 @@
...
+	r *http.Request,
+	db *gorm.DB,
+	mediaName string,
+	authenticateFn func(*models.Media, *gorm.DB, *http.Request) (bool, string, int, error),
+	getCachePathFn func(albumID, mediaID int, filename string) string,
+) {
+	var mediaURLs []models.MediaURL
+	if err := db.Model(&models.MediaURL{}).
+		Preload("Media").
+		Where("media_urls.media_name = ? AND media_urls.purpose = ?", mediaName, models.VideoWeb).
```
googollee, line 34: I see here has an additional condition `media_urls.purpose = ?`. Could you provide more detail about it?

api/routes/videos.go
```diff
@@ -1,72 +1,179 @@
...
+	contextAwareProcessFn := func(db *gorm.DB, media *models.Media) error {
+		// If the client disconnects during processing, this will be detected
+		ctx := r.Context()
+
+		// Create a channel to communicate processing completion
+		done := make(chan error, 1)
+
+		go func() {
+			// Run the actual processing in a goroutine
+			done <- processSingleMediaFn(db, media)
```
googollee: If you want to add a context, please update `ProcessSingleMedia`. When the context cancelled, current code won't cancel `ProcessSingleMedia` in the background.

If you decide to run `ProcessSingleMedia` in the background, it's not necessary to wait for it done. Just `go ProcessSingleMedia()` and return a response.

What is the behavior here?

kkovaletp: @coderabbitai, please read the previous comment in this thread, reevaluate, and deeply analyze the project code, based on the latest commit in this branch. Once done, please suggest a consistent and reliable way to implement the changes requested in the previous comment. Please provide the exact code replacements in a copyable form.

api/routes/authenticate_routes.go
kkovaletp: Good point. I didn't think from this perspective(
I updated error messages for all relevant `http.StatusForbidden` cases to be the same and not expose the details for security reasons.
To be still able to investigate issues, I've added `log.Debug()` calls with all relevant details: I assume that if we have the `INFO` log level by default, those debug logging calls are ignored, so it is safe for production. However, I don't see the log level configuration anywhere: not in the `utils/environment_variables.go`, and not in the `log/*.go` files.
Please suggest how to configure the log level. If that is missing now, is it easy to add this functionality before merging this PR?

api/routes/authenticate_routes.go
kkovaletp: Same as the previous comment, resolving to keep things consolidated

api/routes/videos.go
kkovaletp: Done

api/routes/videos.go
kkovaletp: Sorry, I didn't find this function by name and didn't get what you meant by this comment. Do you want me to rename the `handleVideoRequest()` to `GetCacheFilename`, making it public (because of 1st capital letter)? Why?
Please explain with more context)

api/routes/videos.go
kkovaletp: I don't remember such cases, but this is just a precaution - do you see a negative impact from this sanitizing code?

api/routes/videos.go
kkovaletp: It refers to the next field in the table:
```sql
select distinct mu.purpose
from media_urls mu ;
```
```
purpose        |
---------------+
thumbnail      |
video-thumbnail |
video-web      |
high-res       |
original       |
```
By adding this filter to the query, we get only the records with `purpose` = `video-web`.
Is this logic wrong here?

api/utils/media_cache.go
kkovaletp: Thanks! I missed that function.
I'll try to protect it (as far as I see, the `MediaCachePath()` is the only unprotected place). Also, I'll rename the mutex to the `testCachePathLocker` to be consistent with my previous PR naming.

kkovaletp: @coderabbitai, please carefully analyze this case, including previous messages in this conversation thread and the source code in this branch.
When I made the changes:
- Consistently renamed the mutex to the `testCachePathLocker`
- Wrapped the `testCachePath` reading in the `MediaCachePath()` function by the `testCachePathLocker` locking:
```
func MediaCachePath() string {
	testCachePathLocker.RLock()
	if testCachePath != "" {
		return testCachePath
	}
	testCachePathLocker.RUnlock()

	photoCache := EnvMediaCachePath.GetValue()
	if photoCache == "" {
		photoCache = "./media_cache"
	}

	return photoCache
}
```
I executed the API unit tests locally and got the following panic exception:
```
panic: test timed out after 10m0s
	running tests:
		TestPhotoRoutes (9m56s)

goroutine 75 [running]:
testing.(*M).startAlarm.func1()
	/usr/local/go/src/testing/testing.go:2484 +0x394
created by time.goFunc
	/usr/local/go/src/time/sleep.go:215 +0x2d

goroutine 1 [chan receive, 10 minutes]:
testing.(*T).Run(0xc0001a2540, {0xd185e4?, 0xc0000ed9d8?}, 0xd452e8)
	/usr/local/go/src/testing/testing.go:1859 +0x431
testing.runTests.func1(0xc0001a2540)
	/usr/local/go/src/testing/testing.go:2279 +0x37
testing.tRunner(0xc0001a2540, 0xc0000edb18)
	/usr/local/go/src/testing/testing.go:1792 +0xf4
testing.runTests(0xc000010408, {0x12a9360, 0x3, 0x3}, {0x0?, 0x0?, 0x1326b80?})
	/usr/local/go/src/testing/testing.go:2277 +0x4b4
testing.(*M).Run(0xc00023f400)
	/usr/local/go/src/testing/testing.go:2142 +0x64a
github.com/photoview/photoview/api/test_utils.IntegrationTestRun(0xc00023f400)
	/app/api/test_utils/integration_setup.go:56 +0x3bf
github.com/photoview/photoview/api/routes_test.TestMain(0x1349f60?)
	/app/api/routes/routes_test.go:11 +0x13
main.main()
	_testmain.go:63 +0xa8

goroutine 69 [sync.RWMutex.Lock, 10 minutes]:
sync.runtime_SemacquireRWMutex(0x25?, 0xc0?, 0xc00032d200?)
	/usr/local/go/src/runtime/sema.go:105 +0x25
sync.(*RWMutex).Lock(0xe5ae8708?)
	/usr/local/go/src/sync/rwmutex.go:155 +0x6a
github.com/photoview/photoview/api/utils.ConfigureTestCache({0xd178be, 0xd})
	/app/api/utils/media_cache.go:53 +0x6f
github.com/photoview/photoview/api/routes.TestPhotoRoutes(0xc0004e0a80)
	/app/api/routes/photos_test.go:122 +0x688
testing.tRunner(0xc0004e0a80, 0xd452e8)
	/usr/local/go/src/testing/testing.go:1792 +0xf4
created by testing.(*T).Run in goroutine 1
	/usr/local/go/src/testing/testing.go:1851 +0x413

goroutine 8 [select, 10 minutes]:
database/sql.(*DB).connectionOpener(0xc0002608f0, {0xe096c0, 0xc00022acd0})
	/usr/local/go/src/database/sql/sql.go:1261 +0x87
created by database/sql.OpenDB in goroutine 7
	/usr/local/go/src/database/sql/sql.go:841 +0x130
FAIL	github.com/photoview/photoview/api/routes	600.250s
```

Please carefully analyze the panic output and suggest a consistent and robust fix by providing the exact code replacements in a copyable form.

api/routes/videos.go
kkovaletp: @coderabbitai, I like your implementation in option 2; however, you copied and duplicated the original code of the `ProcessSingleMedia()` function from the `scanner` package instead of updating it consistently to support the context, as requested in the original comment.
As far as I understand, code duplication is a bad programming pattern. But do you think that it is necessary here? Why exactly? If not, and that was just a coding mistake, please update your suggestion for option 2 and consistently implement the context awareness in the `scanner.ProcessSingleMedia()` function.
Do we need to update some of the unit tests related to the code modified by this change?

kkovaletp: @coderabbitai, while implementing your suggested solution, I noticed that the current `processSingleMediaFn` function has a significant difference in its code compared to the suggested one by you in step 4 for the `api/routes/videos_test.go`:
- this is my current code:
```
unc mockProcessSingleMedia(t *testing.T, shouldSucceed bool, mediaID int, albumID int) func() {
	// Save original implementation
	savedFn := processSingleMediaFn

	// Replace with mock implementation
	processSingleMediaFn = func(ctx context.Context, db *gorm.DB, media *models.Media) error {
		if shouldSucceed {
			// On success: create the expected video file in cache
			mediaURLs := []models.MediaURL{}
			if err := db.Where("media_id = ? AND purpose = ?", media.ID, models.VideoWeb).
				Find(&mediaURLs).Error; err != nil {
				return err
			}

			if len(mediaURLs) == 0 {
				return fmt.Errorf("no media URLs found")
			}

			// Get the cache path
			tempCachePath := utils.GetTestCachePath()
			albumDir := filepath.Join(tempCachePath, strconv.Itoa(albumID))
			mediaDir := filepath.Join(albumDir, strconv.Itoa(mediaID))
			if err := os.MkdirAll(mediaDir, 0755); err != nil {
				return err
			}

			videoPath := filepath.Join(mediaDir, mediaURLs[0].MediaName)
			return os.WriteFile(videoPath, []byte("mocked processed video content"), 0644)
		}

		// On failure: return an error
		return fmt.Errorf("mock processing error")
	}

	// Return cleanup function
	return func() {
		processSingleMediaFn = savedFn
	}
}
```
- and here is your suggestion:
```
func mockProcessSingleMedia(t *testing.T, shouldSucceed bool, mediaID int, albumID int) func() {
	// Save original implementation
	savedFn := processSingleMediaFn
	
	processSingleMediaFn = func(ctx context.Context, db *gorm.DB, media *models.Media) error {
		if shouldSucceed {
			// Create the expected cached video file
			cachePath := path.Join(utils.GetTestCachePath(), strconv.Itoa(albumID), strconv.Itoa(mediaID), "video.mp4")
			if err := os.MkdirAll(path.Dir(cachePath), 0755); err != nil {
				return err
			}
			return os.WriteFile(cachePath, []byte("test video content"), 0644)
		}
		return errors.New("simulated processing failure")
	}
	
	return func() {
		processSingleMediaFn = savedFn
	}
}

// Update the originalProcessSingleMedia variable
var originalProcessSingleMedia = func(ctx context.Context, db *gorm.DB, media *models.Media) error {
	return scanner.ProcessSingleMedia(ctx, db, media)
}
```
Please carefully analyze the function diffs and the testing code for this module. Once done, please tell me why you chaged the function, and provide me the most consistent function for the tests.

kkovaletp: @coderabbitai, thanks. I see that in the current implementation, the `ctx context.Context` is declared as a function parameter, but is never used inside the function. Is that expected for the testing implementation, or also a mistake?
If it is a mistake, please suggest a consistent and robust fix

kkovaletp: @coderabbitai, thanks for the fix, but this testing mocked function now looks even less similar to the one in the production code. I know that the purpose of this testing file is to test the video routes, not the scanner media functionality, but can you please analyze and make sure that the new implementation is still consistent and follows the purpose of testing the video routes package in a consistent manner with the production code behavior?

kkovaletp: @coderabbitai, thanks. Following this purpose and based on the production code behavior, should I add the cancellation test to this file? If yes, does the test code from your previous suggestion still match the final mock implementation? If not, please suggest the exact, consistent, and robust code for it.

kkovaletp: @coderabbitai, please carefully analyze the existing testing code + the final implementation of the `mockProcessSingleMedia()`, as well as the corresponding production code.
When I inserted your suggested test function for cancellation testing, I saw that you used non-existent DB setup and teardown functions, also, the `user` var was defined and never used.
Please make sure that your code suggestion is consistent, reliable, and correct!

kkovaletp: @coderabbitai, thanks, but the final assert in your latest test implementation hides many possible errors. Please define a more specific assert for this test

kkovaletp: @coderabbitai, I had to replace the `realRouter` with the testing router without auth flow, as it failed because of that and it doesn't make sense to implement the auth flow here.
Then I run the tests locally and see that this test fails with the next error:
```
    videos_test.go:457:
        	Error Trace:	/app/api/routes/videos_test.go:457
        	            				/app/api/routes/videos_test.go:468
        	Error:      	Not equal:
        	            	expected: 200
        	            	actual  : 404
        	Test:       	TestVideoRoutes/Context_cancellation_during_processing
        	Messages:   	Status should remain default when context cancelled
    videos_test.go:460:
        	Error Trace:	/app/api/routes/videos_test.go:460
        	            				/app/api/routes/videos_test.go:468
        	Error:      	Should be empty, but was 404 page not found
        	Test:       	TestVideoRoutes/Context_cancellation_during_processing
        	Messages:   	Response body should be empty when context cancelled
```
Please carefully analyze the error and suggest a consistent and robust fix. Make sure that no other errors hide behind these 2.

kkovaletp: @coderabbitai, your fix changed the behavior of the test, but it still fails:
```
    videos_test.go:456:
        	Error Trace:	/app/api/routes/videos_test.go:456
        	            				/app/api/routes/videos_test.go:467
        	Error:      	Not equal:
        	            	expected: 200
        	            	actual  : 500
        	Test:       	TestVideoRoutes/Context_cancellation_during_processing
        	Messages:   	Status should remain default when context cancelled
    videos_test.go:459:
        	Error Trace:	/app/api/routes/videos_test.go:459
        	            				/app/api/routes/videos_test.go:467
        	Error:      	Should be empty, but was internal server error
        	Test:       	TestVideoRoutes/Context_cancellation_during_processing
        	Messages:   	Response body should be empty when context cancelled
```
Make sure to test yourself and provide a really robust and consistent fix!

kkovaletp: @googollee, finally, after many iterations, I managed to implement the requested changes.
Please check the code and let me know if there is something I still need to address.

api/utils/media_cache.go
kkovaletp: @googollee, I've done the requested change. Please check the code and let me know if there is something I need to address.

api/routes/videos_test.go
```diff
@@ -0,0 +1,474 @@
...
+	// Create media with VideoWeb purpose
+	media := &models.Media{
+		Title:    fmt.Sprintf("Test Video %s", testID),
+		Path:     filepath.Join(t.TempDir(), mediaName),
+		PathHash: fmt.Sprintf("testhash-%s", testID),
+		AlbumID:  album.ID,
+		Album:    *album,
+		DateShot: time.Now(),
+		Type:     "video",
+	}
```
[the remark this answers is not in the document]
kkovaletp: This test struct follows the corresponding struct in the production code in the `graphql/models/media.go`.
@googollee, do we have the described defect in the production code, or is this a false detection?

api/routes/authenticate_routes.go
googollee: Currently there is no way to config the log level. If you use `log` package in the `photoview`, you can change [the default logger](https://github.com/photoview/photoview/blob/9369f72d3b0092744b1e370baac7638bf9e39128/api/log/context.go#L11) for a different level. It's annoying for debug now.

api/routes/videos.go
googollee: I can't say it's wrong. I see the origin didn't have this logic. If it works well, it's good to have it.

api/routes/videos.go
googollee: It's better to sanitize the string when it stores to the DB, instead of reading. Could we add that part until we have such problems?

api/routes/videos.go
googollee: I mean it's better to give a proper name to this inline function, not `handleVideoRequest()`.

api/routes/videos.go
googollee: I see this function is in `video.go`. Did you miss submitting code here?

api/routes/videos_test.go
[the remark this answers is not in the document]
googollee: I agree that this copy is not ideal code. I guess a easy fix is changing the field to `Album *Album`. I'm OK to keep it as what it is now, since it's a legacy code.

api/routes/videos_test.go
```diff
@@ -0,0 +1,391 @@
...
+				registerMockVideoRoutesForTesting(db, router, cachePath)
+
+				// Make request with nonexistent video name
+				req := httptest.NewRequest("GET", "/nonexistent.mp4", nil)
+				rr := httptest.NewRecorder()
+				router.ServeHTTP(rr, req)
+
+				// Validate response
+				assert.Equal(t, http.StatusNotFound, rr.Code)
+				assert.Equal(t, "not found", rr.Body.String())
```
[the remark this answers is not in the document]
googollee: There is no change in `MediaCachePath()`. Did I miss anything?

api/routes/videos.go
googollee: Sorry, I'm in a wrong commit.

api/routes/videos_test.go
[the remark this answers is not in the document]
googollee: Sorry, I'm in a wrong commit. Looks good.

api/routes/videos.go
```diff
@@ -1,72 +1,144 @@
...
+		log.Error(r.Context(), "Can not handle media_purpose for video",
+			"purpose", mediaURL.Purpose,
+			"expected", models.VideoWeb)
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte(internalServerError))
+		return
+	}
+
+	if _, err := os.Stat(cachedPath); err != nil {
+		if os.IsNotExist(err) {
```
googollee: It's better to revert the condition to reduce branches:

```
if !os.IsNotExist(err) {
  // Return an error
  return
}

// Process media
```

api/routes/videos.go
```diff
@@ -1,72 +1,144 @@
...
+				if r.Context().Err() != nil {
+					log.Warn(r.Context(), "video processing cancelled due to client disconnect",
+						"mediaID", media.ID,
+						"reason", r.Context().Err())
+					return // Don't send response if client disconnected
+				}
 
-		if mediaURL.Purpose == models.VideoWeb {
-			cachedPath = path.Join(utils.MediaCachePath(), strconv.Itoa(int(media.AlbumID)), strconv.Itoa(int(mediaURL.MediaID)), mediaURL.MediaName)
+				log.Error(r.Context(), "processing video not found in cache:",
```
googollee: In a structed log, it's not a one line. change the message to `processing video not found in cache`, no `:` in the suffix. See [slog details](https://pkg.go.dev/log/slog#pkg-overview)

api/routes/videos.go
kkovaletp: @googollee, this is the place, it is called. The function's code is between lines 24 and 129. 
This is the main code of this module, and it indeed handles HTTP requests for videos.
You are probably confused by the function, being part of the call params:
```
func(albumID, mediaID int, filename string) string {
	return path.Join(utils.MediaCachePath(), strconv.Itoa(albumID), strconv.Itoa(mediaID), filename)
},
```

api/routes/authenticate_routes.go
kkovaletp: Ok, I'll comment `log.debug()` calls in the code for now and hope that once we have log level configuration functionality, we can get back and enable them

api/utils/media_cache.go
kkovaletp: @googollee, the bot decided to resolve this thread before you had a chance to look at the result, so I unresolved it back to let you review the result

api/routes/videos.go
kkovaletp: @googollee, and again: the bot has resolved the conversation thread before you had a chance to look at the result, so I reverted it back to let you review the result.

api/routes/videos_test.go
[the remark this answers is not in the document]
kkovaletp: ok, I'll not touch this part for now, as I assume it will cause many more code changes across the product and make this PR even bigger and more complex.

api/routes/videos_test.go
```diff
@@ -0,0 +1,471 @@
...
+	original := utils.GetTestCachePath()
+	utils.ConfigureTestCache(tempPath)
+	return func() {
+		utils.ConfigureTestCache(original)
+	}
+}
+
+// mockProcessSingleMedia replaces scanner.ProcessSingleMedia with a mock function during tests
+// and returns a function to restore the original implementation
+var originalProcessSingleMedia = processSingleMediaFn
```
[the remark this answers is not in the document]
kkovaletp: our tests are not ready to run in parallel, no need for this code change for now

api/routes/videos.go
kkovaletp: Done

api/routes/videos.go
kkovaletp: Thanks, I already removed the `:` in some log messages, but missed it in several other ones.

api/routes/videos.go
kkovaletp: Removed the sanitizing code. Anyway, it impacted only the corresponding log message.

api/routes/videos.go
```diff
@@ -1,72 +1,136 @@
...
+		return
+	}
+
+	if len(mediaURLs) > 1 {
+		log.Warn(r.Context(), "Multiple video web URLs found",
+			"name", mediaName,
+			"count", len(mediaURLs),
+			"using", mediaURLs[0],
+		)
+	}
```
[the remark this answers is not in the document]
kkovaletp: We trust the data in our DB. Sanitizing should be done at the DB write places.

api/routes/videos.go
```diff
@@ -1,72 +1,136 @@
...
+	mediaName string,
+	authenticateFn func(*models.Media, *gorm.DB, *http.Request) (bool, string, int, error),
+	getCachePathFn func(albumID, mediaID int, filename string) string,
+) {
+	var mediaURLs []models.MediaURL
+	if err := db.Model(&models.MediaURL{}).
+		Preload("Media").
+		Where("media_urls.media_name = ? AND media_urls.purpose = ?", mediaName, models.VideoWeb).
+		Find(&mediaURLs).
+		Error; err != nil || len(mediaURLs) == 0 || mediaURLs[0].Media == nil {
```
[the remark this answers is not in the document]
kkovaletp: Having more than 1 row from this query is an exceptional case, and we have a code to detect and log it, so it is better than hiding this case in the DB and blindly taking only 1 row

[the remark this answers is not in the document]
kkovaletp: Added the `Order("created_at DESC")` to get the newest row 1st

api/utils/media_cache.go
googollee: Yes, I checked before. Looks good.

api/routes/videos.go
googollee: Hmmm, I didn't say that clearly, sorry.

I mean:

```
func generateCacheFilename(albumID, mediaID int, filename string) string {
	return path.Join(utils.MediaCachePath(), strconv.Itoa(albumID), strconv.Itoa(mediaID), filename)
}
```

and 

```
handleVideoRequest(w, r, db, mediaName, authenticateMedia, generateCacheFilename)
```

api/routes/authenticate_routes.go
googollee: You can keep those `log.Debug()`. It won't break anything, but just no good way to change the log level to `Debug` now.

api/routes/videos.go
googollee: It looks good.

api/routes/authenticate_routes.go
kkovaletp: As far as I understand the current logger implementation, if I keep them uncommented, those Debug log messages will be part of the log on a regular basis, as there is no log level configuration - all levels are active.
If my understanding is correct, having those messages logged constantly would be a serious security issue, as they contain sensitive information about the authentication process.

api/routes/videos.go
kkovaletp: Ok, now I got it: to extract this unnamed function and call it by name in the parameter.
Done.

googollee (approved):

PR #1206, opened 2025-05-16
Drop linux/arm/v7

googollee:
Since the latest release of [jellyfin-ffmpeg](https://github.com/jellyfin/jellyfin-ffmpeg/releases/tag/v7.1.1-3) doesn't support `armhf`, we can't support `arm/v7` with the full feature. Just drop the platform.

## Summary by CodeRabbit

- **Chores**
	- Updated build and dependency workflows to remove support for the linux/arm/v7 platform. Future builds will target only linux/amd64 and linux/arm64 architectures.


kkovaletp (approved):
Sad to do that, but this is our critical dependency, so we don't have other options.
Approving, but can you please all a note about this to the Readme, like we did for ARM v6 some time ago?

Issue #1208, opened 2025-05-21
Refactor the Scan job code


kkovaletp:
As discussed in the PR #1204, here is the list of high-level requirements for the Scan Job code refactoring:
- Investigate which parts (tasks) of the job don't run in parallel and make them run in parallel if possible (for example, the initial folder scan task)
- Make sure that all the code uses the new Backend's logger instead of throwing the messages to the console directly or using other methods
- Enhance the log messages with the worker context and processed item to make them useful and descriptive
- Make sure that the task assignment to workers strategy is optimal, so that all workers are busy with similar loads and work effectively
- Investigate and find the most optimal way to run image processing workers without spinning a new process per each media

kkovaletp:
I've asked ChatGPT about different task management libs and approaches for GO and he suggested the `ants`, as the most compatible with Photoview solution, but also listed other more powerful alternatives, explaining their pros and cons:
https://chatgpt.com/share/68304426-ee00-800c-bacc-3dc0ba73e663

googollee:
Most of those libraries require some dependency. The background task of photoview is not such complex and it's no need to over engineering it.

`ants` is a goroutine pool, which is designed for >1k goroutine. I don't think it fits photoview.

My idea is like:

- Launch N goroutines in the background and init Magick instance in each goroutine.
  - One goroutine reuses Magick instance when handling new jobs.
- Queue sends jobs to those goroutines with proper parameters.
- Another queue receives results from those goroutine and update database according to results.

For simplicity, we could store the queue in memory. I'm also thinking about storing the queue in DB, with a new table. Storing in DB helps to resume scan when server reboots, but it's more complex.

kkovaletp:
> I'm also thinking about storing the queue in DB, with a new table. Storing in DB helps to resume scan when server reboots, but it's more complex.

I don't know how intensive the IO would be for such a queue from our several workers. If there is intensive enough communication, the DB backend might be a bottleneck from a performance perspective. That is why, typically, Redis is used.
Also, I know that PostgreSQL is powerful enough to handle this role, but what about MariaDB and especially SQLite? Will they handle this role without slowing down the scan or the entire app?

googollee:
I don't think the bottleneck is on DB side. There is only <10 requests since photoview is single server. The bottleneck should be at reformat/thumbnail process. Any DB should be OK here.

googollee:
I did a quick test with MagickWand:

```
[...]

PR #1210, opened 2025-05-23
Fix a race condition in the Periodic Scanner module

kkovaletp:
This PR was initiated by the issue, detected by the Coverity scanner in my forked repo:

> CID 1588234: Data race condition (GUARDED_BY_VIOLATION)
> missing_lock: Accessing mainPeriodicScanner.ticker without holding lock periodic_scanner.periodicScanner.mutex. Elsewhere, periodic_scanner.periodicScanner.ticker is written to with periodic_scanner.periodicScanner.mutex held 1 out of 1 times.

### DIFF
--- a/api/scanner/periodic_scanner/periodic_scanner.go
+++ b/api/scanner/periodic_scanner/periodic_scanner.go
@@ -76,15 +76,19 @@ func ChangePeriodicScanInterval(duration time.Duration) {
 func scanIntervalRunner() {
 	for {
 		log.Print("Scan interval runner: Waiting for signal")
-		if mainPeriodicScanner.ticker != nil {
+		mainPeriodicScanner.mutex.Lock()
+		ticker := mainPeriodicScanner.ticker
+		if ticker != nil {
+			mainPeriodicScanner.mutex.Unlock()
 			select {
 			case <-mainPeriodicScanner.ticker_changed:
 				log.Print("Scan interval runner: New ticker detected")
-			case <-mainPeriodicScanner.ticker.C:
+			case <-ticker.C:
 				log.Print("Scan interval runner: Starting periodic scan")
 				scanner_queue.AddAllToQueue()
 			}
 		} else {
+			mainPeriodicScanner.mutex.Unlock()
 			<-mainPeriodicScanner.ticker_changed
 			log.Print("Scan interval runner: New ticker detected")
 		}


api/scanner/periodic_scanner/periodic_scanner.go
```diff
@@ -73,20 +75,65 @@ func ChangePeriodicScanInterval(duration time.Duration) {
...
 		if mainPeriodicScanner.ticker != nil {
+			mainPeriodicScanner.ticker.Stop()
+			mainPeriodicScanner.ticker = nil
+		}
+		mainPeriodicScanner.mutex.Unlock()
+
+		// Reset the global scanner
+		mainPeriodicScanner = nil
+	}
+}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, thanks for finding and reporting this issue! Please carefully analyze the codebase and provide the complete code replacements for all cases and functions affected by this issue

api/scanner/periodic_scanner/periodic_scanner.go
kkovaletp, line 1: @coderabbitai, we've made significant changes in this code file by fixing several potential race conditions and implementing a better shutdown flow. Now it is a good time to check the coverage of the complete final code of this module and implement missing unit tests.
Please carefully analyze this code file in the context of the project, then please analyze all the existing unit tests of the backend code together with the testing framework configuration and all related data in the context of the project. Once done, please use all your relevant learnings from our previous PRs regarding successful test development and all relevant knowledge from your knowledge base and the web. Finally, please suggest the exact code replacements to extend the existing project unit tests and cover all missing cases in this code file in a consistent and robust way, making sure that these new tests cover the behavior of the code and don't dive too deep into the internal implementation. When preparing the reply message, make sure to include all the requested code blocks and integration instructions if needed. Please don't offer or create any PRs, issues, or commits.

kkovaletp: @coderabbitai, thanks for the testing code and the integration options. I'd go with option 1 (adding the queue to the struct). I did the code changes in the `periodic_scanner.go`, should I change the testing code somehow in order to use this new injection? Should I modify the product code to use this new injection as well? Does it give any benefits for the product code?
My 2nd topic to discuss is the `scannerCalled` variable in the "runner with active ticker" test case: VScode reports that it is declared, but never used. Since we use the injection, should we rewrite this and optionally some other tests to better cover the code?

kkovaletp: @coderabbitai, you've replied to me with the error "CodeRabbit did not find changes to make". Please make sure to overcome this issue and properly prepare the reply with all the requested info, then please reply it to me, as requested.

kkovaletp: @coderabbitai, are you sure that these lock and unlock lines should be in the `InitializePeriodicScannerWithQueue` function as part of the setting initial interval code? This function already has this lock set at the beginning and unlock in the `defer`.
Please carefully review your last suggestion and fix all possible inconsistencies and defects, like this one, or the reason why the message "INFO Periodic scan interval changed: %s !BADKEY=5m0s" contains the "BADKEY" word in the console output.

api/scanner/periodic_scanner/periodic_scanner_test.go
```diff
@@ -0,0 +1,299 @@
...
+		mockQueue := &MockScannerQueue{}
+
+		err := createTestSiteInfo(db, 300)
+		assert.NoError(t, err)
+
+		err = InitializePeriodicScannerWithQueue(db, mockQueue)
+		assert.NoError(t, err)
+
+		// Wait a bit for initialization to complete
+		time.Sleep(50 * time.Millisecond)
```
kkovaletp, line 93: @coderabbitai, you've raised the next nitpick suggestion regarding this line:

> Consider replacing sleep with synchronization for more reliable tests.
> The hardcoded sleep for initialization completion could make tests flaky in slower environments. > Think of it like waiting for a delivery - instead of guessing how long it takes, it's better to get a confirmation when it arrives.
> Consider using a synchronization mechanism instead:

```
assert.Eventually(t, func() bool {
			mainPeriodicScannerMutex.Lock()
			defer mainPeriodicScannerMutex.Unlock()
			return mainPeriodicScanner != nil
		}, time.Second, 10*time.Millisecond)
```

Please carefully validate this suggestion: is it really better for this particular test and other tests in this file? Will it work better? Isn't it more risky to cause deadlocks by using a mutex? Do you see any other potential issues with this implementation?

api/scanner/periodic_scanner/periodic_scanner_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+		// Use a channel to synchronize and count calls
+		callChan := make(chan struct{}, 5) // Buffer for multiple calls
+		mockQueue.On("AddAllToQueue").Return(nil).Run(func(args mock.Arguments) {
+			select {
+			case callChan <- struct{}{}:
+			default:
+				// Channel full, but that's okay
+			}
+		}).Maybe()
```
kkovaletp, line 156: @coderabbitai, you've raised the next nitpick suggestion regarding this line:

> Consider a more reliable test synchronization
> The test uses `Maybe()` on mock expectations combined with buffered channels for synchronization. While functional, this pattern can make tests less predictable:
> `Maybe()` means the expectation might not be called, but you still assert expectations 
> Sleep-based synchronization can be flaky across different environments
> Consider using more deterministic synchronization:

```
mockQueue.On("AddAllToQueue").Return(nil).Times(1).Run(func(args mock.Arguments) {
 			select {
 			case callChan <- struct{}{}:
 			default:
 				// Channel full, but that's okay
 			}
		})
```

Please carefully validate this suggestion: Is it really better for this particular test? Will it work more reliably in the scope of the test suite execution? We switched from the `Times(1)` to `Maybe()` on purpose to fix unpredictable run-time behavior. Is it really better to get back with this implementation?

api/scanner/periodic_scanner/periodic_scanner_test.go
```diff
@@ -0,0 +1,299 @@
...
+		select {
+		case <-runnerDone:
+			// Success
+		case <-time.After(500 * time.Millisecond):
+			t.Fatal("Runner did not exit after ticker change")
+		}
+
+		// Test passes if no deadlock occurs
+		mockQueue.AssertExpectations(t)
+	})
```
kkovaletp, line 283: @coderabbitai, you've raised the next nitpick suggestion regarding this line:

> Enhance ticker change test realism
> The test starts with `ticker: nil`, which may not fully exercise the real-world scenario where ticker changes occur on an active ticker. Consider testing with an initial ticker that gets changed.

```
ps := &periodicScanner{
			ticker:         time.NewTicker(100 * time.Millisecond), // Start with ticker
 			ticker_changed: make(chan bool, 1),
 			done:           make(chan struct{}),
 			mutex:          &sync.Mutex{},
 			scannerQueue:   mockQueue,
 		}
```

> Don't forget to stop the initial ticker in cleanup:

```
if ps.ticker != nil {
			ps.ticker.Stop()
		}
```

Please carefully validate this suggestion: Is it really better for this particular test? Will it work more reliably in the scope of the test suite execution? Do you see some potential risks or issues with this implementation?

api/scanner/periodic_scanner/periodic_scanner.go
```diff
@@ -42,51 +58,119 @@ func InitializePeriodicScanner(db *gorm.DB) error {
...
-				scanner_queue.AddAllToQueue()
+			case <-ps.done:
+				log.Info("Scan interval runner: Shutting down")
+				return
+			case <-ps.ticker_changed:
+				log.Info("Scan interval runner: New ticker detected")
+			case <-ticker.C:
+				log.Info("Scan interval runner: Starting periodic scan")
+				if err := ps.scannerQueue.AddAllToQueue(); err != nil {
+					log.Error("Scan interval runner: Failed to add all users to queue", err)
```
kkovaletp, line 163: @coderabbitai, you've raised the next nitpick suggestion regarding this line:

> Consider improving error logging format
> The error logging call could benefit from using the log package's formatting capabilities for better structured logging.
> Consider using a format that separates the message from the error:

```
log.Error("Scan interval runner: Failed to add all users to queue: %v", err)
```

Please carefully validate this suggestion: is it really better for this particular code in this file? Will it work better?
Originally, it was `log.Error("Scan interval runner: Failed to add all users to queue: %s", err)` and caused the `BADKEY` error at the execution time. The function accepts a string as the 1st parameter and any args after it. What would be the best way to pass the error variable or other context variables in such cases?

api/scanner/periodic_scanner/periodic_scanner_test.go
kkovaletp: @coderabbitai, thanks. Should I apply the same change to other tests in this file? To which ones? Please carefully verify each test you're going to list in your reply, to not introducing a regression.

api/scanner/periodic_scanner/periodic_scanner_test.go
kkovaletp, line 1: @coderabbitai, please iterate over all verifications in this file and suggest failure message strings, which would clearly explain the reason for the failure if it happens.
Please give me the list of code lines to modify, with the line number and the exact code replacement line per each affected line of code.

kkovaletp:
@googollee, please review this PR with the next part of my fixes from the Coverity backlog.
I tested this branch manually to make sure that my edits didn't break end-to-end scanner flow - it works fine for my testing instance.

api/scanner/periodic_scanner/periodic_scanner.go
```diff
@@ -1,26 +1,39 @@
...
 type periodicScanner struct {
 	ticker         *time.Ticker
 	ticker_changed chan bool
+	done           chan struct{}
 	mutex          *sync.Mutex
 	db             *gorm.DB
+	scannerQueue   ScannerQueue
 }
 
+var mainPeriodicScannerMutex sync.Mutex
```
googollee, line 33: I know it's the easy way to fix race condition. In my opinion, I don't like a global locker. It's easy to lock with IO actions, which causes deadlocking or inter-locking when IO can't finish on time.

How urgent do you think to finish this problem? I'll cover this with #1208.

api/server.go
```diff
@@ -112,7 +122,48 @@ func main() {
 
 	}
 
-	log.Panic(http.ListenAndServe(apiListenURL.Host, handlers.CompressHandler(rootRouter)))
+	srv := &http.Server{
+		Addr:    apiListenURL.Host,
+		Handler: handlers.CompressHandler(rootRouter),
+	}
+
+	go func() {
```
googollee, line 130: This makes more complex than necessary.

You can call `svr.Shutdown()` at the end of shutting down function in `setupGracefulShutdown`, and keep other parts as previous. Then when a signal receives, the shutting function in `setupGracefulShutdown` will close queue and scanner, and shut down the server. Then the server in the main goroutine will quit gracefully.

api/scanner/periodic_scanner/periodic_scanner.go
```diff
@@ -42,51 +58,119 @@ func InitializePeriodicScanner(db *gorm.DB) error {
...
 
-	ChangePeriodicScanInterval(scanInterval)
 	return nil
 }
 
+func InitializePeriodicScanner(db *gorm.DB) error {
+	return InitializePeriodicScannerWithQueue(db, &RealScannerQueue{})
+}
+
 func ChangePeriodicScanInterval(duration time.Duration) {
```
googollee, line 91: No code calls `ChangePeriodicScanInterval` now. Is that a mistake?

api/scanner/periodic_scanner/periodic_scanner.go
```diff
@@ -1,26 +1,39 @@
...
+
+func (r *RealScannerQueue) AddAllToQueue() error {
+	return scanner_queue.AddAllToQueue()
+}
+
 type periodicScanner struct {
 	ticker         *time.Ticker
 	ticker_changed chan bool
+	done           chan struct{}
 	mutex          *sync.Mutex
```
googollee, line 28: As refactoring, could you rename this `mutex` as `tickerLocker` and put it together with `ticker`? From the code, this locker only protects `ticker` when the interval is changing.

Also change the type to `sync.Mutex` instead of `*sync.Mutex`.

api/scanner/periodic_scanner/periodic_scanner.go
googollee: From the code, we could remove this mutex. The `mainPeriodicScanner` should be created when the server launches, and it should not be changed during running. Of course the content of `mainPeriodicScanner` may be changed, and that should be protected by the inner mutex.

api/scanner/periodic_scanner/periodic_scanner.go
```diff
@@ -42,51 +58,119 @@ func InitializePeriodicScanner(db *gorm.DB) error {
...
 	} else {
-		log.Print("Periodic scan interval changed: disabled")
+		log.Info("Periodic scan interval changed: disabled")
 	}
 
-	{
-		mainPeriodicScanner.mutex.Lock()
-		defer mainPeriodicScanner.mutex.Unlock()
+	mainPeriodicScannerMutex.Lock()
+	scanner := mainPeriodicScanner
```
googollee, line 101: Remove mutex around it, and change `mainPeriodicScanner` directly below instead of copying to `scanner`.

api/scanner/periodic_scanner/periodic_scanner.go
```diff
@@ -42,51 +58,119 @@ func InitializePeriodicScanner(db *gorm.DB) error {
...
+		case scanner.ticker_changed <- true:
+		default:
+			// Channel might be full, but that's okay
+		}
+	}
+}
+
+// ShutdownPeriodicScanner gracefully shuts down the periodic scanner
+func ShutdownPeriodicScanner() {
+	mainPeriodicScannerMutex.Lock()
```
googollee, line 123: No need to lock here.

api/scanner/periodic_scanner/periodic_scanner.go
kkovaletp: It is called from this place:

https://github.com/kkovaletp/photoview/blob/66d672e96725626e250a6946af54d030a27aafea/api/graphql/resolvers/scanner.go#L74

Also, I've tested the case of changing the scan interval from the GUI, and I saw the changes applied on the backend

api/server.go
kkovaletp: @googollee, sorry, but I didn't get how exactly I need to modify the `server.go` file in order to implement your suggestion and keep the code consistent. You comment on the code line to run the HTTP server in a GO routine, but suggest changes for the server shutdown process - this confuses me(
Here are threads with the CR bot, from which I implement the current changes:
https://github.com/kkovaletp/photoview/pull/478#discussion_r2105181531 - from this message till the end of the thread.
https://github.com/kkovaletp/photoview/pull/478#discussion_r2105283774 - the complete thread.
It looks like there were other related conversations with the bot, but I cannot find them by search in both PRs now... Hope that these 2 clarify the intentions for the changes enough.

Could you please review those threads and instruct me on what to do with the `server.go` code?

api/scanner/periodic_scanner/periodic_scanner.go
```diff
@@ -42,51 +55,113 @@ func InitializePeriodicScanner(db *gorm.DB) error {
...
 		if mainPeriodicScanner.ticker != nil {
+			mainPeriodicScanner.ticker.Stop()
+			mainPeriodicScanner.ticker = nil
+		}
+		mainPeriodicScanner.tickerLocker.Unlock()
+
+		// Reset the global scanner
+		mainPeriodicScanner = nil
+	}
+}
```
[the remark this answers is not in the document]
kkovaletp: @googollee, please confirm that this is ok

api/scanner/periodic_scanner/periodic_scanner.go
kkovaletp: Addressed

api/server.go
googollee: The code looks like:

```go
func setupGracefulShutdown(server *http.Server) {
	c := make(chan os.Signal, 1)
	signal.Notify(c, os.Interrupt, syscall.SIGTERM)

	go func() {
		<-c
		log.Println("Shutting down Photoview...")

		// Shutdown scanners in correct order
		periodic_scanner.ShutdownPeriodicScanner()
		scanner_queue.CloseScannerQueue()

		server.Shutdown()
	}()
}

func main() {
	// ...
	svc := &http.Server{...}
        setupGracefulShutdown(svc)

        if err := srv.Shutdown(shutdownCtx); err != nil {
		log.Printf("HTTP server shutdown error: %s", err)
	}
	// ...
}
```

api/scanner/periodic_scanner/periodic_scanner.go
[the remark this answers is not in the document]
googollee: It's no need to set `mainPeriodicScanner` as nil, since it's a global instance. The `mainPeriodicScanner.done` is closed, and it won't perform any actions if it's called again.

api/server.go
kkovaletp: @googollee, thanks! I tried to merge your code into the branch and tested the server start, basic functionality, and shutdown with the new code.
Please check whether I implemented your suggestion in the correct way, or if there is still something to do.

api/scanner/periodic_scanner/periodic_scanner.go
```diff
@@ -42,51 +55,113 @@
...
 		}
 
 		mainPeriodicScanner.ticker = newTicker
-		mainPeriodicScanner.ticker_changed <- true
+		select {
+		case mainPeriodicScanner.ticker_changed <- true:
+		default:
+			// Channel might be full, but that's okay
+		}
 	}
```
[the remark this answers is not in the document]
kkovaletp: @googollee, I'm still waiting for your confirmation before closing this suggestion ;-)

api/server.go
```diff
@@ -112,7 +120,48 @@ func main() {
...
+	setupGracefulShutdown(srv, ctx)
+
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Panicf("HTTP server failed: %s", err)
+		}
+	}()
+
+	// Wait for shutdown signal
+	<-ctx.Done()
```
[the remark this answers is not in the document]
kkovaletp: @googollee, the bot doesn't like my latest code changes, based on your suggestion... Maybe it is not consistent and I need to do some more edits in the code...
BTW, it says that the server will hang at shutdown, but in reality, it doesn't happen: the server shuts down correctly. Maybe I didn't reproduce the exact case to make it hang...

Anyway, please confirm that this is fine, and I can close this thread.

api/server.go
googollee: It's no need to create a context in the main goroutine. The context of waiting for shutdown is another progress.

```
func setupGracefulShutdown(server *http.Server) {
	c := make(chan os.Signal, 1)
	signal.Notify(c, os.Interrupt, syscall.SIGTERM)

	go func() {
		<-c
		log.Println("Shutting down Photoview...")
		ctx, cancel := context.WithTimeout(context.Background(), time.Minute) // Wait for 1m to shutdown
		defer cancel()

		// Shutdown scanners in correct order
		periodic_scanner.ShutdownPeriodicScanner()
		scanner_queue.CloseScannerQueue()

		server.Shutdown(ctx)
	}()
}
```

Also it's not necessary to shut down server twice.

api/server.go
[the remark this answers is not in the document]
googollee: Please check my last comment.

api/server.go
kkovaletp: @googollee, thanks. I hope that I have a better understanding of your suggestion.
I've implemented and manually tested the changes: everything works fine now.

I tried to remove the next lines from the main function too:
```
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
...
	// Wait for shutdown signal
	<-ctx.Done()
```
But if I do that, the server init process falls into an infinite loop, so I had to get them back to make it stop after the init is done and serve incoming requests until the shutdown signal comes.

googollee: How do you write the `server.go` now? I see `srv.ListenAndServe()` is running in a goroutine currently, which should not.

The main process runs `srv.ListenAndServe()` directly and stops at here. When `setupGracefulShutdown()` calls `Shutdown()`, it should return and the process quits.

kkovaletp: @googollee, please read the https://github.com/kkovaletp/photoview/pull/478#discussion_r2105283774 conversation thread - this is the one where the bot suggested implementing the HTTP server in a goroutine.
Hope that it answers your question

googollee: It's not necessary to use another goroutine in the main goroutine. The main task is `srv.ListenAndServe()`, and just block the main goroutine on it naturally. The timeout protection is implemented in `setupGracefulShutdown()`.

kkovaletp: @googollee, I've removed the goroutine, and it also allowed me to drop the other code.
Please check. I think that now it should be ok)

kkovaletp: After your latest merged PR with the refactored logger, I had to update my code for the `periodic_scanner` to reflect the changes. You said that the is a queue and worker context:

> For example, in the queue, there is a queue context. The queue creates a new context from the queue context for each worker, and still uses the queue context for logging information about the queue.

But I didn't find it in the `mainPeriodicScanner` itself or its nested objects, so I set `nil` for all logging calls. 
If this is incorrect, please tell me where I should get the correct context variable for the logging in that file.

googollee: That's much clearer than before. Thanks for the work. It's fine to use `nil` for logging now. I'm working on a new queue.

api/scanner/periodic_scanner/periodic_scanner.go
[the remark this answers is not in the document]
googollee: Well, it's possible in the `periodic_scanner` without looking into `server.go`. I don't have a good way to fix it now. It's good to add another mutex to lock `mainPeriodicScanner` now. I'm sorry of letting you remove that before.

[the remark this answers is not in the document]
kkovaletp: @coderabbitai, after your suggestion, there were many code changes in the branch. So, please reevaluate the latest commit in the PR branch and suggest the most reliable and consistent solution for this issue. It is fine if it were the one originally proposed, if you're sure that it meets the requirements in the best way. I just want to double-check that it is still the best option.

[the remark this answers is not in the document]
kkovaletp: @googollee, ok, I've implemented the solution suggested by the bot, which was previously reverted. Does it solve the problem for now, or would you like me to implement something else?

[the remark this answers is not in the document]
kkovaletp: @googollee 
Also, I've added back the additional mutex, which protects the `scanner` variable setting, as it was before.
Would you like to return this mutex to other places in this file as well, or would having it only here be enough?

api/scanner/periodic_scanner/periodic_scanner.go
```diff
@@ -42,51 +56,116 @@ func InitializePeriodicScanner(db *gorm.DB) error {
...
+		scannerQueue:   queue,
 	}
 
-	go scanIntervalRunner()
+	go mainPeriodicScanner.scanIntervalRunner()
+
+	var newTicker *time.Ticker = nil
+	if scanInterval > 0 {
+		newTicker = time.NewTicker(scanInterval)
+		log.Info(nil, "Periodic scan interval changed: "+scanInterval.String())
```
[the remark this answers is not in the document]
kkovaletp: This will be done in another PR after the new code for the queue is merged

api/scanner/periodic_scanner/periodic_scanner.go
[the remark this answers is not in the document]
googollee: It's OK to put it here.

PR #1216, opened 2025-06-06
Update log for passing attrs with context.

googollee:
This improves the log library.

## Summary by CodeRabbit

- **New Features**
  - Introduced context-aware logging, allowing log messages to include additional contextual information for improved traceability.

- **Bug Fixes**
  - Logging functions now consistently require a context parameter, ensuring structured and context-rich logs throughout the application.

- **Tests**
  - Added tests to verify correct logging behavior with and without context attributes.

- **Chores**
  - Updated `.gitignore` to exclude coverage files from version control.


api/log/context.go
```diff
@@ -0,0 +1,35 @@
...
+		return defaultLogger
+	}
+
+	logger := ctx.Value(loggerKey)
+	if logger == nil {
+		return defaultLogger
+	}
+
+	return logger.(*slog.Logger)
+}
```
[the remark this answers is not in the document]
googollee: Fixed.

api/log/context_test.go
```diff
@@ -0,0 +1,70 @@
...
+	})
+
+	logger := slog.New(handler)
+	oldLogger := defaultLogger
+	defaultLogger = logger
+
+	return &output, func() {
+		defaultLogger = oldLogger
+	}
+}
```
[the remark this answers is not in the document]
googollee: It's no necessary here. The global value should not be changed in the real code. Since this test is not running in parallel, it's fine to keep it simple.

api/log/cover.cov
kkovaletp, line 1: @googollee, I guess this file was committed to the PR by mistake. If that is the case, please remove it. If not, please tell me why we need it in the repo.

kkovaletp:
@googollee, I see that you've updated logging calls in several code modules by adding the `nil` context, and there are no examples of logging with the context.
Can you please explain the benefits of the new logger with the context parameter and the rules where to call it with the context vs where to call it with the `nil` context? Can you please update several logging calls in the code, providing the context to illustrate this PR with the examples of both use cases?

api/log/cover.cov
googollee: It's a mistake. Removed.

googollee:
There are [examples](https://github.com/photoview/photoview/pull/1216/files#diff-4e6f23cb0aefa98c74a2c1267762b4e9409872a855414692e85df4c88d5a3c82R48) with context `nil`. It's compatible with the code without a context. If there is no context in the current code, creating a background context is meaningless.

In [the test](https://github.com/photoview/photoview/pull/1216/files#diff-4e6f23cb0aefa98c74a2c1267762b4e9409872a855414692e85df4c88d5a3c82R60), `ctx1` is a parent context with some attributes. For a child process, it creates `ctx2` with additional attributes and uses it for logging. When the child process is finished, the parent continues to use `ctx1` for logging, without attributes in `ctx2`.

For example, in the queue, there is a queue context. The queue creates a new context from the queue context for each worker, and still uses the queue context for logging information about the queue.

.gitignore
```diff
@@ -23,6 +23,7 @@ node_modules/
 /ui/coverage
 /api/coverage.txt
 /api/test-*-report.xml
+cover.cov
```
kkovaletp: @googollee, Isn't the `*.cov` better pattern? It would cover coverage reports with any names.

googollee: Sure. Updated.

kkovaletp (approved):
Approving

PR #1217, opened 2025-06-10
[WIP] The new scanner queue

googollee:
Part of #1208.

Working in progress. I'm still working on unit tests and integrating with the main server.

@kkovaletp @viktorstrate Please take a look of the new design of the scanner queue. I want to throw the design early for review. If there is any unclear, please let me know. Thanks.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new job queue system for scanning user albums, supporting concurrent processing and dynamic worker scaling.
  - Added the ability to enqueue scan jobs for all users or specific users.
  - Implemented periodic automatic scanning based on configurable intervals.
  - Improved error handling and reporting for batch job processing.
- **Chores**
  - Enhanced internal structure for managing and processing scan jobs efficiently.

### DIFF
--- a/api/scanner/queue/queue.go
+++ b/api/scanner/queue/queue.go
@@ -62,23 +62,25 @@ func NewQueue(db *gorm.DB) (*Queue, error) {
 }
 
 func (q *Queue) Close() {
+	defer q.waitWorkers.Wait()
+
 	q.trigger.Stop()
 	close(q.done)
 }
 
 func (q *Queue) UpdateScanInterval(newInterval time.Duration) error {
 	if newInterval < 0 {
-		return fmt.Errorf("ivalid periodic scan interval(%v): must >=0", newnterval)
+		return fmt.Errorf("ivalid periodic scan interval(%v): must >=0", newInterval)
 	}
 
 	if newInterval == 0 {
-		trigger.Stop()
+		q.trigger.Stop()
 		return nil
-	}	
+	}
 
 	q.trigger.Reset(newInterval)
 	return nil
-	
+}
 
 func (q *Queue) ResizeWorkers(newMax int) {
 	if newMax < 0 {
@@ -114,10 +116,7 @@ func (q *Queue) ResizeWorkers(newMax int) {
 }
 
 func (q *Queue) Run() {
-	defer func() {
-		q.ResizeWorkers(0)
-		q.waitWorkers.Wait()
-	}()
+	defer q.ResizeWorkers(0)
 
 MAIN:
 	for {
--- a/api/scanner/queue/queue.go
+++ b/api/scanner/queue/queue.go
@@ -1,0 +1,184 @@
+package queue
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"sync"
+	"time"
+
+	"github.com/photoview/photoview/api/graphql/models"
+	"github.com/photoview/photoview/api/log"
+	"github.com/photoview/photoview/api/scanner"
+	"github.com/photoview/photoview/api/scanner/scanner_cache"
+	"gorm.io/gorm"
+)
+
+type Queue struct {
+	db          *gorm.DB
+	waitWorkers sync.WaitGroup
+	ctx         context.Context
+	input       chan []Job
+	handout     chan Job
+	done        chan struct{}
+	trigger     *time.Ticker
+	backlog     []Job
+
+	workers   []*worker
+	workersMu sync.Mutex
+}
+
+func NewQueue(db *gorm.DB) (*Queue, error) {
+	var siteInfo models.SiteInfo
+	if err := db.First(&siteInfo).Error; err != nil {
+		return nil, fmt.Errorf("can't get site info: %w", err)
+	}
+
+	interval := time.Duration(siteInfo.PeriodicScanInterval) * time.Second
+
+	ret := &Queue{
+		db:      db,
+		ctx:     log.WithAttrs(context.Background(), "process", "queue"),
+		input:   make(chan []Job),
+		handout: make(chan Job),
+		done:    make(chan struct{}),
+		trigger: time.NewTicker(interval),
+	}
+
+	ret.UpdateWorkers(siteInfo.ConcurrentWorkers)
+
+	return ret, nil
+}
+
+func (q *Queue) Close() {
+	q.trigger.Stop()
+	close(q.done)
+}
+
+func (q *Queue) UpdateInterval(newInterval time.Duration) {
+	q.trigger.Reset(newInterval)
+}
+
+func (q *Queue) UpdateWorkers(newMax int) {
+	q.workersMu.Lock()
+	defer q.workersMu.Unlock()
+
+	if len(q.workers) == newMax {
+		return
+	}
+
+	if len(q.workers) < newMax {
+		closing := q.workers[newMax:]
+		q.workers = q.workers[:newMax]
+
+		for _, worker := range closing {
+			worker.Close()
+		}
+
+		return
+	}
+
+	// len(q.workers) > newMax
+	q.workers = slices.Grow(q.workers, newMax-len(q.workers))
+	for len(q.workers) < newMax {
+		worker := newWorker(log.WithAttrs(q.ctx, "worker_id", len(q.workers)), q.db, q.handout, &q.waitWorkers)
+		q.waitWorkers.Add(1)
+		go worker.Run()
+		q.workers = append(q.workers, worker)
+	}
+}
+
+func (q *Queue) Run() {
+	defer func() {
+		q.UpdateWorkers(0)
+		q.waitWorkers.Wait()
+	}()
+
+	for {
+		if len(q.backlog) > 0 {
+			select {
+			case q.handout <- q.backlog[0]:
+				q.backlog = q.backlog[1:]
+
+			case input := <-q.input:
+				q.backlog = append(q.backlog, input...)
+			case <-q.trigger.C:
+				q.AddAllUsers(q.ctx)
+			case <-q.done:
+				break
+			}
+
+			continue
+		}
+
+		select {
+		case input := <-q.input:
+			q.backlog = append(q.backlog, input...)
+		case <-q.trigger.C:
+			q.AddAllUsers(q.ctx)
+		case <-q.done:
+			break
+		}
+	}
+}
+
+func (q *Queue) AddAllUsers(ctx context.Context) error {
+	var users []*models.User
+	if err := q.db.Find(&users).Error; err != nil {
+		return fmt.Errorf("get all users from database error: %w", err)
+	}
+
+	var jobs []Job
+	for _, user := range users {
+		job, err := q.findUserJobs(user)
+		if err != nil {
+			return fmt.Errorf("failed to add user (id: %d) for scanning: %w", user.ID, err)
+		}
+		jobs = append(jobs, job...)
+	}
+
+	select {
+	case q.input <- jobs:
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-q.done:
+		return fmt.Errorf("queue is closed")
+	}
+
+	return nil
+}
+
+func (q *Queue) AddUser(ctx context.Context, user *models.User) error {
+	jobs, err := q.findUserJobs(user)
+	if err != nil {
+		return fmt.Errorf("find albums for user (id: %d) error: %w", err)
+	}
+
+	select {
+	case q.input <- jobs:
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-q.done:
+		return fmt.Errorf("queue is closed")
+	}
+
+	return nil
+}
+
+func (q *Queue) findUserJobs(user *models.User) ([]Job, error) {
+	albumCache := scanner_cache.MakeAlbumCache()
+	albums, album_errors := scanner.FindAlbumsForUser(q.db, user, albumCache)
+	if len(album_errors) != 0 {
+		return nil, Errors(album_errors)
+	}
+
+	jobs := make([]Job, 0, len(albums))
+	for _, album := range albums {
+		jobs = append(jobs, Job{
+			album: album,
+			cache: albumCache,
+		})
+	}
+
+	return jobs, nil
+}
--- a/api/scanner/queue/queue.go
+++ b/api/scanner/queue/queue.go
@@ -70,7 +70,7 @@ func (q *Queue) Close() {
 
 func (q *Queue) UpdateScanInterval(newInterval time.Duration) error {
 	if newInterval < 0 {
-		return fmt.Errorf("ivalid periodic scan interval(%v): must >=0", newInterval)
+		return fmt.Errorf("invalid periodic scan interval(%v): must >=0", newInterval)
 	}
 
 	if newInterval == 0 {
--- a/api/scanner/queue/queue.go
+++ b/api/scanner/queue/queue.go
@@ -108,7 +108,12 @@ MAIN:
 			case input := <-q.input:
 				q.backlog = append(q.backlog, input...)
 			case <-q.trigger.C:
-				q.AddAllAlbums(q.ctx)
+				jobs, err := q.findAllAlbumsJobs()
+				if err != nil {
+					log.Error(q.ctx, "interval scan", "error", err)
+					continue
+				}
+				q.backlog = append(q.backlog, jobs...)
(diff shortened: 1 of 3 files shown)


api/scanner/queue/worker.go
```diff
@@ -0,0 +1,62 @@
...
+
+	for {
+		select {
+		case job := <-w.input:
+			w.processJob(job)
+		case <-w.done:
+			break
+		}
+	}
+}
```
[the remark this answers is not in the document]
googollee: The `break` here breaks the for loop but not `select`, and then quits the function. If using `return`, it may miss the logic after `for` loop, if there is any.

api/scanner/queue/worker.go
```diff
@@ -0,0 +1,62 @@
...
+	close(w.done)
+}
+
+func (w *worker) Run() {
+	defer w.parentWaiter.Done()
+
+	for {
+		select {
+		case job := <-w.input:
+			w.processJob(job)
```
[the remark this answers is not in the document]
googollee: `w.input` won't be closed. It's `w.done` to close the queue. It's not necessary to close `w.input` from the outside.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,184 @@
...
+		return fmt.Errorf("queue is closed")
+	}
+
+	return nil
+}
+
+func (q *Queue) AddUser(ctx context.Context, user *models.User) error {
+	jobs, err := q.findUserJobs(user)
+	if err != nil {
+		return fmt.Errorf("find albums for user (id: %d) error: %w", err)
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	// len(q.workers) > newMax
+	q.workers = slices.Grow(q.workers, newMax-len(q.workers))
+	for len(q.workers) < newMax {
+		worker := newWorker(log.WithAttrs(q.ctx, "worker_id", len(q.workers)), q.db, q.handout, &q.waitWorkers)
+		q.waitWorkers.Add(1)
+		go worker.Run()
+		q.workers = append(q.workers, worker)
+	}
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,184 @@
...
+	close(q.done)
+}
+
+func (q *Queue) UpdateInterval(newInterval time.Duration) {
+	q.trigger.Reset(newInterval)
+}
+
+func (q *Queue) UpdateWorkers(newMax int) {
+	q.workersMu.Lock()
+	defer q.workersMu.Unlock()
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,184 @@
...
+
+		select {
+		case input := <-q.input:
+			q.backlog = append(q.backlog, input...)
+		case <-q.trigger.C:
+			q.AddAllUsers(q.ctx)
+		case <-q.done:
+			break
+		}
+	}
```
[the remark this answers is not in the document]
googollee: Updated with breaking label now.

api/scanner/queue/worker.go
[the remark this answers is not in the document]
googollee: OK. You are right. I updated with breaking label now.

api/scanner/queue/worker.go
```diff
@@ -0,0 +1,62 @@
...
+		case <-w.done:
+			break
+		}
+	}
+}
+
+func (w *worker) processJob(job Job) {
+	task := scanner_task.NewTaskContext(w.ctx, w.db, job.album, job.cache)
+	if err := scanner.ScanAlbum(task); err != nil {
+		log.Error(w.ctx, "Faild to scan album", "error", err, "album", job.album.Title)
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/queue/worker.go
```diff
@@ -0,0 +1,63 @@
...
+	parentWaiter *sync.WaitGroup
+}
+
+func newWorker(ctx context.Context, db *gorm.DB, input <-chan Job, parentWaiter *sync.WaitGroup) *worker {
+	return &worker{
+		ctx:          ctx,
+		input:        input,
+		done:         make(chan struct{}),
+		parentWaiter: parentWaiter,
+	}
```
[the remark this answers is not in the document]
googollee: Fixed.

kkovaletp:
@googollee, so far so good)
I guess it is too early for a deep review. 
I didn't find a workers' balancer (the code, which controls the workers and balances tasks between them, making sure all of them are busy evenly and no one is stuck). Our scan job sequence has dependencies and constraints, so some tasks have to wait for the prerequisites before they can run, even if there are free workers for them. Maybe this logic would be part of the job, but it is better to double-check)

googollee:
> I didn't find a workers' balancer

`Worker.input` channel of all workers are the same instance, which is `Queue.handout` channel. See how a worker is [created](https://github.com/photoview/photoview/pull/1217/files#diff-f82fb2dea4c85866b276c74c4a99d22bca3894dd59f23657e42505de3f896de2R88). The channel will balance the job between workers automatically if more than one workers wait for loading.

googollee:
@kkovaletp if possible, please provide some ideas about the interface of `Queue`, like "how to use it in `server.go`, is it clear to understand what a method does from the name/design." Thank you.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,190 @@
...
+	albums, album_errors := scanner.FindAlbumsForUser(q.db, user, albumCache)
+	if len(album_errors) != 0 {
+		return nil, Errors(album_errors)
+	}
+
+	jobs := make([]Job, 0, len(albums))
+	for _, album := range albums {
+		jobs = append(jobs, Job{
+			album: album,
+			cache: albumCache,
```
kkovaletp, line 238: @googollee, is it ok here that the same album cache value is reused for all albums in the range?

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,190 @@
...
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-q.done:
+		return fmt.Errorf("queue is closed")
+	}
+
+	return nil
+}
+
+func (q *Queue) AddUser(ctx context.Context, user *models.User) error {
```
kkovaletp: @googollee, the function name doesn't represent the action: it adds scan jobs for the user, not the user itself. How about `ScanUserAlbums`?

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	var jobs []Job
+
+	for _, user := range users {
+		job, err := q.findUserJobs(user)
+		if err != nil {
+			return fmt.Errorf("failed to add user (id: %d) for scanning: %w", user.ID, err)
+		}
+		jobs = append(jobs, job...)
```
kkovaletp: @googollee, why not to call the `AddUser()` inside the loop? It looks like we do pretty the same actions and duplicate the code. Did I miss something?

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,190 @@
...
+			q.backlog = append(q.backlog, input...)
+		case <-q.trigger.C:
+			q.AddAllUsers(q.ctx)
+		case <-q.done:
+			break MAIN
+		}
+	}
+}
+
+func (q *Queue) AddAllUsers(ctx context.Context) error {
```
kkovaletp: @googollee, the function name doesn't represent the action: it adds scan jobs for all users, not the users themselves. How about `ScanAllUsersAlbums`?

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,190 @@
...
+		if len(q.backlog) > 0 {
+			select {
+			case q.handout <- q.backlog[0]:
+				q.backlog = q.backlog[1:]
+
+			case input := <-q.input:
+				q.backlog = append(q.backlog, input...)
+			case <-q.trigger.C:
+				q.AddAllUsers(q.ctx)
+			case <-q.done:
```
kkovaletp: @googollee, I see such selects several times in this file with minor changes. Is it possible to extract it to a private function and call it with params when needed?
My intention here is to make the code maintainable and readable better. If you think that this will make it worse, it is fine to reject this suggestion)

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,190 @@
...
+func (q *Queue) Close() {
+	q.trigger.Stop()
+	close(q.done)
+}
+
+func (q *Queue) UpdateInterval(newInterval time.Duration) {
+	q.trigger.Reset(newInterval)
+}
+
+func (q *Queue) UpdateWorkers(newMax int) {
```
kkovaletp: @googollee, the function name doesn't represent the action: it resizes the pool of available workers. I don't see any other updates for the existing workers. 
How about `ResizeWorkersPool`?

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return ret, nil
+}
+
+func (q *Queue) Close() {
+	q.trigger.Stop()
+	close(q.done)
+}
+
+func (q *Queue) UpdateInterval(newInterval time.Duration) {
```
kkovaletp: @googollee, the name is ok, but how about `UpdateScanInterval`?

kkovaletp (commented):
@googollee, I don't have any ideas about the usage from the `server.go` now. Maybe I'll have some later or comment on your implementation...
But here are my comments regarding the names and some code)

api/scanner/queue/queue.go
googollee: From the current code, it doesn't share the cache between albums. I tend to keep the current usage. If it's necessary to share it, we could update the code at that time.

api/scanner/queue/queue.go
googollee: Sure. Updated.

api/scanner/queue/queue.go
googollee: `AddUser()` has communication. It's better to reduce the count of communication.

api/scanner/queue/queue.go
googollee: Updated.

api/scanner/queue/queue.go
googollee: There is no same select in the code. Since Go forces a select with static branches, it can't merge two selects if one has more branches than the other.

The select here needs to send out jobs in the backlog. The select below doesn't, because the backlog is empty. In my opinion, this `Run()` is simple enough, since the function is 32 lines. Making two functions like `handleChannelsWithBacklog`/ `handleChannelsWithoutBacklog` with different selects doesn't give more meanings here.

api/scanner/queue/queue.go
googollee: How about `ResizeWrokers`? The queue doesn't have a concept of `Pool` now. I tend not to use this word.

api/scanner/queue/queue.go
googollee: Updated.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,190 @@
...
+			select {
+			case q.handout <- q.backlog[0]:
+				q.backlog = q.backlog[1:]
+
+			case input := <-q.input:
+				q.backlog = append(q.backlog, input...)
+			case <-q.trigger.C:
+				q.AddAllAlbums(q.ctx)
+			case <-q.done:
+				break MAIN
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (q *Queue) Close() {
+	q.trigger.Stop()
+	close(q.done)
+}
+
+func (q *Queue) UpdateInterval(newInterval time.Duration) {
+	q.trigger.Reset(newInterval)
+}
```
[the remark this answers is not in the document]
googollee: Trust the input and let it panic if invalid.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,209 @@
...
+	interval := time.Duration(siteInfo.PeriodicScanInterval) * time.Second
+
+	ret := &Queue{
+		db:      db,
+		ctx:     log.WithAttrs(context.Background(), "process", "queue"),
+		input:   make(chan []Job),
+		handout: make(chan Job),
+		done:    make(chan struct{}),
+		trigger: time.NewTicker(interval),
+	}
```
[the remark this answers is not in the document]
googollee: Trust the input and let it panic if invalid.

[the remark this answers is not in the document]
googollee: OK. I updated the function to check the interval now.

api/scanner/queue/queue.go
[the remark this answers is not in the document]
googollee: OK. I updated the function to check the interval now.

api/scanner/queue/queue.go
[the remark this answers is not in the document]
googollee: Stop the ticker when the interval is 0.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,216 @@
...
+
+	ret.ResizeWorkers(siteInfo.ConcurrentWorkers)
+
+	return ret, nil
+}
+
+func (q *Queue) Close() {
+	q.trigger.Stop()
+	close(q.done)
+}
```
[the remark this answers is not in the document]
googollee: Updated.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	if newInterval == 0 {
+		q.trigger.Stop()
+		return nil
+	}
+
+	q.trigger.Reset(newInterval)
+	return nil
+}
```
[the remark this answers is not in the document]
googollee: `q.trigger` is created when creating a `Queue` and never be set as `nil` again.  No need to check `nil` here.

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,240 @@
...
+	if len(album_errors) != 0 {
+		return nil, Errors(album_errors)
+	}
+
+	jobs := make([]Job, 0, len(albums))
+	for _, album := range albums {
+		jobs = append(jobs, Job{
+			album: album,
+			cache: albumCache,
+		})
```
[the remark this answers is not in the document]
googollee: It's possible. But currently I tend to use the same logic as `scanner_queue.AddUserToQueue()`. I'll check if it's valid with unit tests later.

api/scanner/queue/queue.go
kkovaletp: @googollee, sorry, maybe I'm missing something, but as I see, the `albumCache` variable gets its value in the 1st line of the function - outside of the loop. Then, in this loop, each album gets its `cache` property assigned to the same `albumCache` value. That is why I asked the question: Is it ok here? Because, as far as I understand, each album should have its own sub-folder in the main cache folder.

api/scanner/queue/queue.go
kkovaletp: @googollee, what is the correct name of the struct, the workers exist in? Is it an array? Let's use that word instead of the `pool`.
I don't like the `ResizeWrokers` because it means to make the workers bigger or smaller instead of changing the # of workers.

api/scanner/queue/queue.go
kkovaletp: Ok, I found the https://github.com/photoview/photoview/pull/1217#discussion_r2139778987 thread, talking about the same issue with a much better problem description, so closing this one to keep things simple

api/scanner/queue/queue.go
googollee: How about `RescaleWorkers`?

kkovaletp: `RescaleWorkers` looks good.

googollee: Done. I'll continue the remain things.

googollee:
We agreed on the design of the queue. I close this pr now and keep working on it. After the other part is done, I'll submit a new PR.

PR #1221, opened 2025-06-21
Refactor Docker build workflows

kkovaletp:
My intention was to optimize and speed up Docker build workflows to make CI job runs faster, as every time we commit something, we need to wait up to 15 minutes for jobs to complete. This becomes even more critical for me to process Dependabot's weekly version update PRs: before this change, I had to spend ~0.5-1 day to get all UI and API version update PRs merged.

Not all changes look obvious, so I'll add comments with explanations to some of them.

## Summary by CodeRabbit

- **Chores**
  - Improved Docker build and CI workflows with more granular and efficient caching for dependencies and test images.
  - Enhanced dependency build scripts to support caching of build artifacts, reducing redundant builds.
  - Updated documentation in the dependencies directory for clarity.
  - Refined Dockerfile build arguments and environment variable handling for more consistent builds.
  - Improved script formatting and robustness across dependency build scripts.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,12 +129,15 @@ jobs:
           tags:        ${{ steps.docker_meta.outputs.tags }}
           labels:      ${{ steps.docker_meta.outputs.labels }}
           annotations: ${{ steps.docker_meta.outputs.annotations }}
-          cache-from:  type=gha
-          cache-to:    type=gha,mode=max
+          cache-from: |
+            type=gha,scope=test-ui-${{ hashFiles('ui/package-lock.json') }}
+            type=gha,scope=test-api-${{ hashFiles('api/go.sum', 'scripts/install_*.sh') }}
+          cache-to: |
+            type=gha,mode=max,scope=test-ui-${{ hashFiles('ui/package-lock.json') }}
+            type=gha,mode=max,scope=test-api-${{ hashFiles('api/go.sum', 'scripts/install_*.sh') }}
           build-args: |
             NODE_ENV=production
             VERSION=${{ github.ref_name }}
-            COMMIT_SHA=${{ github.sha }}
 
   dockle:
     name: Dockle Container Analysis
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -59,13 +59,17 @@ jobs:
           set -euo pipefail
           GITHUB_TOKEN="${{ secrets.GITHUB_TOKEN }}"
           # Fetch latest version tags from GitHub releases
-          LIBRAW_VERSION=$(curl -fsSL --retry 3 ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+          LIBRAW_VERSION=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
+            ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
             "https://api.github.com/repos/LibRaw/LibRaw/releases/latest" | jq -r '.tag_name')
-          LIBHEIF_VERSION=$(curl -fsSL --retry 3 ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+          LIBHEIF_VERSION=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
+            ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
             "https://api.github.com/repos/strukturag/libheif/releases/latest" | jq -r '.tag_name')
-          IMAGEMAGICK_VERSION=$(curl -fsSL --retry 3 ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+          IMAGEMAGICK_VERSION=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
+            ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
             "https://api.github.com/repos/ImageMagick/ImageMagick/releases/latest" | jq -r '.tag_name')
-          JELLYFIN_FFMPEG_VERSION=$(curl -fsSL --retry 3 ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+          JELLYFIN_FFMPEG_VERSION=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
+            ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
             "https://api.github.com/repos/jellyfin/jellyfin-ffmpeg/releases/latest" | jq -r '.tag_name')
           # Output as environment variables
           echo "LIBRAW_VERSION=${LIBRAW_VERSION}" | tee -a $GITHUB_OUTPUT
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -59,13 +59,13 @@ jobs:
           set -euo pipefail
           GITHUB_TOKEN="${{ secrets.GITHUB_TOKEN }}"
           # Fetch latest version tags from GitHub releases
-          LIBRAW_VERSION=$(curl -fsSL ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+          LIBRAW_VERSION=$(curl -fsSL --retry 3 ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
             "https://api.github.com/repos/LibRaw/LibRaw/releases/latest" | jq -r '.tag_name')
-          LIBHEIF_VERSION=$(curl -fsSL ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+          LIBHEIF_VERSION=$(curl -fsSL --retry 3 ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
             "https://api.github.com/repos/strukturag/libheif/releases/latest" | jq -r '.tag_name')
-          IMAGEMAGICK_VERSION=$(curl -fsSL ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+          IMAGEMAGICK_VERSION=$(curl -fsSL --retry 3 ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
             "https://api.github.com/repos/ImageMagick/ImageMagick/releases/latest" | jq -r '.tag_name')
-          JELLYFIN_FFMPEG_VERSION=$(curl -fsSL ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+          JELLYFIN_FFMPEG_VERSION=$(curl -fsSL --retry 3 ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
             "https://api.github.com/repos/jellyfin/jellyfin-ffmpeg/releases/latest" | jq -r '.tag_name')
           # Output as environment variables
           echo "LIBRAW_VERSION=${LIBRAW_VERSION}" | tee -a $GITHUB_OUTPUT
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -11,7 +11,7 @@ on:
 
 env:
   IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
-  IS_CACHING: ${{ github.event_name == 'pull_request' }}
+  IS_CACHING: true
   DOCKER_USERNAME: viktorstrate
   DOCKER_IMAGE: photoview/dependencies
   PLATFORMS: linux/amd64,linux/arm64
@@ -51,6 +51,28 @@ jobs:
             type=raw,value=latest
             type=sha
 
+      - name: Detect dependency versions
+        id: versions
+        working-directory: dependencies
+        shell: bash
+        run: |
+          set -euo pipefail
+          GITHUB_TOKEN="${{ secrets.GITHUB_TOKEN }}"
+          # Fetch latest version tags from GitHub releases
+          LIBRAW_VERSION=$(curl -fsSL ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+            "https://api.github.com/repos/LibRaw/LibRaw/releases/latest" | jq -r '.tag_name')
+          LIBHEIF_VERSION=$(curl -fsSL ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+            "https://api.github.com/repos/strukturag/libheif/releases/latest" | jq -r '.tag_name')
+          IMAGEMAGICK_VERSION=$(curl -fsSL ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+            "https://api.github.com/repos/ImageMagick/ImageMagick/releases/latest" | jq -r '.tag_name')
+          JELLYFIN_FFMPEG_VERSION=$(curl -fsSL ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} \
+            "https://api.github.com/repos/jellyfin/jellyfin-ffmpeg/releases/latest" | jq -r '.tag_name')
+          # Output as environment variables
+          echo "LIBRAW_VERSION=${LIBRAW_VERSION}" | tee -a $GITHUB_OUTPUT
+          echo "LIBHEIF_VERSION=${LIBHEIF_VERSION}" | tee -a $GITHUB_OUTPUT
+          echo "IMAGEMAGICK_VERSION=${IMAGEMAGICK_VERSION}" | tee -a $GITHUB_OUTPUT
+          echo "JELLYFIN_FFMPEG_VERSION=${JELLYFIN_FFMPEG_VERSION}" | tee -a $GITHUB_OUTPUT
+
       - name: Build and push
         uses: docker/build-push-action@v6
         with:
@@ -61,8 +83,15 @@ jobs:
           tags:        ${{ steps.docker_meta.outputs.tags }}
           labels:      ${{ steps.docker_meta.outputs.labels }}
           annotations: ${{ steps.docker_meta.outputs.annotations }}
-          cache-from:  ${{ ( env.IS_CACHING == 'true' && 'type=gha' ) || '' }}
-          cache-to:    ${{ ( env.IS_CACHING == 'true' && 'type=gha,mode=max' ) || '' }}
-          no-cache:    ${{ env.IS_CACHING != 'true' }}
           sbom:        true
           provenance:  mode=max
+          no-cache:    ${{ env.IS_CACHING != 'true' }}
+          cache-from:  type=gha,scope=dependencies
+          cache-to:    type=gha,mode=max,scope=dependencies
+          secrets: |
+            github_token=${{ secrets.GITHUB_TOKEN }}
+          build-args: |
+            LIBRAW_VERSION=${{ steps.versions.outputs.LIBRAW_VERSION }}
+            LIBHEIF_VERSION=${{ steps.versions.outputs.LIBHEIF_VERSION }}
+            IMAGEMAGICK_VERSION=${{ steps.versions.outputs.IMAGEMAGICK_VERSION }}
+            JELLYFIN_FFMPEG_VERSION=${{ steps.versions.outputs.JELLYFIN_FFMPEG_VERSION }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,13 +56,13 @@ jobs:
     - name: Build test image
       uses: docker/build-push-action@v6
       with:
-        pull: true
-        push: false
-        load: true
-        target: api
-        tags: photoview/api
-        cache-from: type=gha
-        cache-to: type=gha,mode=max
+        pull:       true
+        push:       false
+        load:       true
+        target:     api
+        tags:       photoview/api
+        cache-from: type=gha,scope=test-api-${{ hashFiles('api/go.sum', 'scripts/install_*.sh') }}
+        cache-to:   type=gha,mode=max,scope=test-api-${{ hashFiles('api/go.sum', 'scripts/install_*.sh') }}
 
     - name: Test
       id: test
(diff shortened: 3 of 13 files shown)


Dockerfile
```diff
@@ -1,42 +1,36 @@
...
-
 WORKDIR /app/ui
 
 COPY ui/package.json ui/package-lock.json /app/ui/
 RUN npm ci
 
 COPY ui/ /app/ui
-RUN if [ "${BUILD_DATE}" = "undefined" ]; then \
-    export BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ'); \
+RUN export BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ'); \
```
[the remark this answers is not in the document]
kkovaletp: All commands in this RUN block are safe to ignore their return values.

Dockerfile
```diff
@@ -1,42 +1,36 @@
...
 
 # See for details: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
 ARG NODE_ENV
 ENV NODE_ENV=${NODE_ENV:-production}
 
+# Set environment variable REACT_APP_API_ENDPOINT from build args, uses "<web server>/api" as default
 ARG REACT_APP_API_ENDPOINT
 ENV REACT_APP_API_ENDPOINT=${REACT_APP_API_ENDPOINT}
```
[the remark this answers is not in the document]
kkovaletp: No need, as the fallback value is handled by the JS code inside the project.

.github/workflows/build.yml
```diff
@@ -129,12 +129,15 @@ jobs:
...
           labels:      ${{ steps.docker_meta.outputs.labels }}
           annotations: ${{ steps.docker_meta.outputs.annotations }}
-          cache-from:  type=gha
-          cache-to:    type=gha,mode=max
+          cache-from: |
+            type=gha,scope=test-ui-${{ hashFiles('ui/package-lock.json') }}
+            type=gha,scope=test-api-${{ hashFiles('api/go.sum', 'scripts/install_*.sh') }}
+          cache-to: |
+            type=gha,mode=max,scope=test-ui-${{ hashFiles('ui/package-lock.json') }}
+            type=gha,mode=max,scope=test-api-${{ hashFiles('api/go.sum', 'scripts/install_*.sh') }}
```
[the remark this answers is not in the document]
kkovaletp: They are named in the same way as the corresponding scopes in the `tests.yml` to let them be reused if matched

dependencies/download_jellyfin-ffmpeg.sh
```diff
@@ -1,14 +1,32 @@
...
-curl -L -o ./jellyfin-ffmpeg.deb "$URL"
+echo "Downloading jellyfin-ffmpeg ${JELLYFIN_FFMPEG_VERSION} (cache miss)..."
 
+echo Compiler: "${DEB_HOST_MULTIARCH}" Arch: "${DEB_HOST_ARCH}"
+
+URL="https://api.github.com/repos/jellyfin/jellyfin-ffmpeg/tarball/${JELLYFIN_FFMPEG_VERSION}"
+echo download jellyfin-ffmpeg from "$URL"
 mkdir -p /output/deb
-cp ./jellyfin-ffmpeg.deb /output/deb/
+curl -L -o /output/deb/jellyfin-ffmpeg.deb ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} "$URL"
```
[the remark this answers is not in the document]
kkovaletp: outdated

dependencies/download_jellyfin-ffmpeg.sh
```diff
@@ -1,14 +1,32 @@
...
+
+# After successful download, cache the results
+echo "Caching jellyfin-ffmpeg ${JELLYFIN_FFMPEG_VERSION} build results..."
+mkdir -p "${CACHE_DIR}/output"
+cp -r /output/* "${CACHE_DIR}/output/"
+touch "$CACHE_MARKER"
+
+echo "jellyfin-ffmpeg ${JELLYFIN_FFMPEG_VERSION} downloaded and cached"
```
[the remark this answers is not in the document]
kkovaletp: unneded complexity

dependencies/build_libheif.sh
```diff
@@ -1,19 +1,47 @@
 #!/bin/bash
 set -euo pipefail
 
-: ${DEB_HOST_MULTIARCH=`uname -m`-linux-gnu}
-: ${DEB_HOST_ARCH=`dpkg --print-architecture`}
+: "${DEB_HOST_MULTIARCH:=$(uname -m)-linux-gnu}"
+: "${DEB_HOST_ARCH:=$(dpkg --print-architecture)}"
+CACHE_DIR="${BUILD_CACHE_DIR:-/build-cache}/libheif-${LIBHEIF_VERSION}"
+CACHE_MARKER="${CACHE_DIR}/libheif-${LIBHEIF_VERSION}-complete"
```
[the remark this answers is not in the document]
kkovaletp: if version is not provided, it should fail

dependencies/build_imagemagick.sh
```diff
@@ -1,15 +1,43 @@
 #!/bin/bash
 set -euo pipefail
 
-: ${DEB_HOST_MULTIARCH=`uname -m`-linux-gnu}
-: ${DEB_HOST_ARCH=`dpkg --print-architecture`}
+: "${DEB_HOST_MULTIARCH:=$(uname -m)-linux-gnu}"
+: "${DEB_HOST_ARCH:=$(dpkg --print-architecture)}"
+CACHE_DIR="${BUILD_CACHE_DIR:-/build-cache}/ImageMagick-${IMAGEMAGICK_VERSION}"
+CACHE_MARKER="${CACHE_DIR}/ImageMagick-${IMAGEMAGICK_VERSION}-complete"
```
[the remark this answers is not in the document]
kkovaletp: if version is not provided, it should fail

dependencies/Dockerfile
```diff
@@ -1,26 +1,62 @@
...
+FROM base-prep AS build-libraw
+ARG LIBRAW_VERSION
+ARG TARGETARCH
+ENV BUILD_CACHE_DIR=/build-cache
+WORKDIR /tmp/build-libraw
+COPY --chmod=0755 build_libraw.sh .
+RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=libraw-${TARGETARCH}-${LIBRAW_VERSION} \
+    --mount=type=secret,id=github_token \
+    export GITHUB_TOKEN=$(cat /run/secrets/github_token) \
+    && export $(cat /env) && ./build_libraw.sh
```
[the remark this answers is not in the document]
kkovaletp: It is not printed anywhere and is protected by the GitHub runner's logger, so it is safe to use it this way. Currently, I use inline parameter expansion to conditionally add the auth header to the `curl` command, based on the existence of the token. It won't work like that if I replace the variable with the `cat` command, and will require a more complex form of `if-else` construction.

.github/workflows/build.yml
kkovaletp, line 137: Explanation:
Docker build cache works as "all or nothing", so with the previous definition, a change to the API part also invalidates the UI part of the cache.
Another issue is that we build testing images from the same Dockerfile, but we were not able to reuse its cache.

With these 2 parallel caches, labeled with keys for API and UI, it should work much better and allow at least partial cache reuse for unchanged layers.

.github/workflows/build.yml
```diff
@@ -129,12 +129,15 @@ jobs:
...
           build-args: |
             NODE_ENV=production
             VERSION=${{ github.ref_name }}
-            COMMIT_SHA=${{ github.sha }}
```
kkovaletp, line 137: Explanation:
Every changed value of a build argument triggers cache invalidation, such as when there is a change in the Dockerfile or the repository's files, but it impacts all layers referenced. This means that with this build arg, we had to always rebuild the UI even if there were no changes.

I've removed it from here and populated it as a regular env var in the Dockerfile's RUN step of the UI build command. Now it runs only if there is a change.
Side effect: We are going to have old (cached) UI, pointing to old commit in every branch if the UI cache is regularly used and not removed from Github cache storage (10GB or 30 days limit with removing the oldest cache files 1st). Cross-branch builds still invalidate the previous cache because of the VERSION build arg. Since we don't change the UI part, I think that this side effect is acceptable...

.github/workflows/dependencies.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 env:
   IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
-  IS_CACHING: ${{ github.event_name == 'pull_request' }}
+  IS_CACHING: true
```
kkovaletp, line 14: Explanation:
As now the caching algorithm is much better, it is safe to use it everywhere. I keep this global variable set to `true` just for easy debugging purposes. If it is not needed, I can remove it and move the `true` value to the build step params.

.github/workflows/dependencies.yml
```diff
@@ -61,8 +87,15 @@ jobs:
...
+          no-cache:    ${{ env.IS_CACHING != 'true' }}
+          cache-from:  type=gha,scope=dependencies
+          cache-to:    type=gha,mode=max,scope=dependencies
+          secrets: |
+            github_token=${{ secrets.GITHUB_TOKEN }}
+          build-args: |
+            LIBRAW_VERSION=${{ steps.versions.outputs.LIBRAW_VERSION }}
+            LIBHEIF_VERSION=${{ steps.versions.outputs.LIBHEIF_VERSION }}
+            IMAGEMAGICK_VERSION=${{ steps.versions.outputs.IMAGEMAGICK_VERSION }}
+            JELLYFIN_FFMPEG_VERSION=${{ steps.versions.outputs.JELLYFIN_FFMPEG_VERSION }}
```
kkovaletp, line 101: Explanation:
If at least 1 version is changed, all the build cache is invalidated. Duplicating it 4 times with different cache keys (1 per lib's version) is a bad idea, as we have 10GB limit for the cache in the repo (or org, not sure), so we can quickly fill all 10GB and loose the cache benefits (taking into account also the main image build caches).
So, the solution is to combine 1 generic cache key with lib's version/arch-specific file-level caching on the Dockerfile level. It allows us to use much less space in the caching storage and have more cached builds within 30 days, of course.

.github/workflows/tests.yml
```diff
@@ -56,13 +56,13 @@ jobs:
...
-        tags: photoview/api
-        cache-from: type=gha
-        cache-to: type=gha,mode=max
+        pull:       true
+        push:       false
+        load:       true
+        target:     api
+        tags:       photoview/api
+        cache-from: type=gha,scope=test-api-${{ hashFiles('api/go.sum', 'scripts/install_*.sh') }}
+        cache-to:   type=gha,mode=max,scope=test-api-${{ hashFiles('api/go.sum', 'scripts/install_*.sh') }}
```
kkovaletp, line 65: Explanation:
Use the same cache keys to be able to reuse the same cache between build and test workflows. The same is for the UI part here.

Dockerfile
```diff
@@ -1,42 +1,37 @@
...
-ENV VERSION=${VERSION:-undefined}
-ENV REACT_APP_BUILD_VERSION=${VERSION:-undefined}
-
-ARG BUILD_DATE
-ENV BUILD_DATE=${BUILD_DATE:-undefined}
-ENV REACT_APP_BUILD_DATE=${BUILD_DATE:-undefined}
-
-ARG COMMIT_SHA
-ENV COMMIT_SHA=${COMMIT_SHA:-}
-ENV REACT_APP_BUILD_COMMIT_SHA=${COMMIT_SHA:-}
```
kkovaletp, line 27: Explanation:
Removed these build args and defined the variables' values in the RUN step to overcome the build cache invalidation issue.
The VERSION is still an arg (line 5), but is handled better for local builds (line 32).
The commit sha is not provided as a build arg from the workflow step, but as an implicit default Github env variable (line 4). I hope that the implicit arg passing overcomes the cache invalidation issue, but if not, it can be dropped and replaced by the same command used for local builds (line 30).

dependencies/Dockerfile
```diff
@@ -1,26 +1,60 @@
...
+RUN /tmp/prepare.sh \
+    && mkdir -p /output/bin /output/lib /output/include /output/pkgconfig /output/etc /output/deb
+
+FROM base-prep AS build-libraw
+ARG LIBRAW_VERSION
+ARG TARGETARCH
+ENV BUILD_CACHE_DIR=/build-cache
+WORKDIR /tmp/build-libraw
+COPY --chmod=0755 build_libraw.sh .
+RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=libraw-${TARGETARCH}-${LIBRAW_VERSION} \
```
kkovaletp, line 14: Explanation:
Each build/download stage runs in parallel to other build stages and architectures.
This mount is file-level caching with ID, working in the same way as a cache key. As it contains the arch and the version, it is recreated every time a new version is detected, and it is not mixed between different arches per the same version, so there is no need to maintain concurrent access.

dependencies/Dockerfile
```diff
@@ -1,26 +1,60 @@
...
+COPY --from=build-imagemagick /output/ /output/
+COPY --from=download-jellyfin-ffmpeg /output/ /output/
+COPY --chmod=0755 output.sh /
 RUN /output.sh
 
-FROM debian:bookworm-slim AS release
-COPY --from=build /artifacts.tar.gz /
+FROM scratch AS release
+COPY --from=final-assembly /artifacts.tar.gz /
+USER nobody
```
kkovaletp, line 60: Explanation:
- I use the `scratch` base image, as we don't need to run it, so there is no need to waste space and traffic to keep our default image content.
- I added the `USER nobody` just to silence a linter. Since the base image is not runnable, it has no real effect, but also costs nothing.

dependencies/build_imagemagick.sh
```diff
@@ -1,15 +1,44 @@
...
+  libxml2-dev:"${DEB_HOST_ARCH}" \
+  libfftw3-dev:"${DEB_HOST_ARCH}" \
+  zlib1g-dev:"${DEB_HOST_ARCH}" \
+  liblzma-dev:"${DEB_HOST_ARCH}" \
+  libbz2-dev:"${DEB_HOST_ARCH}"
+
+URL="https://api.github.com/repos/ImageMagick/ImageMagick/tarball/${IMAGEMAGICK_VERSION}"
+echo download ImageMagick from "$URL"
+curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 -o ./magick.tar.gz \
+  ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} "$URL"
```
kkovaletp, line 49: Explanation:
For all build/download stages, `curl` now runs with better options:
-f - fails on errors
-s - hides the download progress, which is ugly in logs
-S - shows the errors in the log
-L - follows redirects
retries 2 times in case of error with 5 sec delay between retries and within 60 sec in total, but this 60 sec limit doesn't affect normal downloads - they can take more time as long as they run normally.
It uses a GitHub token to pass the authentication process to overcome the GitHub API rate limit issue on the CI.

dependencies/build_imagemagick.sh
```diff
@@ -20,13 +49,13 @@ cd ImageMagick-*
...
   --without-perl --disable-doc \
-  --host=${DEB_HOST_MULTIARCH}
+  --host="${DEB_HOST_MULTIARCH}"
 make
 make install
 cd ..
 
 mkdir -p /output/bin /output/etc /output/lib /output/include /output/pkgconfig
-cp /usr/local/bin/magick /output/bin/
+cp -a /usr/local/bin/magick /output/bin/
```
kkovaletp, line 66: I hope that this is fine to add the `-a` option here. If it wasn't added on purpose - please let me know)

dependencies/download_jellyfin-ffmpeg.sh
```diff
@@ -1,14 +1,34 @@
...
-URL=$(curl -s https://api.github.com/repos/jellyfin/jellyfin-ffmpeg/releases/latest | grep \"browser_download_url\".*bookworm_${DEB_HOST_ARCH} | cut -d : -f 2,3 | tr -d ' ,"')
-echo download jellyfin-ffmpeg from $URL
-curl -L -o ./jellyfin-ffmpeg.deb "$URL"
+echo "Downloading jellyfin-ffmpeg ${JELLYFIN_FFMPEG_VERSION} (cache miss)..."
 
+echo Compiler: "${DEB_HOST_MULTIARCH}" Arch: "${DEB_HOST_ARCH}"
+
+VER="${JELLYFIN_FFMPEG_VERSION#v}"
+MAJOR_VER=$(echo "${VER}" | cut -d. -f1)
+URL="https://github.com/jellyfin/jellyfin-ffmpeg/releases/download/${JELLYFIN_FFMPEG_VERSION}/jellyfin-ffmpeg${MAJOR_VER}_${VER}-bookworm_${DEB_HOST_ARCH}.deb"
```
kkovaletp, line 31: Explanation:
As we download the pre-built version, I use the arch-specific browser-download-url. It is not hosted by api.github.com, so it doesn't have rate limits, and there is no need for auth for public repos - that is why we don't have a GitHub token here.

dependencies/output.sh
```diff
@@ -2,4 +2,4 @@
 set -euo pipefail
 
 cd /output
-tar czfv /artifacts.tar.gz *
+tar czfv /artifacts.tar.gz ./*
```
kkovaletp, line 5: Explanation:
The `./` is needed to fix the issue: "Use ./*glob* or -- *glob* so names with dashes won't become options."

dependencies/prepare.sh
```diff
@@ -1,27 +1,34 @@
...
+  autoconf \
+  automake \
+  libtool \
+  m4 \
+  pkg-config \
+  cmake
 
-dpkg-architecture -a $DEBIAN_ARCH >/env
+dpkg-architecture -a "$DEBIAN_ARCH" >/env
+echo "PKG_CONFIG_PATH=/usr/lib/${DEB_HOST_MULTIARCH}/pkgconfig" >>/env
```
kkovaletp, line 29: I moved this line before the export to make it part of it. If it shouldn't - let me know

ui/apollo.config.js
```diff
@@ -6,7 +6,7 @@ const schemasFolder = path.join(__dirname, '..', 'api', 'graphql', 'resolvers');
 
 let completeSchema;
 try {
-   completeSchema = readdirSync(schemasFolder)
+  completeSchema = readdirSync(schemasFolder)
```
kkovaletp, line 9: Just an indentation fix

kkovaletp (commented):
@googollee, I've added my comments and explanations for key parts of the refactored implementation. I hope that it will help you better understand it during the code reading.
Please resolve them as other conversation threads once you're done with each of them.

dependencies/prepare.sh
```diff
@@ -1,27 +1,34 @@
...
 set -euo pipefail
 
-: ${TARGETPLATFORM=linux/`dpkg --print-architecture`}
+: "${TARGETPLATFORM:=linux/$(dpkg --print-architecture)}"
+: "${DEB_HOST_MULTIARCH:=$(uname -m)-linux-gnu}"
 
-TARGETOS="$(echo $TARGETPLATFORM | cut -d"/" -f1)"
-TARGETARCH="$(echo $TARGETPLATFORM | cut -d"/" -f2)"
-TARGETVARIANT="$(echo $TARGETPLATFORM | cut -d"/" -f3)"
+#TARGETOS="$(echo "$TARGETPLATFORM" | cut -d"/" -f1)"
```
googollee: We can remove `TARGETOS` and `TARGETVARIANT`, since we do not support `arm/v6` and `arm/v7` now.

.github/workflows/build.yml
googollee: I'm fine with it. Thanks.

dependencies/download_jellyfin-ffmpeg.sh
googollee: For those version args like `JELLYFIN_FFMPEG_VERSION`, it's annoying for local developer to set them properly, instead of just running scripts.

Could we try to get the latest version if they are empty?

kkovaletp: Do you see a real need in running such scripts locally? As far as I see, they need to be executed only on the CI in the scope of the image build, and local users/developers should pull the `dependencies:latest` and use it, or in the case of a direct host deployment, install all 3rd-party dependencies manually on the host before Photoview deployment.
Please explain a use case for the regular local execution of such scripts, and I'll develop a fallback code for the latest version resolution (it will be a copy of the "Detect dependency versions" step + some wrappers). I think that this code duplication would make sense if there is a real need to do the local runs regularly, as it creates more maintenance overhead.

googollee: I ran it a lot when I was developing. Thinking about we need to update those scripts in the future. It's annoying to set those variables since we can't run GitHub workflows locally.

kkovaletp: I've added the fallback code

googollee (approved):

dependencies/prepare.sh
kkovaletp: Done

dependencies/download_jellyfin-ffmpeg.sh
```diff
@@ -1,14 +1,42 @@
 #!/bin/bash
 set -euo pipefail
 
-: ${DEB_HOST_MULTIARCH=x86_64-linux-gnu}
-: ${DEB_HOST_ARCH=amd64}
+: "${DEB_HOST_MULTIARCH:=$(uname -m)-linux-gnu}"
+: "${DEB_HOST_ARCH:=$(dpkg --print-architecture)}"
+CACHE_DIR="${BUILD_CACHE_DIR:-/build-cache}/jellyfin-ffmpeg-${JELLYFIN_FFMPEG_VERSION}"
+CACHE_MARKER="${CACHE_DIR}/jellyfin-ffmpeg-${JELLYFIN_FFMPEG_VERSION}-complete"
```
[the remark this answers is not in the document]
kkovaletp: Fixed for all 4 scripts

dependencies/build_libheif.sh
```diff
@@ -1,19 +1,56 @@
...
```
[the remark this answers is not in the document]
kkovaletp: Fixed for all 4 scripts

dependencies/build_libraw.sh
```diff
@@ -1,20 +1,50 @@
...
 
-: ${DEB_HOST_MULTIARCH=`uname -m`-linux-gnu}
-: ${DEB_HOST_ARCH=`dpkg --print-architecture`}
+# Fallback to the latest version if LIBRAW_VERSION is not set
+if [ -z "$LIBRAW_VERSION" ]; then
+  echo "WARN: LibRaw version is empty, most likely the script runs not on CI."
+  echo "Fetching the latest version from LibRaw repo..."
+  LIBRAW_VERSION=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
+    "https://api.github.com/repos/LibRaw/LibRaw/releases/latest" | jq -r '.tag_name')
+fi
```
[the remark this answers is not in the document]
kkovaletp: Fixed in all 4 build scripts

dependencies/build_libraw.sh
```diff
@@ -1,20 +1,50 @@
...
+
+apt-get install -y --no-install-recommends \
+  libjpeg62-turbo-dev:"${DEB_HOST_ARCH}" \
+  liblcms2-dev:"${DEB_HOST_ARCH}" \
+  zlib1g-dev:"${DEB_HOST_ARCH}"
+
+URL="https://api.github.com/repos/LibRaw/LibRaw/tarball/${LIBRAW_VERSION}"
+echo download libraw from "$URL"
+curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 -o ./libraw.tar.gz \
+  ${GITHUB_TOKEN:+-H "Authorization: Bearer ${GITHUB_TOKEN}"} "$URL"
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, the GitHub token cannot contain spaces. Please construct and show me what HTTP request curl will send with the current implementation in case when the token variable is set and unset.

PR #1224, opened 2025-06-25
Drop tables instead of truncating tables.

googollee:


api/graphql/models/album.go
```diff
@@ -99,6 +99,7 @@ func (a *Album) Thumbnail(db *gorm.DB) (*Media, error) {
 		)
 		SELECT * FROM media
 		WHERE media.album_id IN (SELECT id FROM sub_albums)
+		ORDER BY media.id
```
kkovaletp, line 102: If you handle the case, when more than 1 record can be found, doesn't it make sense to sort them in the DESC order, so that we always take the newest one (assuming that the `media.id` is incrementing)?

api/database/database.go
```diff
@@ -204,50 +205,16 @@ func MigrateDatabase(db *gorm.DB) error {
...
-			}
-		case drivers.MYSQL:
-			if err := tx.Exec(fmt.Sprintf("TRUNCATE TABLE %s", table)).Error; err != nil {
-				return err
-			}
-		case drivers.SQLITE:
-			if err := tx.Exec(fmt.Sprintf("DELETE FROM %s", table)).Error; err != nil {
-				return err
-			}
+		if err := db.Migrator().DropTable(model); err != nil {
```
kkovaletp, line 210: You replace 2 functions with truncating tables inside the transaction by a drop table call. Such simplification looks suspicious :)
As far as I understand, a truncated table still exists with all its metadata and relations, while a dropped table is just removed from the DB and needs to be recreated. Don't we need to handle this difference?
Another difference is that the order of dropping tables matters, while here we just iterate over the range. Are you sure that the tables are always in the correct order in that range?
And finally, how good is this change covered by our tests? Don't we need to cover some additional possible scenarios?

api/scanner/media_type/media_type_test.go
```diff
@@ -1,9 +1,15 @@
 package media_type
 
 import (
+	"flag"
 	"testing"
 )
 
+func init() {
+	// Avoid panic with providing flags in `test_utils/integration_setup.go`.
+	flag.CommandLine.Init("media_type", flag.ContinueOnError)
```
kkovaletp, line 10: Won't this flag mask some errors or defects in some cases?

api/database/database.go
```diff
@@ -204,50 +205,16 @@ func MigrateDatabase(db *gorm.DB) error {
...
+			errs = append(errs, err)
 		}
 	}
+
+	if err := errors.Join(errs...); err != nil {
+		return fmt.Errorf("drop tables error: %w", err)
+	}
+
 	return nil
 }
```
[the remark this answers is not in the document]
googollee: Postgres drops tables cascade. SQLite doesn't check foreign keys. I didn't know MySQL behaviors, but I will update if it fails.

api/graphql/models/album.go
googollee: [The test case](https://github.com/photoview/photoview/blob/73ae3e8132b89e2f8f1f90950ddd5d3cfbebc56e/api/graphql/models/actions/album_actions_test.go#L167) checks it in ASC order. This case randomly failed, so I put an order here.

In my opinion, I prefer that the album thumbnail should not be changed when adding photos.

api/database/database.go
googollee: For unit tests, we should clean tables with that metadata and relations and re-create them to ensure testing in a scratch environment. That's why I use `drop table` here.

api/scanner/media_type/media_type_test.go
googollee: The test is running with `-database` and `-filesystem`, see [integration_setup.go](https://github.com/photoview/photoview/blob/73ae3e8132b89e2f8f1f90950ddd5d3cfbebc56e/api/test_utils/integration_setup.go#L21). If I hope to run a single test like `media_type` package, it requires removing those flags; otherwise, Go reports invalid flags.

With this `ContinueOnError`, it's not necessary to remove flags to run a single test.

api/graphql/models/album.go
kkovaletp: Yes, I agree. Albums should have 1 thumbnail each at the specific time. But as you mentioned, all this ordering is to overcome a failure, caused by a hidden bug, causing an album to have more than 1 thumbnail at some point.
So, in the scope of dealing with the consequences of that bug, isn't it better to sort those thumbs in DESC order to get the latest added one? And to update the test as well in that case)
Maybe there is some explanation on why it should be in the ASC order. If yes, please tell me)

api/database/database.go
kkovaletp: ok, now I checked in the IDE, and it looks like this code is used only for testing. Why is it in the production code file? Does it make sense to move it to the test_utils?

api/database/database.go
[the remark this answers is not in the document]
googollee: GORM uses `CASCADE` to drop table with PostgreSQL, see [their code](https://github.com/go-gorm/postgres/blob/7f25bff35c074404bb3462bf3ef479e6018c5f1e/migrator.go#L240).

api/graphql/models/album.go
googollee: I don't have very strong opinion about the order, but just keep the current behavior. I'll create another PR to update the behavior. Keep this PR as fixing the issue.

api/database/database.go
googollee: Because `database_models` is a private variable in `database` library, and it's not a good idea to public it.

In my opinion, it's fine to keep that function in `database`.

api/graphql/models/album.go
googollee: See https://github.com/photoview/photoview/pull/1225

api/database/database.go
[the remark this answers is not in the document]
kkovaletp: @googollee, so what is the final decision regarding the MySQL case here? Is there a robust way to implement this drop table call for all DBs, or should we have a separate branch for MySQL with the `CASCADE` explicitly mentioned?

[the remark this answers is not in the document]
googollee: @kkovaletp  MySQL also drops table with `CASCADE`, see [the code](https://github.com/go-gorm/mysql/blob/d744156745998478ffa3c5395c3336ae6f0b0c0c/migrator.go#L298).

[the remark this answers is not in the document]
kkovaletp: Great, so no changes are needed.

kkovaletp (approved):
LGTM

PR #1225, opened 2025-06-26
Change default thumbnail of album.

googollee:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated album cover selection to use the most recent photo as the default thumbnail when no specific cover is set or after a reset.

* **Tests**
  * Adjusted test cases to align with the new default album cover selection behavior.


kkovaletp (approved):
LGTM

PR #1226, opened 2025-06-26
Make TaskContext as a real context.

googollee:


googollee:
Hmm, I don't know why tests failed. I can't reproduce it locally. Since the code doesn't change any logic, I have no clue why. @coderabbitai any idea?

googollee:
I guess it relates with current queue or periodical scanner. I'll submit with the new queue later.

PR #1229, opened 2025-06-27
Fix the local build of the dependencies image, including the debug image instructions

### LINKED ISSUE #1227: Can't build `dependencies` locally.
Log:

```
$ docker build -t photoview/dependencies .
...
 => [download-jellyfin-ffmpeg 2/3] COPY --chmod=0755 download_jellyfin-ffmpeg.sh .                               0.3s
 => ERROR [build-imagemagick 3/3] RUN --mount=type=cache,target=/build-cache,id=imagemagick-amd64-${IMAGEMAGICK  0.6s
 => CANCELED [build-libheif 3/3] RUN --mount=type=cache,target=/build-cache,id=libheif-amd64-${LIBHEIF_VERSION}  0.7s
 => CANCELED [build-libraw 3/3] RUN --mount=type=cache,target=/build-cache,id=libraw-amd64-${LIBRAW_VERSION}     0.9s
 => ERROR [download-jellyfin-ffmpeg 3/3] RUN --mount=type=cache,target=/build-cache,id=ffmpeg-amd64-${JELLYFIN_  0.8s
------                                                                                                                
 > [build-imagemagick 3/3] RUN --mount=type=cache,target=/build-cache,id=imagemagick-amd64-${IMAGEMAGICK_VERSION}     --mount=type=secret,id=github_token     export GITHUB_TOKEN=$(cat /run/secrets/github_token)     && export $(cat /env) && ./build_imagemagick.sh:
0.505 cat: /run/secrets/github_token: No such file or directory
0.508 ./build_imagemagick.sh: line 5: IMAGEMAGICK_VERSION: unbound variable
------
------
 > [download-jellyfin-ffmpeg 3/3] RUN --mount=type=cache,target=/build-cache,id=ffmpeg-amd64-${JELLYFIN_FFMPEG_VERSION}     export $(cat /env) && ./download_jellyfin-ffmpeg.sh:
0.775 ./download_jellyfin-ffmpeg.sh: line 5: JELLYFIN_FFMPEG_VERSION: unbound variable
------
Dockerfile:47
--------------------
  46 |     COPY --chmod=0755 download_jellyfin-ffmpeg.sh .
  47 | >>> RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=ffmpeg-${TARGETARCH}-${JELLYFIN_FFMPEG_VERSION} \
  48 | >>>     export $(cat /env) && ./download_jellyfin-ffmpeg.sh
  49 |     
--------------------
ERROR: failed to solve: process "/bin/bash -euo pipefail -c export $(cat /env) && ./download_jellyfin-ffmpeg.sh" did not complete successfully: exit code: 1
```

@kkovaletp Could you take a look?

kkovaletp:
Fixes #1227 

## Summary by CodeRabbit

* **Documentation**
  * Updated the README with details on running the dependencies directory for debugging and instructions for unpacking the dependencies archive.

* **Chores**
  * Improved handling of the GITHUB_TOKEN environment variable in the build process for greater robustness.
  * Changed the base image for the release stage to enhance compatibility.
  * Added the jq package to the dependency installation process.
  * Made minor adjustments to script error handling and variable reference styles for consistency.
  * Enhanced Docker build caching by expanding cache scope to include dependency files and adjusted cache usage in test workflows.
  * Modified dependency installation to include recommended packages for improved build completeness.

### DIFF
--- a/dependencies/Dockerfile
+++ b/dependencies/Dockerfile
@@ -61,6 +61,5 @@ COPY --from=download-jellyfin-ffmpeg /output/ /output/
 COPY --chmod=0755 output.sh /
 RUN /output.sh
 
-FROM scratch AS release
+FROM debian:bookworm-slim AS release
 COPY --from=final-assembly /artifacts.tar.gz /
-USER nobody
--- a/dependencies/Dockerfile
+++ b/dependencies/Dockerfile
@@ -13,7 +13,9 @@ WORKDIR /tmp/build-libraw
 COPY --chmod=0755 build_libraw.sh .
 RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=libraw-${TARGETARCH}-${LIBRAW_VERSION} \
     --mount=type=secret,id=github_token \
-    export GITHUB_TOKEN=$(cat /run/secrets/github_token) \
+    if [ -f /run/secrets/github_token ]; then \
+    export GITHUB_TOKEN=$(cat /run/secrets/github_token); \
+    fi \
     && export $(cat /env) && ./build_libraw.sh
 
 FROM base-prep AS build-libheif
@@ -24,7 +26,9 @@ WORKDIR /tmp/build-libheif
 COPY --chmod=0755 build_libheif.sh .
 RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=libheif-${TARGETARCH}-${LIBHEIF_VERSION} \
     --mount=type=secret,id=github_token \
-    export GITHUB_TOKEN=$(cat /run/secrets/github_token) \
+    if [ -f /run/secrets/github_token ]; then \
+    export GITHUB_TOKEN=$(cat /run/secrets/github_token); \
+    fi \
     && export $(cat /env) && ./build_libheif.sh
 
 FROM base-prep AS build-imagemagick
@@ -35,7 +39,9 @@ WORKDIR /tmp/build-imagemagick
 COPY --chmod=0755 build_imagemagick.sh .
 RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=imagemagick-${TARGETARCH}-${IMAGEMAGICK_VERSION} \
     --mount=type=secret,id=github_token \
-    export GITHUB_TOKEN=$(cat /run/secrets/github_token) \
+    if [ -f /run/secrets/github_token ]; then \
+    export GITHUB_TOKEN=$(cat /run/secrets/github_token); \
+    fi \
     && export $(cat /env) && ./build_imagemagick.sh
 
 FROM base-prep AS download-jellyfin-ffmpeg
--- a/dependencies/Dockerfile
+++ b/dependencies/Dockerfile
@@ -36,6 +36,8 @@ ARG IMAGEMAGICK_VERSION
 ARG TARGETARCH
 ENV BUILD_CACHE_DIR=/build-cache
 WORKDIR /tmp/build-imagemagick
+COPY --from=build-libraw /output/ /libs/
+COPY --from=build-libheif /output/ /libs/
 COPY --chmod=0755 build_imagemagick.sh .
 RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=imagemagick-${TARGETARCH}-${IMAGEMAGICK_VERSION} \
     --mount=type=secret,id=github_token \
--- a/dependencies/Dockerfile
+++ b/dependencies/Dockerfile
@@ -18,7 +18,7 @@ RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=libraw-${TARGETARCH}-${LIBRA
     fi \
     && export $(cat /env) && ./build_libraw.sh
 
-FROM base-prep AS build-libheif
+FROM build-libraw AS build-libheif
 ARG LIBHEIF_VERSION
 ARG TARGETARCH
 ENV BUILD_CACHE_DIR=/build-cache
@@ -31,13 +31,11 @@ RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=libheif-${TARGETARCH}-${LIBH
     fi \
     && export $(cat /env) && ./build_libheif.sh
 
-FROM base-prep AS build-imagemagick
+FROM build-libheif AS build-imagemagick
 ARG IMAGEMAGICK_VERSION
 ARG TARGETARCH
 ENV BUILD_CACHE_DIR=/build-cache
 WORKDIR /tmp/build-imagemagick
-COPY --from=build-libraw /output/ /libs/
-COPY --from=build-libheif /output/ /libs/
 COPY --chmod=0755 build_imagemagick.sh .
 RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=imagemagick-${TARGETARCH}-${IMAGEMAGICK_VERSION} \
     --mount=type=secret,id=github_token \
--- a/dependencies/Dockerfile-debug
+++ b/dependencies/Dockerfile-debug
@@ -1,0 +1,2 @@
+FROM debian:bookworm-slim
+COPY --from=dependencies:self /artifacts.tar.gz /
--- a/dependencies/Dockerfile-debug
+++ b/dependencies/Dockerfile-debug
@@ -1,2 +1,0 @@
-FROM debian:bookworm-slim
-COPY --from=dependencies:self /artifacts.tar.gz /
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -1,3 +1,24 @@
 # Dependencies
 
 This directory contains scripts and Dockerfile to build third-party dependencies for Photoview. It is not intended for end-users or runtime execution.
+
+To locally debug this image you need to run 2 simple steps:
+
+1. Build the image locally and tag it as `dependencies:self`. For example:
+
+    `docker buildx build --pull --tag dependencies:self .`
+
+2. Build the debug image, which you then can run and investigate the content:
+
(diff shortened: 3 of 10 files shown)


kkovaletp:
@googollee, please try this fix. I think that it should unlock the image local build and debug process, as well as provide a consistent explanation of simple steps in the Readme.

I hope that it is better than completely reverting the previous PR with many improvements implemented. I think that it is better the debug and find the issue, causing the test to fail, on the `master` branch with this fix implemented, as we didn't see it in the feature branch, meaning that once we revert the PR, it would be difficult to find the issue.

dependencies/Dockerfile-debug
```diff
@@ -0,0 +1,2 @@
+FROM debian:bookworm-slim
```
googollee: Do you have any concern that not using `debian:bookworm-slim` as a release image? It's not a large image.

dependencies/README.md
```diff
@@ -1,3 +1,24 @@
 # Dependencies
 
 This directory contains scripts and Dockerfile to build third-party dependencies for Photoview. It is not intended for end-users or runtime execution.
+
+To locally debug this image you need to run 2 simple steps:
```
googollee: I must say, it's not simple steps. I don't like this.

dependencies/Dockerfile-debug
kkovaletp: Well, not so big concern... I just feel that it is not right to push a runnable image with just 1 archive file, which is not going to be run...
If you insist, I can get back to the default base image with the `root` user, I just thought that the current implementation is more elegant and clear, while still allowing to analyze the archive file easily (by building till the assembly target or from the debug Dockerfile).

dependencies/README.md
kkovaletp: The text doesn't express your emotions, but I feel that you're angry and don't want to try or explain why((

Because those steps are simple: 1st is what you do in any case to build the image locally. Now you need to just specify the tag for it. And the 2nd is a couple-of-seconds simple build to copy the file from 1 image to another.

googollee: The commands are straightforward, yet they introduce multiple layers that add complexity at the start of development.

When I first read the `README.md`, my initial question was: why must the image be tagged as `dependencies:self`? Could I use a different name instead? After some investigation, I noticed that this tag is referenced by `Dockerfile-debug`. Keeping read `README.md`, it led to another question: what exactly is the `final-assembly` stage? How does it differ from the previous tag?

On the other hand, the `README.md` does not clarify how to inspect the image on Docker Hub either. Simply pulling the image isn’t helpful, as it can’t be run directly. After reading through the entire `Dockerfile`, `Dockerfile-debug`, and `README.md`, I realized that I need to tag the image as `dependencies:self` and then use `Dockerfile-debug` to build a new image properly.

For this project, I hope to keep the development workflow as simple and approachable as possible. A straightforward process makes local development easier and lowers the barrier for others to contribute. If someone wants to verify the image, the simplest approach should be to pull and run it directly. If it fails to run, they may give up at this point. Even if they wish to dig deeper, they’ll still need to spend extra time understanding how the files work together.

I didn't see any problem making this image runnable. If we care about the size of the image very much, I suggest using `busybox` as the base release image.

Sorry, I got a bit emotional this afternoon.

dependencies/Dockerfile-debug
kkovaletp: I've reverted the base image and the user for the final image

dependencies/README.md
kkovaletp: Ok, thanks for your explanation. I really thought that this config was simple to use and straightforward to debug for everyone with Docker basics. Also, I'm a perfectionist, and sometimes go too far with optimizations, so this configuration change doesn't introduce noticeable performance or security degradation, and I reverted it back to the Debian image and root user.

dependencies/build_imagemagick.sh
```diff
@@ -1,14 +1,15 @@
...
-set -euo pipefail
 
 # Fallback to the latest version if IMAGEMAGICK_VERSION is not set
-if [[ -z "$IMAGEMAGICK_VERSION" ]]; then
+if [[ -z "${IMAGEMAGICK_VERSION}" ]]; then
   echo "WARN: ImageMagick version is empty, most likely the script runs not on CI."
   echo "Fetching the latest version from ImageMagick repo..."
   IMAGEMAGICK_VERSION=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
     "https://api.github.com/repos/ImageMagick/ImageMagick/releases/latest" | jq -r '.tag_name')
 fi
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, as far as I understand the `$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60  "https://api.github.com/repos/ImageMagick/ImageMagick/releases/latest" | jq -r '.tag_name')` command, in the case of some error on the CURL side, it won't fail silently, it will output the exact error to the STDERR. Additionally, the `jq` command will get the empty string and fail with a parse error, which is also printed to the log, causing the script to fail at this point and not run anything else. This means that there is no need for the `set -e` for this code. Am I right?

kkovaletp:
@googollee, I think that I found and fixed the issue with the test failure: it was caused by ImageMagick, missing both LibHeif and LibRAW at its building time. 
I tried to run those 2 build stages for the libs in parallel, and then copy the results to the IM build stage in the same way, as we do in the main image, but not all IM files were fixed, so I had to get back to sequential build for these 3 dependencies, as it was originally implemented (but with the file-level cache per dependency).
Now tests still fail, as they use the dependencies image from Docker Hub, not the one from this PR.

BTW, during this issue investigation, I noticed the following strange inconsistency, existing even before my refactoring was merged:
https://github.com/photoview/photoview/blob/73ae3e8132b89e2f8f1f90950ddd5d3cfbebc56e/dependencies/prepare.sh#L26
Here you define the `PKG_CONFIG_PATH` variable, but later, after the build, you don't use it, and instead, copy the files from the static `/usr/local/lib/pkgconfig` folder:
https://github.com/photoview/photoview/blob/73ae3e8132b89e2f8f1f90950ddd5d3cfbebc56e/dependencies/build_imagemagick.sh#L35-L37
https://github.com/photoview/photoview/blob/73ae3e8132b89e2f8f1f90950ddd5d3cfbebc56e/dependencies/build_libheif.sh#L24
https://github.com/photoview/photoview/blob/73ae3e8132b89e2f8f1f90950ddd5d3cfbebc56e/dependencies/build_libraw.sh#L25
However, later in the main image, you restore these files to the path from the variable:
https://github.com/photoview/photoview/blob/73ae3e8132b89e2f8f1f90950ddd5d3cfbebc56e/Dockerfile#L75

Is this done intentionally? Why don't we use the same path to store those files after the build and to install them in the main image?

dependencies/Dockerfile
```diff
@@ -13,29 +13,35 @@ WORKDIR /tmp/build-libraw
 COPY --chmod=0755 build_libraw.sh .
 RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=libraw-${TARGETARCH}-${LIBRAW_VERSION} \
     --mount=type=secret,id=github_token \
-    export GITHUB_TOKEN=$(cat /run/secrets/github_token) \
+    if [ -f /run/secrets/github_token ]; then \
```
googollee, line 16: `build_*.sh` doesn't use GITHUB_TOKEN now. Do you want to keep this for future use?

googollee (approved):

googollee:
> Is this done intentionally? Why don't we use the same path to store those files after the build and to install them in the main image?

Yes, it's intentional. Since we compile those packages, they can't follow Debian package rules to put `*.pc` under `/usr/lib/${DEB_HOST_MULTIARCH}/pkgconfig`(`${PKG_CONFIG_PATH}`) for cross-compiling. Another way of fixing it is by adding `/usr/local/lib/pkgconfig` to the pkg-config path when building the main repo. Adding pkg-config path with CGO is hard to get right, so I just put those `*.pc` files in the default Debian cross-compile directory.

dependencies/Dockerfile
kkovaletp: Actually, they use it (if available) for downloading tarballs from api.github.com to overcome the request rate limit:
https://github.com/photoview/photoview/blob/0192e4e64020f383d8a214707e1aee14e4a22bc7/dependencies/build_imagemagick.sh#L49
https://github.com/photoview/photoview/blob/0192e4e64020f383d8a214707e1aee14e4a22bc7/dependencies/build_libheif.sh#L41
https://github.com/photoview/photoview/blob/0192e4e64020f383d8a214707e1aee14e4a22bc7/dependencies/build_libraw.sh#L37
The download FFmpeg script doesn't use it, as it gets the file from the regular github.com, where there is no such limit.

I didn't change this code. I just made sure that it doesn't cause failures and silently skipped in the case of local runs, where the token is not provided.

kkovaletp:
@googollee, when you do the final code review, please expand the unchanged code in the affected files and review all of it together - to make sure that now it is consistent and I didn't miss anything important)

googollee:
@kkovaletp could you verify that the build script is working well without GITHUB_TOKEN?

kkovaletp:
@googollee, yes, it works fine (as far as I see):
I executed the `docker buildx build --security-opt seccomp=unconfined --pull --tag dependencies:self .` from the `dependencies` folder on the latest commit of this branch in a new shell tab, and here is the importable docker build result for the run:
[deps-local.dockerbuild.zip](https://github.com/user-attachments/files/20966633/deps-local.dockerbuild.zip)
Unzip it to have the `deps-local.dockerbuild` file and import it into your Docker Desktop UI on the Builds page.

googollee:
LGTM. Tests should be fixed with the new dependencies image.

PR #1232, opened 2025-06-30
Make TaskContext as context.Context.

googollee:
Part of #1208

This PR allows context passed between the queue and workers.

### DIFF
--- a/api/scanner/scanner_task/scanner_task.go
+++ b/api/scanner/scanner_task/scanner_task.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"flag"
 	"io/fs"
+	"time"
 
 	"github.com/photoview/photoview/api/graphql/models"
 	"github.com/photoview/photoview/api/scanner/media_encoding"
@@ -99,4 +100,8 @@ func (c TaskContext) Done() <-chan struct{} {
 
 func (c TaskContext) Err() error {
 	return c.ctx.Err()
+}
+
+func (c TaskContext) Deadline() (time.Time, bool) {
+	return c.ctx.Deadline()
 }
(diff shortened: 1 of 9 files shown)


api/scanner/scanner_utils/scanner_error.go
```diff
@@ -1,18 +1,19 @@
...
+	"github.com/photoview/photoview/api/log"
 	"github.com/photoview/photoview/api/utils"
 )
 
-func ScannerError(format string, args ...interface{}) {
+func ScannerError(ctx context.Context, format string, args ...any) {
 	message := fmt.Sprintf(format, args...)
 
-	log.Printf("ERROR: %s", message)
+	log.Error(ctx, "ERROR: %s", message)
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/scanner_task/scanner_task.go
```diff
@@ -100,3 +101,7 @@ func (c TaskContext) Done() <-chan struct{} {
 func (c TaskContext) Err() error {
 	return c.ctx.Err()
 }
+
+func (c TaskContext) Deadline() (time.Time, bool) {
```
kkovaletp: @googollee, is this new function used somewhere? Did I miss the usage?
If you plan to use it in your future PRs, doesn't it make sense to introduce it in the PR, where it is going to be used?

googollee: It's a required method by [`context.Context`](https://pkg.go.dev/context#Context). To use `TaskContext` as a normal `context.Context` in log, it's required.

googollee: I updated the `TaskContext` with an embedded `context.Context`. Now those methods of `context.Context` are exposed to `TaskContext` directly, and no need those proxy methods. It should be more clear than before.

kkovaletp: Yes, now it looks clearer, thanks.

PR #1235, opened 2025-07-01
New queue

### LINKED ISSUE #1208: Refactor the Scan job code

googollee:
Fix  #1208.

Try with `ghcr.io/googollee/photoview:new-queue-1208` first.

## Summary by CodeRabbit

* **New Features**
  * Introduced a new, robust background job queue system for scanning photo albums, supporting concurrent processing, dynamic worker scaling, and periodic triggers.
  * Added automatic blurhash generation for media thumbnails during scanning.

* **Improvements**
  * Unified scanner queue and periodic scanner into a single, streamlined queue system for better performance and maintainability.
  * Enhanced error handling and logging throughout the scanning and media processing tasks.

* **Bug Fixes**
  * Improved test reliability by initializing flag sets to prevent conflicts during test execution.

* **Chores**
  * Updated test scripts to enable Go's race detector for improved concurrency safety.
  * Removed obsolete code and tests related to the old scanner queue and periodic scanner implementations.

### DIFF
--- a/api/scanner/scanner_tasks/blurhash_task.go
+++ b/api/scanner/scanner_tasks/blurhash_task.go
@@ -27,7 +27,7 @@ func (t BlurhashTask) AfterProcessMedia(ctx scanner_task.TaskContext, mediaData 
 	}
 
 	var media *models.Media
-	if err := ctx.GetDB().Where("id = ?", mediaData.Media.ID).First(&media).Error; err != nil {
+	if err := ctx.GetDB().Preload("MediaURL").Where("id = ?", mediaData.Media.ID).First(&media).Error; err != nil {
 		return fmt.Errorf("failed to get media(id:%d): %w", mediaData.Media.ID, err)
 	}
 


api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+	"slices"
+	"sync"
+	"time"
+
+	"github.com/photoview/photoview/api/log"
+)
+
+type commonJob interface {
+	Key() int
+	String() string
```
kkovaletp, line 18: @googollee, can you please give it a better name? `String` doesn't explain anything.

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+	defer func() {
+		q.wait.Wait()
+		log.Info(q.ctx, "closed queue", "remain_jobs", q.lenJobs())
+	}()
+
+	log.Info(q.ctx, "closing queue")
+	q.trigger.Stop()
+	close(q.done)
+	if err := q.RescaleWorkers(0); err != nil {
+		log.Error(q.ctx, "closing all workers error", "error", err)
```
kkovaletp: @googollee, the message doesn't sound so good. Maybe, the "failed to close all workers" is better...

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+func (q *commonQueue[Job]) RunBackground() {
+	q.wait.Add(1)
+	go q.run()
+}
+
+// ConsumeAllBacklog waits all jobs to be done in the queue backlog. It doesn't require `RunBackground`.
+// This function is useful with unit tests.
+func (q *commonQueue[Job]) ConsumeAllBacklog(ctx context.Context) {
+	for {
+		job, ok := q.popBacklog()
```
kkovaletp, line 101: @googollee, why do you use here the reversed logic for the function status detection? Usually, we have the `err` var instead of `ok`...

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+		if !ok {
+			log.Info(q.ctx, "consumed all backlog")
+			return
+		}
+
+		log.Info(q.ctx, "consuming all backlog", "job", job)
```
kkovaletp: @googollee, this line's message looks too similar to the one from line 98. Please rephrase it, because they are confusing now.

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+			q.pushBacklog(job)
+			return
+		}
+	}
+}
+
+// UpdateScanInterval updates the interval of background periodic jobs.
+func (q *commonQueue[Job]) UpdateScanInterval(newInterval time.Duration) error {
+	if newInterval < 0 {
+		return fmt.Errorf("invalid periodic scan interval(%d): must >=0", newInterval)
```
kkovaletp: @googollee, just wording: "must be >= 0" (be, space before 0)

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	if newInterval == 0 {
+		q.trigger.Stop()
+		return nil
+	}
+
+	q.trigger.Reset(newInterval)
```
kkovaletp, line 129: @googollee, don't we need to catch and handle the error here?

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+	return nil
+}
+
+// RescaleWorkers rescales the number of background workers.
+func (q *commonQueue[Job]) RescaleWorkers(newMax int) error {
+	if newMax < 0 {
+		return fmt.Errorf("invalid concurrent workers (%d): must >=0", newMax)
```
kkovaletp: @googollee, same wording: "must be >= 0"

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	q.workersMu.Lock()
+	defer q.workersMu.Unlock()
+
+	defer func() {
+		log.Info(q.ctx, "rescaled worker", "worker_number", len(q.workers))
```
kkovaletp: @googollee, "rescaled worker**s**", same for the "worker**s**_number"

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+	if len(q.workers) > newMax {
+		closing := q.workers[newMax:]
+		if newMax == 0 {
+			q.workers = nil
+		} else {
+			q.workers = q.workers[:newMax]
+		}
+
+		for _, worker := range closing {
+			worker.Close()
```
kkovaletp, line 159: @googollee, if we downscale workers (let's say, from 5 to 3), this code closes all 5 previously started workers. Why? Isn't it more efficient to close only the last 2 and let the other 3 run further?

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+
+		return nil
+	}
+
+	// len(q.workers) < newMax
+	q.workers = slices.Grow(q.workers, newMax-len(q.workers))
+	for len(q.workers) < newMax {
+		worker := newWorker(log.WithAttrs(q.ctx, "worker_id", len(q.workers)), q.handout, q)
+		q.wait.Add(1)
+		go worker.Run()
```
kkovaletp, line 170: @googollee, if a runtime error occurs at the new worker's start, what is going to happen? Don't we need to catch and handle such errors here?
Does it run fully async with no delay in the main code?

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+		q.workers = append(q.workers, worker)
+	}
+
+	return nil
+}
+
+func (q *commonQueue[Job]) run() {
+	defer q.wait.Done()
+	defer log.Info(q.ctx, "queue background done")
```
kkovaletp: @googollee, why 2 separate `defer` statements instead of 1 with these 2 lines?

api/scanner/queue/common.go
```diff
@@ -0,0 +1,297 @@
...
+	q.jobsMu.Lock()
+	defer q.jobsMu.Unlock()
+
+	return len(q.backlog) + len(q.ongoing)
+}
+
+func (q *commonQueue[Job]) processJob(ctx context.Context, job Job) {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Error(ctx, "panic when process job", "job", job, "panic", r)
```
kkovaletp: @googollee, does the "panic happened during job processing" sound better?

api/scanner/queue/common.go
kkovaletp, line 1: @googollee, I see many lines are not covered by unit tests in the `queue` package (3 code files). Please review the coverage and, if it wouldn't require too much effort, extend the coverage. This is 1 of the critical code modules in the project, and if something happens, we'd have critical bugs for the product...

Another notice: from reading the code, it is not always clear which functions are going to run inside the main server thread, and which in a worker thread (if my assumption that there are some functions outside of the `worker.go`, running in a worker thread, is correct). It would be good to somehow indicate this for all multithread code in a function's name or in its docstring.

And 1 more: from the code, it looks like we now have less logging of the scan queue run process than we had right before this reimplementation. If my assumption is right, can you please review the end-to-end logging flow and add missing log messages to have at least a similar level of transparency, as before?

And finally, am I right to assume that the main thread and workers' threads would write log messages and errors to the same STDOUT/STDERR? If yes, does it handle concurrent writes automatically, or do we need to handle them as well?

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,157 @@
...
+
+// NewQueue creates a new Queue. It reads `db` to get the number of workers and the interval of triggering periodic jobs.
+func NewQueue(db *gorm.DB) (*Queue, error) {
+	siteInfo, err := models.GetSiteInfo(db)
+	if err != nil {
+		return nil, fmt.Errorf("can't get site info: %w", err)
+	}
+
+	if siteInfo.PeriodicScanInterval < 0 {
+		return nil, fmt.Errorf("invalid periodic scan interval (%d): must >=0", siteInfo.PeriodicScanInterval)
```
kkovaletp: @googollee, same wording: "must be >= 0"

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	interval := time.Duration(siteInfo.PeriodicScanInterval) * time.Second
+	var ticker *time.Ticker
+	if interval == 0 {
+		ticker = time.NewTicker(time.Second) // The interval is not matter since the ticker is stopped.
+		ticker.Stop()
+	} else {
+		ticker = time.NewTicker(interval)
+	}
```
kkovaletp: @googollee, this code is the same as in the `common.go`. Does it make sense to inherit or reuse it from there?

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,157 @@
...
+
+	if siteInfo.ConcurrentWorkers < 0 {
+		return nil, fmt.Errorf("invalid concurrent workers (%d): must >=0", siteInfo.ConcurrentWorkers)
```
kkovaletp: @googollee, same wording: "must be >= 0"

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,157 @@
...
+func (q *Queue) processJob(ctx context.Context, job *queueJob) {
+	log.Info(ctx, "process album", "album", job.album.Title)
+
+	task := scanner_task.NewTaskContext(ctx, q.db, job.album, job.cache)
+	if err := scanner.ScanAlbum(task); err != nil {
+		scanner_utils.ScannerError(ctx, "Failed to scan album: %v", err)
+	}
+}
+
+func (q *Queue) periodicTrigger(ctx context.Context) {
```
kkovaletp: @googollee, the function name is confusing, as a trigger typically is a signal to run a function and is not a function itself

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,157 @@
...
+	if err := q.AddAllAlbums(ctx); err != nil {
+		log.Error(ctx, "fill queue error", "error", err)
```
kkovaletp: @googollee, does the "failed to add all albums to the queue" sound better?

api/scanner/queue/queue.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+}
+
+// helpers
+func (q *Queue) findAllAlbumsJobs() ([]*queueJob, error) {
+	log.Info(q.ctx, "find all job")
```
kkovaletp: @googollee, what is the purpose of this log message? Please rephrase it, as I don't get it.

api/scanner/queue/worker.go
```diff
@@ -0,0 +1,52 @@
+package queue
+
+import (
+	"context"
+
+	"github.com/photoview/photoview/api/log"
+)
+
+type workerCallback[Job any] interface {
```
kkovaletp, line 9: @googollee, why do you have `Job any` in several definitions? Isn't it better to restrict it to the correct type of job?

api/scanner/queue/worker.go
```diff
@@ -0,0 +1,52 @@
...
+		ctx:      ctx,
+		input:    input,
+		done:     make(chan struct{}),
+		callback: callback,
+	}
+}
+
+func (w *worker[Job]) Close() {
+	log.Info(w.ctx, "closing worker")
+	close(w.done)
```
kkovaletp, line 33: @googollee, don't we need to ensure the current task is completed and the worker is safe to be closed before closing it?

api/scanner/queue/worker.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (w *worker[Job]) Run() {
+	defer w.callback.finish(w.ctx)
+	defer log.Info(w.ctx, "worker done")
```
kkovaletp: @googollee, same question: why 2 separate defers instead of 1 with both lines?

api/scanner/queue/common.go
googollee: `String() string` is a common method in Go to describe an instance, see [Stringer](https://pkg.go.dev/fmt#Stringer).

api/scanner/queue/common.go
googollee: It's not a reversed logic. Also, there is no error when backlog is empty. It's also a common pattern to use `xxx, ok` to check the validation, see [`t, ok := x.(T)`](https://go.dev/ref/spec#Type_assertions), [`x, ok = <-ch
`](https://go.dev/ref/spec#Receive_operator) and [`v, ok = a[x]`](https://go.dev/ref/spec#Index_expressions).

api/scanner/queue/common.go
googollee: Done.

api/scanner/queue/common.go
googollee: Updated.

api/scanner/queue/common.go
googollee: Updated.

api/scanner/queue/common.go
googollee: No. The code already ensure `newInterval` is a valid value. If it still panics, let the process crash loudly.

api/scanner/queue/common.go
googollee: Updated.

api/scanner/queue/common.go
googollee: It won't close all 5 workers. `closing := q.workers[newMax:]` and it contains 2 workers only.

api/scanner/queue/common.go
googollee: > if a runtime error occurs at the new worker's start, what is going to happen?

What kind of errors? I can't find any errors when worker is launching.

> Does it run fully async with no delay in the main code?

What do you mean `fully async` here?

api/scanner/queue/common.go
googollee: Well, no special reason. Merged in one function.

api/scanner/queue/common.go
googollee: Updated.

api/scanner/queue/queue.go
googollee: Updated.

api/scanner/queue/queue.go
googollee: Good catch. This code moved to `common.go` and I forgot removing them here. Updated.

api/scanner/queue/queue.go
googollee: Updated.

api/scanner/queue/queue.go
googollee: Updated.

api/scanner/queue/worker.go
googollee: Here is a generic worker. `Job` is a name of the job type, which could be any type. See [Generics](https://go.dev/doc/tutorial/generics).

api/scanner/queue/worker.go
googollee: The worker finishs current job before quiting `Run()`. It does what you want.

api/scanner/queue/worker.go
googollee: Merged.

api/scanner/queue/common.go
googollee: For the coverage, I believe `common_test.go` covers most of `common.go` and `worker.go`. I don't know how codecov calculates here.

For example, codecov says `UpdateScanInterval()` is [not covered at all](https://app.codecov.io/gh/photoview/photoview/commit/2a06e563acb452ee0af7f8a7664195c1a8811115/blob/api/scanner/queue/common.go?dropdown=coverage#L112): 

image

But in `common_test.go`, there are [`TestCommonQueueBackgroundPeriod`](https://github.com/photoview/photoview/pull/1235/files#diff-e106d987d551dd75f7553bad253c2ecd7f3829bfea703da7d99da4a142488074R94) and [`TestCommonQueueInvalidCase/UpdateWithInvalidInterval`](https://github.com/photoview/photoview/pull/1235/files#diff-e106d987d551dd75f7553bad253c2ecd7f3829bfea703da7d99da4a142488074R206). When I ran the coverage report locally, the report shows they cover all code of UpdateScanInterval()`:

image.

Since I'm not familiar with codecov, I just leave it there. I may go back to it in the following PR.

googollee: > from reading the code, it is not always clear which functions are going to run inside the main server thread, and which in a worker thread (if my assumption that there are some functions outside of the worker.go, running in a worker thread, is correct). It would be good to somehow indicate this for all multithread code in a function's name or in its docstring.

Basically, there is no code running in the main thread. Do you have any examples of what you want to see?

> it looks like we now have less logging of the scan queue run process than we had right before this reimplementation.

Could you point out what logs are still missing? Since the code is changed, the workflow are not exactly same. If you find any missing log, let me know.

> I right to assume that the main thread and workers' threads would write log messages and errors to the same STDOUT/STDERR?

They will write log to the same stdout/stderr. `slog` library will ensure that `The built-in handlers acquire a lock before calling [io.Writer.Write](https://pkg.go.dev/io#Writer.Write) to ensure that exactly one [Record](https://pkg.go.dev/log/slog#Record) is written at a time in its entirety.`, see [Performance considerations](https://pkg.go.dev/log/slog#hdr-Performance_considerations).

api/scanner/queue/common.go
kkovaletp: Maybe I miss the point, but it looks descriptive for the Stringer type, as it operates with the string. As far as I understood from the code, the title of an album is going to be sent here. Isn't it better to name it a Title then?

Alternatively, you can add a docstring with an explanation of the interface and its methods.

api/scanner/queue/common.go
kkovaletp: > Basically, there is no code running in the main thread. Do you have any examples of what you want to see?

Maybe I misunderstood the architecture, but as far as I understand, the queue management and task management code runs in the main server thread/process, and the task processing code runs in a worker thread/process. For example, init a queue, rescale workers, populate the queue with tasks, etc - is going to run by the main process; and the code to scan folders, files, detect faces, etc - runs by worker processes, as well as the init/close a worker, pick up a task from the queue, etc... Am I correct? Are these examples clear to explain my initial comment?

> Could you point out what logs are still missing?

Sure, I'll test the new implementation and compare the job log with the one from the old implementation. I'll write you if I find some missing parts.

api/scanner/queue/common.go
kkovaletp: I see you added the space, but did not add the "be". Don't you like it? Why?

api/scanner/queue/common.go
kkovaletp: > What kind of errors?

As this is a new process or thread start, there might be some OS-level runtime errors, like not enough memory for the new process, or no free PID, etc...

> What do you mean `fully async` here?

From my understanding, when executing this line, the main process should spin up a new process, give it the context (a set of all variables and objects), the task (code) to run, and the communication channel to return the execution result to the main process. Then the main process keeps running the following lines independently with no delays or any other synchronizations with the new worker, until it receives a status update and/or a request for a new task. This means "fully async run" for me. If that is not the way it runs and there are some other delays, syncs, or other types of interruptions, please let me know how it works and what might be bottlenecks.

api/scanner/queue/queue.go
kkovaletp: Thanks, now I understand the purpose. Actually, the function finds all albums and creates jobs for the found albums. It doesn't find jobs. So, I'd propose the "find albums for all users and create scanner jobs for found albums" message.

api/scanner/queue/worker.go
kkovaletp: Thanks, I see. But in the example, there are 2 types declared, separated by the `|`. Can you predict all possible types and list them in the declaration in the same way? Or, is it better to have `any` placeholder?
From TypeScript, I know that it is always better to define strict types instead of `any`, because usually `any` hides type mismatch defects, which might cause unexpected code behavior later and might be difficult to catch.

api/scanner/queue/common.go
googollee: This function is useful when logging a job, as `logging.Info(ctx, "message", "job", job)`. Go library calls `String()` to get the description of the object. The return is the description of the object, not the title. It's a common pattern in Go.

I added some comments here.

api/scanner/queue/common.go
googollee: > As this is a new process or thread start, there might be some OS-level runtime errors, like not enough memory for the new process, or no free PID, etc...

Don't handle any unexpected errors. Just let the process panic loudly.

`go worker.Run()` will create a new goroutine and push it to the backlog of Go runtime, and then the main process continues. That new goroutine may run immediately if there is a free thread (called M in Go runtime), or wait until there is a free thread(maybe the main thread) to run, or maybe the main process runs out of its time slice and the main thread switches to this new goroutine immediately. No matter which case, there is no result returned.

You can search with `Golang concurrency model` to find a lot of articles explaining it.

api/scanner/queue/worker.go
googollee: Here I don't expect to limit the `Job` type. No matter which type it is, it should provide `processJob(ctx context.Context, job Job)` to process that job. The only limitation is the type `Job` is the same as the type in `processJob(ctx context.Context, job Job)` parameter.

api/scanner/queue/queue.go
googollee: Updated.

kkovaletp:
@googollee, you didn't put a reaction on the bot's nitpick comment https://github.com/photoview/photoview/pull/1235#pullrequestreview-2976456220, so I don't know whether you reviewed it or not. Please review that comment as well

api/scanner/queue/common.go
kkovaletp: Thanks for the explanation, now I understand it a bit better

api/scanner/queue/queue.go
googollee: Updated the log.

api/scanner/queue/common.go
kkovaletp: @googollee, I've tested the new and previous (from the `master`) job implementations and found a list of issues. I'll resolve this thread and create new threads per issue to let us have separate discussions for each issue.

kkovaletp, line 1: @googollee, when the server starts on a clean DB, the next error is thrown in the log during the startup process. It doesn't happen for the `master` image in the same case:
```
2025-07-05 11:22:38.926 | 2025/07/05 11:22:38 /app/api/graphql/models/site_info.go:39 record not found
2025-07-05 11:22:38.926 | [1.797ms] [rows:0] SELECT * FROM "site_info" ORDER BY "site_info"."initial_setup" LIMIT 1
```
Here is the full init log:
```
2025-07-05 11:22:38.269 | 2025/07/05 11:22:38 Starting Photoview...
2025-07-05 11:22:38.269 | 2025/07/05 11:22:38 No .env file found. If Photoview runs in Docker, this is expected and correct.
2025-07-05 11:22:38.271 | 2025/07/05 11:22:38 INFO Found magickwand worker: ImageMagick 7.1.1-47 Q16-HDRI x86_64 c8f4e8cb7:20250329 https://imagemagick.org version=1809
2025-07-05 11:22:38.337 | 2025/07/05 11:22:38 INFO Found executable worker: ffmpeg version="ffmpeg version 7.1.1-Jellyfin Copyright (c) 2000-2025 the FFmpeg developers" codec=h264
2025-07-05 11:22:38.359 | 2025/07/05 11:22:38 INFO Found ffprobe path=/usr/local/bin/ffprobe version="ffprobe version 7.1.1-Jellyfin Copyright (c) 2007-2025 the FFmpeg developers"
2025-07-05 11:22:38.359 | 2025/07/05 11:22:38 Utilizing postgres database driver based on environment variables
2025-07-05 11:22:38.792 | 2025/07/05 11:22:38 Migrating `media_exif.flash` from string to int
2025-07-05 11:22:38.926 |
2025-07-05 11:22:38.926 | 2025/07/05 11:22:38 /app/api/graphql/models/site_info.go:39 record not found
2025-07-05 11:22:38.926 | [1.797ms] [rows:0] SELECT * FROM "site_info" ORDER BY "site_info"."initial_setup" LIMIT 1
2025-07-05 11:22:38.928 | 2025/07/05 11:22:38 INFO rescaled workers process=queue workers_number=3
2025-07-05 11:22:38.928 | 2025/07/05 11:22:38 INFO worker start process=queue worker_id=0
2025-07-05 11:22:38.928 | 2025/07/05 11:22:38 INFO worker start process=queue worker_id=1
2025-07-05 11:22:38.928 | 2025/07/05 11:22:38 INFO worker start process=queue worker_id=2
2025-07-05 11:22:38.929 | 2025/07/05 11:22:38 INFO queue background start process=queue
2025-07-05 11:22:38.929 | 2025/07/05 11:22:38 Found exiftool
2025-07-05 11:22:38.929 | 2025/07/05 11:22:38 Initializing face detector
2025-07-05 11:22:39.760 | 2025/07/05 11:22:39 Photoview API endpoint listening at http://0.0.0.0:80/api
2025-07-05 11:22:39.760 | 2025/07/05 11:22:39 Photoview API public endpoint ready at /api
2025-07-05 11:22:39.760 | 2025/07/05 11:22:39 Photoview UI public endpoint ready at /
```

kkovaletp, line 1: @googollee, the new queue implementation is almost not integrated with the GUI notifications: it shows only scanning errors, but not job start, progress, and job finish notifications. Here are 2 examples, but I'm not sure that this is the complete list of missing notifications:

SCR-20250705-kuwt
SCR-20250705-kvek

Please review the list of notifications, GUI expects, and implement them on the Backend side.

kkovaletp, line 1: @googollee, I've compared the log from `master` and this implementations, and here is the list of missing log messages. Is it ok to have all of them listed in 1 conversation thread, or would you prefer to extract some of them to separate threads for longer discussions?

- Queue status change events, like "Queue waiting", "Queue active", "Queue waiting for lock"
- Periodic scan interval changes, like "Periodic scan interval changed: disabled", "Periodic scan interval changed: 24h0m0s"
- Scan interval runner statuses, like: "Scan interval runner: Waiting for signal", "Scan interval runner: New ticker detected", "Scan interval runner: Starting periodic scan"
- Generating blurhashes at the end of the media scanning and face recognition
- The backlog size is missing in the handout log messages: it will show the progress of the job run in the log
- `pushBacklog()` and `popBacklog()` don't log their invocations. I'm not sure that it is needed: if there is higher-level logging of adding a job to the backlog and there is no real difference, by which one of these 2 functions it is done, then probably we could skip it for now. But if it might be useful for potential debugging, then it would be good to have it

kkovaletp, line 1: @googollee, I calculated and compared job run times between the `master` job and the current implementation on my small testing dataset for the following 3 use cases:
- init - clean DB, setup Photoview from scratch, 1st full scan for all users (while there is only 1 user registered) with 3 workers
- rescan - after the init run is finished, rescale to 5 workers, enable periodic scan, and set 3 min time, wait for the job to start automatically
- user - when the rescan is done, rescale back to 3 workers, disable periodic interval, and run the scan for the user manually

SCR-20250705-mqep

As you can see, the new implementation is faster (in ~0.5-1 sec) for both rescans, where there is nothing to process, and the time is spent on the queue's management code. But it is noticeably slower (in ~3 sec) for the init scan, where the jobs actually process the incoming data and generate outputs. This is a huge difference for so a small dataset. I don't have large datasets to run the testing on, but I expect a proportional increase. 
This is a big problem, as we have users with really large datasets, which they have to scan for the 1st time for more than a day.

Can you please take a look on the performance aspect of the implementation and optimize it?

api/scanner/queue/common.go
googollee: Updated.

api/scanner/queue/common.go
googollee: Updated.

api/scanner/queue/queue.go
googollee: Updated.

api/scanner/queue/queue.go
googollee: Updated.

api/scanner/queue/common.go
googollee: It's reasonable as a clear db not containing any site_info. The code `First()` to find only one record is clearer than `Limit(1).Find(&array)` and then checking array length.

googollee: There is no `Queue waiting`, `Queue active` and `Queue waiting for lock` now. The queue is always active, waiting jobs to process.

There is no new ticker, no signal to create a new ticker.

The handout log is a worker's log. A worker should not know the information of backlog. So it can't print the length of backlog in the log.

Added interval change log and the log of the backlog length.

googollee: Because this is all done in my free time, I'm afraid that I don't have enough time to optimize the performance of every detail. I try to make it in the right way. Since it's background jobs, I don't put any time on the performance.

googollee: One possible performance difference is in [face detection](https://github.com/photoview/photoview/pull/1235/files#diff-a11eededf9a0dd34deedebfa0e8cbcefc576df06f6c75d3c67068dab99496b70R21). In master, the face detection won't be counted since it runs in another goroutine, but with this PR it's counted. I don't think a goroutine is necessary here since face detection is part of background job, especially this goroutine is kind of out-of-control goroutine.

api/scanner/scanner_tasks/blurhash_task.go
```diff
@@ -0,0 +1,76 @@
...
+
+	imageFile, err := os.Open(thumbnail_path)
+	if err != nil {
+		return "", fmt.Errorf("open %q error: %w", thumbnail_path, err)
+	}
+
+	imageData, _, err := image.Decode(imageFile)
+	if err != nil {
+		return "", fmt.Errorf("decode %q error: %w", thumbnail_path, err)
+	}
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/queue/common.go
kkovaletp: I understand this and I agree with you from the technical side.
However, I was getting a lot of questions and false issue reports regarding the "No .env file found" log message during the Photoview start process. That is why I added the "If Photoview runs in Docker, this is expected and correct." explanation with the hope to make this message more clear and "fix" this misunderstanding.

Now, when we're going to report an error every time the product starts for the 1st time (even while it is expected), I'm afraid to get a lot more issues and questions about this...
So, if we know the exact case, when it might happen, and that in this case it is not an error from the business process PoV, can we handle it in a way, not making an impression that there is a problem?

kkovaletp: > There is no `Queue waiting`, `Queue active` and `Queue waiting for lock` now. The queue is always active, waiting jobs to process.

Ok, I see. Then, it should be clear from your newly added backlog length message when the scan task is finished. Can you please log the Periodic scan interval and user manual request triggers (to run a scan for all users and for the particular user) - they would clearly indicate the time and the cause for the scan task to start. Yes, later there is a log message "find albums for all users and create scanner jobs" and "find albums for a user and create scanner jobs", but I'm not sure that this is the real starting point for the task. I mean by "task" the single iteration of all scan workflow with different jobs)

> * Generating blurhashes at the end of the media scanning and face recognition

This one looks to be still not answered, right?

> The handout log is a worker's log. A worker should not know the information of backlog. So it can't print the length of backlog in the log.

That wasn't clear for me. This is a good example ot 1 of our previous discussions about labeling different types of queue log messages to clearly show on which side of the queue they happen: on a worker (with worker ID) or on the queue manager side. I see that you added the backlog size logging, so the initial purpose is solved.

> * `pushBacklog()` and `popBacklog()` don't log their invocations. I'm not sure that it is needed: if there is higher-level logging of adding a job to the backlog and there is no real difference, by which one of these 2 functions it is done, then probably we could skip it for now. But if it might be useful for potential debugging, then it would be good to have it

This one looks to be not answered as well. If no logging is needed here, just let me know. No response doesn't clarify anything)

kkovaletp: I see and I appreciate your contribution.
However, there is no rush to finish this PR ASAP and merge it. If you need some extra time to optimize things and make them run better, I guess that this would be the right reason to wait for it some more time...

Yes, this is a background job, not blocking users from using the GUI. But still, if a user has a media library with tens or even hundreds of TB, a full scan time becomes crucial for such cases. The new implementation should not make it longer. Hopefully, it can speed it up a bit because of a better job assignment and initial folders scanning in parallel. Maybe, because of some other optimizations...

> One possible performance difference is in [face detection](https://github.com/photoview/photoview/pull/1235/files#diff-a11eededf9a0dd34deedebfa0e8cbcefc576df06f6c75d3c67068dab99496b70R21).

It could be. I guess the main point of running it in another goroutine was to utilize the multi-core/multi-threading capabilities of modern CPUs to let it run in parallel to other scan jobs and the main process.

googollee: > It could be. I guess the main point of running it in another goroutine was to utilize the multi-core/multi-threading capabilities of modern CPUs to let it run in parallel to other scan jobs and the main process.

There are already multiple workers to utility multi cores. It's not a necessary to do it again.

googollee: > Generating blurhashes at the end of the media scanning and face recognition

I added log of blurhash generating. I didn't see previous queue has any log about face recognition. Is there an example?

> `pushBacklog()` and `popBacklog()` don't log their invocations.

They are not part of the queue logic, just helper functions. I don't have ideas about what kind of log it should provide.

googollee: OK. Rolled back `GetSiteInfo()`.

googollee: I rolled back the face detection task. Could you test the init scan again and compare with `master`? Thanks.

BTW, this rollback can't pass tests because race detection is enabled with `-race` in this PR.

googollee:
I feel that there are some works that need to be done before switching to the new queue, like testing with blurhash and getting the performance data. I'll close this one now and work on them first.

kkovaletp:
Sure

PR #1239, opened 2025-07-05
Fix GitHub commit passing, UI build and cache for different purposes

kkovaletp:
During today's testing of the new queue, I noticed that the commit hash wasn't correctly passed to the UI build after my previous Docker build refactoring. Unfortunately, it is not possible to fix it in a straightforward way while keeping the UI build stage cached. So I had to implement a workaround with a static placeholder on the build stage and its replacement on the final release preparation steps.
Also, I noticed that the `NODE_ENV` build arg is always different for build and test workflows, so the build cache (even with the same key) won't be used. That is why I renamed the build cache for the UI part and added a `cache-to` parameter for UI testing.
Finally, as we always know the NODE_ENV value when we build the image, I replaced the static `npm ci` command with a conditional one, which installs dev-dependencies only for non-production builds - this is much more secure, and a bit faster.

## Summary by CodeRabbit

* **Chores**
  * Improved build caching for UI components in continuous integration workflows, resulting in faster and more efficient builds.
  * Updated the Docker build process to inject the Git commit SHA into the UI at release time, enhancing traceability of deployed versions.
  * Optimized dependency installation during UI builds for better performance in production and development environments.
  * Enhanced build workflows to dynamically fetch branches and use commit SHAs, improving build consistency and image labeling.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -144,7 +144,7 @@ jobs:
           build-args: |
             NODE_ENV=production
             VERSION=${{ github.ref_name }}
-            GITHUB_SHA=${{ steps.git.outputs.short_sha }}
+            COMMIT_SHA=${{ steps.git.outputs.short_sha }}
 
   dockle:
     name: Dockle Container Analysis
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -117,18 +117,43 @@ jobs:
             ${{ github.event_name != 'schedule' && 'type=semver,pattern={{major}}' || '' }}
             ${{ github.event_name != 'schedule' && 'type=sha' || '' }}
 
+      - name: Prepare build arguments
+        id: args
+        shell: bash
+        run: |
+          set -euo pipefail
+          echo "short_sha=$(git rev-parse --short HEAD)" | tee -a $GITHUB_OUTPUT
+
+          token=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
+            -u "${{ env.DOCKER_USERNAME }}:${{ env.DOCKER_PASSWORD }}" \
+            "https://auth.docker.io/token?service=registry.docker.io&scope=repository:photoview/dependencies:pull" \
+            | jq -r .token)
+          echo "deps_tag=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
+            -H "Authorization: Bearer ${token}" \
+            "https://registry.hub.docker.com/v2/repositories/photoview/dependencies/tags?page_size=100" \
+            | jq -r '
+              .results
+              | map(select(.name | startswith("sha-")))
+              | sort_by(.tag_last_pushed)
+              | reverse
+              | .[0].name
+            ')" | tee -a $GITHUB_OUTPUT
+
       - name: Build and push
         uses: docker/build-push-action@v6
         with:
           context:     .
+          target:      release
           sbom:        true
           provenance:  mode=max
           platforms:   ${{ env.PLATFORMS }}
           pull:        true
           push:        ${{ env.IS_PUSHING_IMAGES }}
           tags:        ${{ steps.docker_meta.outputs.tags }}
-          labels:      ${{ steps.docker_meta.outputs.labels }}
           annotations: ${{ steps.docker_meta.outputs.annotations }}
+          labels: |
+            ${{ steps.docker_meta.outputs.labels }}
+            photoview.dependencies.tag: ${{ steps.args.outputs.deps_tag }}
           cache-from: |
             type=gha,scope=ui-${{ hashFiles('ui/package-lock.json') }}
             type=gha,scope=test-api-${{ hashFiles('api/go.sum', 'scripts/install_*.sh', 'dependencies/*') }}
@@ -138,7 +163,8 @@ jobs:
           build-args: |
             NODE_ENV=production
             VERSION=${{ github.ref_name }}
-            GITHUB_SHA=${{ github.sha }}
+            GITHUB_SHA=${{ steps.args.outputs.short_sha }}
+            DEPS_TAG=${{ steps.args.outputs.deps_tag }}
 
   dockle:
     name: Dockle Container Analysis
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,7 +85,7 @@ jobs:
         run: |
           set -euo pipefail
           git fetch --all
-          echo "short_sha=$(git rev-parse --short HEAD)" | tee -a $GITHUB_OUTPUT
+          echo "COMMIT_SHORT_SHA=$(git rev-parse --short HEAD)" | tee -a $GITHUB_OUTPUT
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
@@ -144,7 +144,7 @@ jobs:
           build-args: |
             NODE_ENV=production
             VERSION=${{ github.ref_name }}
-            COMMIT_SHA=${{ steps.git.outputs.short_sha }}
+            COMMIT_SHA=${{ steps.git.outputs.COMMIT_SHORT_SHA }}
 
   dockle:
     name: Dockle Container Analysis
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,13 +123,7 @@ jobs:
         run: |
           set -euo pipefail
           echo "short_sha=$(git rev-parse --short HEAD)" | tee -a $GITHUB_OUTPUT
-
-          token=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -21,15 +20,21 @@
...
 
 COPY ui/package.json ui/package-lock.json /app/ui/
-RUN npm ci
+RUN if [ "$NODE_ENV" = "production" ]; then \
+    echo "Installing production dependencies only..."; \
+    npm ci --omit=dev; \
+    else \
+    echo "Installing all dependencies..."; \
+    npm ci; \
+    fi
```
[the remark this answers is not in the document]
kkovaletp: No, it is not. As far as I see, the build runs fine for the production case

.github/workflows/build.yml
```diff
@@ -117,27 +117,54 @@ jobs:
...
+            -H "Authorization: Bearer ${token}" \
+            "https://registry.hub.docker.com/v2/repositories/photoview/dependencies/tags?page_size=100" \
+            | jq -r '
+              .results
+              | map(select(.name | startswith("sha-")))
+              | sort_by(.tag_last_pushed)
+              | reverse
+              | .[0].name
+            ')" | tee -a $GITHUB_OUTPUT
+
```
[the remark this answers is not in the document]
kkovaletp: The fallback to the `latest` tag is set in the Dockerfile in that COPY directive, when the variable is expanded

kkovaletp:
@googollee, please review this small fix and optimization change

Dockerfile
```diff
@@ -1,35 +1,40 @@
...
-
-COPY ui/ /app/ui
+ARG VERSION=unknown-branch
+ARG TARGETARCH
 # hadolint ignore=SC2155
-RUN export BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ'); \
+RUN export BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%S+00:00(UTC)'); \
     export REACT_APP_BUILD_DATE=${BUILD_DATE}; \
-    export COMMIT_SHA=${GITHUB_SHA:-$(git rev-parse --short HEAD || echo 000000)}; \
+    export COMMIT_SHA="-=<GitHub-CI-commit-sha-placeholder>=-"; \
```
googollee, line 35: What does `-=<GitHub-CI-commit-sha-placeholder>=-` mean? Will it expand to GitHub SHA?

Dockerfile
```diff
@@ -1,35 +1,40 @@
...
-
-WORKDIR /app/ui
+ARG UI_PUBLIC_URL=/
+ENV UI_PUBLIC_URL=${UI_PUBLIC_URL}
 
-COPY ui/package.json ui/package-lock.json /app/ui/
-RUN npm ci
-
-COPY ui/ /app/ui
+ARG VERSION=unknown-branch
```
googollee, line 30: Just to confirm, will `VERSION` expand to a version if the tag is available?

Dockerfile
```diff
@@ -122,6 +127,11 @@ RUN --mount=type=bind,from=api,source=/dependencies/,target=/dependencies/ \
 COPY api/data /app/data
 COPY --from=ui /app/ui/dist /app/ui
 COPY --from=api /app/api/photoview /app/photoview
+# This is a w/a for letting the UI build stage to be cached
+# and not rebuilt every new commit because of the build_arg value change.
+ARG COMMIT_SHA=?commit?
```
googollee: What does `?commit?` means here?

.github/workflows/build.yml
```diff
@@ -80,7 +80,12 @@ jobs:
           ref: ${{ matrix.tags.ref }}
 
       - name: Fetch branches
-        run: git fetch --all
+        id: git
+        shell: bash
+        run: |
+          set -euo pipefail
+          git fetch --all
+          echo "short_sha=$(git rev-parse --short HEAD)" | tee -a $GITHUB_OUTPUT
```
googollee: Suggest to use `COMMIT_SHA` directly.

Dockerfile
```diff
@@ -1,35 +1,40 @@
...
-ARG NODE_ENV
-ENV NODE_ENV=${NODE_ENV:-production}
+ARG NODE_ENV=production
+ENV NODE_ENV=${NODE_ENV}
+
+WORKDIR /app/ui
+
+COPY ui/package.json ui/package-lock.json /app/ui/
+RUN if [ "$NODE_ENV" = "production" ]; then \
+    echo "Installing production dependencies only..."; \
```
googollee: Two more spaces between `if` and `else`. Same as `else` and `fi.

Dockerfile
kkovaletp: Yep, that is a totally unique placeholder string to search for on the latest step in the `release` stage (lines 130-134) and replace it with the real commit SHA (if provided) or another nicer placeholder, set in the ARG as the default value.

Dockerfile
kkovaletp: Yes, this is the full syntax of the `ARG` directive: `ARG <external argument name>[=<default value if no argument is provided externally>]`.
So, if something is provided from the build command arguments for this parameter, it will be passed to the env var from here. If nothing is provided, this default version will be passed instead.

Dockerfile
kkovaletp: This is the nicer final default value for the commit SHA (if nothing is provided from the build command), shown at the end of the Settings page after the version. The UI label length for the commit SHA is limited to ~8 characters (didn't really count them), so when I set a more meaningful default value (like "unknown_commit"), it was cropped on UI to something like "unknown", which doesn't look so clear. So, the "?commit?" is the best default value within the limit, I was able to invent)

.github/workflows/build.yml
kkovaletp: Do you mean the `GITHUB_SHA` default env variable, same as the `${{ github.sha }}` expression?
I used it at the 1st iteration, but the issue is that it contains a full (long) commit SHA. Not the short version.

Well, as GUI crops it anyway to the 1st ~8 chars, it is not a big deal to pass a long commit SHA as a build parameter, but I thought that this is a cleaner way...
If you'd like to drop an extra Bash command and go with the long commit SHA, I'll change it. Let me know.

Dockerfile
kkovaletp: Are you talking about line indentation?
This is a pain point: I use VScode with Docker plugins, and since ~1 month ago, after a version update, it started forcing the line indentation changes based on its understanding during each file save operation. I tried to indent those lines with more spaces, but every time I save it, the IDE moves them back...
I can do that change in a text editor and commit, but then I'll have to remember to revert the IDE formatting for these lines every time I edit this file from VScode

.github/workflows/build.yml
googollee: No. I mean use `COMMIT_SHA` instead of `short_sha`. If you want to keep the name, then `COMMIT_SHORT_SHA` is better.

Dockerfile
googollee: Proper line indentation improves readability. So it's better to keep them with branches. I don't use VSCode, so I won't be able to give any suggestions here. Maybe disable auto-format for `Dockerfile`?

Dockerfile
googollee: Is there any reason not to use `COMMIT_SHA` directly here?

Dockerfile
googollee: Then I'll say `nocommit` is better as a default. `?commit?` is quite confusing to me, and it looks like a mistake.

Also, in Git, a short SHA is 7 characters. `?commit?` doesn't fit with it. I guess the length is not such a matter thing. We could use `no_commit` for clarity.

Dockerfile
kkovaletp: Yes, it is the main reason I started the refactoring of the build workflow with my previous PR:
The commit SHA is changed for every commit, and once it is injected into the Dockerfile as an ARG, it invalidates all the cached directives from the ARG directive till the end of the stage. All stages, dependent on the stage with the changed ARG, are rebuilt without using the cache as well. As a result, we had to spend ~8-9 mins for each single build run.

After I moved the commit SHA injection to the latest possible place in the Dockerfile after all time-consuming directives, cache invalidation doesn't impact the build time, as now the step with patching previously built Settings page and replacing the placeholder with the injected SHA takes just a few seconds.

Dockerfile
kkovaletp: The label size on the UI is hardcoded and not synced with the git short commit length. Here is what it looks like for "no_commit" and "NoCommit" placeholders:
SCR-20250707-stkd
SCR-20250707-ssnu

I chose the "NoCommit" if you don't mind)

.github/workflows/build.yml
kkovaletp: The `short_sha` is just a variable, used as a pipe between 2 build workflow steps to pass the SHA. Anyway, I've renamed this variable to `COMMIT_SHORT_SHA`

Dockerfile
kkovaletp: I found how to disable Dockerfile formatting. It looks like I needed extra motivation to spend 30 mins on searching for the solution)

PR #1242, opened 2025-07-09
Auto-orient image during processing

kkovaletp:
I've noticed an image orientation bug while making thumbnails:

SCR-20250709-pohz
SCR-20250709-podu

Which causes these artifacts after clicking on them:

SCR-20250709-poqq
SCR-20250709-ponb

So, I added an image auto-orientation step for both the Thumbnails and Encode functions. It works stably now, and all images are oriented correctly.

### DIFF
--- a/api/scanner/media_encoding/executable_worker/magickwand.go
+++ b/api/scanner/media_encoding/executable_worker/magickwand.go
@@ -44,6 +44,10 @@ func (cli *MagickWand) EncodeJpeg(inputPath string, outputPath string, jpegQuali
 		return fmt.Errorf("ImagickWand read %q error: %w", inputPath, err)
 	}
 
+	if err := wand.AutoOrientImage(); err != nil {
+		return fmt.Errorf("ImagickWand auto-orient %q error: %w", inputPath, err)
+	}
+
 	if err := wand.SetFormat("JPEG"); err != nil {
 		return fmt.Errorf("ImagickWand set JPEG format for %q error: %w", inputPath, err)
 	}
@@ -73,6 +77,10 @@ func (cli *MagickWand) GenerateThumbnail(inputPath string, outputPath string, wi
 
 	if err := wand.ThumbnailImage(width, height); err != nil {
 		return fmt.Errorf("ImagickWand generate thumbnail for %q error: %w", inputPath, err)
+	}
+
+	if err := wand.AutoOrientImage(); err != nil {
+		return fmt.Errorf("ImagickWand auto-orient %q error: %w", inputPath, err)
 	}
 
 	if err := wand.SetFormat("JPEG"); err != nil {


api/scanner/media_encoding/executable_worker/magickwand.go
```diff
@@ -75,6 +79,10 @@ func (cli *MagickWand) GenerateThumbnail(inputPath string, outputPath string, wi
 		return fmt.Errorf("ImagickWand generate thumbnail for %q error: %w", inputPath, err)
 	}
 
+	if err := wand.AutoOrientImage(); err != nil {
+		return fmt.Errorf("ImagickWand auto-orient %q error: %w", inputPath, err)
+	}
+
```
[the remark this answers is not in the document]
kkovaletp: duplicate

api/scanner/media_encoding/executable_worker/magickwand.go
```diff
@@ -75,6 +79,10 @@ func (cli *MagickWand) GenerateThumbnail(inputPath string, outputPath string, wi
 		return fmt.Errorf("ImagickWand generate thumbnail for %q error: %w", inputPath, err)
 	}
 
+	if err := wand.AutoOrientImage(); err != nil {
+		return fmt.Errorf("ImagickWand auto-orient %q error: %w", inputPath, err)
+	}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, I tried this, but got thumbnails for rotated images with incorrect (opposite) dimensions, as the `wand.ThumbnailImage()` requires the dimensions to be provided explicitly, while I cannot predict or detect whether an image was rotated by the `wand.AutoOrientImage()` or not. Do you have some suggestions for this?
BTW, when you say "better results", what do you mean exactly?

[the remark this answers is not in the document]
kkovaletp: Thanks for your explanation and suggestions. As the impact on the resulting thumbnails is minimal and both suggested approaches add more new code, increasing the risk of having other bugs, I'd go with the current implementation.

googollee:
It's possible to add some of this type of media file in `api/scanner/test_media` and update `scanner_test` to load those media and verify the dimension.

kkovaletp:
@googollee, this is a good idea to add incorrectly rotated files to the test media folder and cover this case by scan test, but I see 2 problems:
1. I have the next 2 images, which both need to be added, as one is JPG and the other is RAW - they behave a bit differently. But the RAW file is pretty big: 36 MB. I'm not sure that it would be ok to store such a large file in the repo, so that each clone and checkout will download it. 
2. I tried to understand how to write this new verification in the `scanner_test` and got lost in that code.

Is it ok if I attach those 2 files here in a ZIP, so you can take them and update the test?

BTW, I found that the image dimensions were still incorrect in the info panel and fixed them for the Small (thumbnail) image. I'll ask the bot to find the right place to fix it for the Large image as well, which is generated for a RAW original image. I think that we shouldn't change the dimensions of the Original image on the panel.

api/scanner/media_encoding/encode_photo.go
kkovaletp, line 1: @coderabbitai, after my 1st commit in this PR was done, I did more testing and noticed that even while the images are now rotated correctly, as well as their thumbnails, the dimensions for large (high-res) and small (thumbnail) images on the info side panel were still incorrect.
That is why I prepared the fix for the thumbnail dimensions and committed them with my 2nd commit.
Please carefully analyze the code of this PR in the context of the product code, and suggest the best way to fix the incorrect dimensions issue for the high-res image for those images, auto-rotated during the scanning.

kkovaletp: @coderabbitai, sorry, but I'm confused: I traced the code and see the following chain:
1. `scanner/scanner_tasks/processing_tasks/processing_functions.go:15`:
```
err := imageData.EncodeHighRes(imagePath)
...
photoDimensions, err := media_encoding.GetPhotoDimensions(imagePath)
```
2. `scanner/media_encoding/encode_photo.go:125`:
```
func (img *EncodeMediaData) EncodeHighRes(outputPath string) error {
...
imgPath := img.Media.Path
...
err := executable_worker.Magick.EncodeJpeg(imgPath, outputPath, 70)
```
This means that the `EncodeHighRes()` gets the path for the new, potentially rotated image, and later the `GetPhotoDimensions()` is called with the same path to the just generated high-res image. So, if it was auto-rotated, it should return the correct dimensions, as it is called after the image was generated.
What do I miss?

kkovaletp: @coderabbitai, you suggest to apply a workaround, which breaks the expected purpose of the `IdentifyDimension()` instead of fixing the main issue: the `EncodeHighRes()` (or the `EncodeJpeg()`) function has to consistently update the EXIF metadata of the newly created high-res image instead of copying it from the original.
Please suggest the best consistent fix for this main issue.

kkovaletp: @coderabbitai, please reevaluate this issue, taking into account my latest commit to this PR.
Now the issue with the high-res image dimensions for rotated images is solved.
However, I noticed that the dimensions for the original image are mistakenly updated if it is a RAW image with no sidecar JPG image (such images show original, large, and small image dimensions on the info side panel). This is strange, as the original image dimensions are not updated for rotated images, generated from JPG sources (such images show only original and small image dimensions on the info side panel).
Please investigate the code and find the root cause of this strange and inconsistent behavior.

kkovaletp:
I fixed the dimensions for all sizes and cases.
Now the open question is only the test...

api/scanner/media_encoding/executable_worker/magickwand.go
```diff
@@ -71,8 +79,25 @@ func (cli *MagickWand) GenerateThumbnail(inputPath string, outputPath string, wi
...
+
+	if err := wand.AutoOrientImage(); err != nil {
+		return fmt.Errorf("ImagickWand auto-orient %q error: %w", inputPath, err)
+	}
+	// Reset EXIF orientation to 1 (top-left) since image is now properly oriented
+	if err := wand.SetImageOrientation(imagick.ORIENTATION_TOP_LEFT); err != nil {
+		return fmt.Errorf("ImagickWand set orientation for %q error: %w", inputPath, err)
+	}
+
+	if originalWidth != wand.GetImageWidth() && originalHeight != wand.GetImageHeight() {
```
googollee, line 94: Please swap `width` and `height` here if necessary, and keep calling `ThumbnailImage()` once.

api/scanner/media_encoding/executable_worker/magickwand.go
```diff
@@ -71,8 +79,25 @@ func (cli *MagickWand) GenerateThumbnail(inputPath string, outputPath string, wi
...
 
-	if err := wand.ThumbnailImage(width, height); err != nil {
-		return fmt.Errorf("ImagickWand generate thumbnail for %q error: %w", inputPath, err)
+	originalWidth := wand.GetImageWidth()
+	originalHeight := wand.GetImageHeight()
+
+	if err := wand.AutoOrientImage(); err != nil {
+		return fmt.Errorf("ImagickWand auto-orient %q error: %w", inputPath, err)
+	}
+	// Reset EXIF orientation to 1 (top-left) since image is now properly oriented
```
googollee, line 88: Could you move this step after swapping `width` and `height`?

kkovaletp: No, because I can detect the need for swapping the dimensions based on the auto-rotation result. Before the image is auto-rotated, I cannot know whether I need to swap the dimensions.

api/scanner/media_encoding/executable_worker/magickwand.go
kkovaletp: done

PR #1245, opened 2025-07-11
Fix codecov and add orientation tests.

googollee:
## Summary by CodeRabbit

* **New Features**
  * Added vertical orientation checks for specific media files in test suites, ensuring correct thumbnail dimensions for vertical images.
  * Introduced command-line flags for selectively running database and filesystem integration tests.

* **Bug Fixes**
  * Improved thread safety for EXIF parser initialization to prevent issues in concurrent scenarios.

* **Refactor**
  * Centralized and simplified integration test flag handling for more consistent test setup.
  * Updated logging throughout the application to use structured, context-aware logs for better clarity and consistency.
  * Reduced database setup log verbosity to show only warnings and errors.
  * Consolidated image loading and orientation logic to remove redundancy and simplify error handling.

* **Chores**
  * Streamlined test setup and flag initialization for improved maintainability.
  * Updated CI workflow to archive coverage results as artifacts and removed test results upload step.


api/scanner/media_encoding/executable_worker/magickwand.go
```diff
@@ -113,22 +84,39 @@ func (cli *MagickWand) GenerateThumbnail(inputPath string, outputPath string, wi
...
 
-func (cli *MagickWand) IdentifyDimension(inputPath string) (width, height uint, err error) {
-	if !cli.IsInstalled() {
-		err = fmt.Errorf("ImagickWand is not initialized")
+func (cli *MagickWand) IdentifyDimension(inputPath string) (width, height uint, reterr error) {
+	wand, err := cli.createWandFromFile(inputPath)
+	if err != nil {
+		reterr = err
 		return
 	}
```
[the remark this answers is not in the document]
googollee: The `original` Media URL means the original file which generates high resolution and thumbnail files. It' doesn't mean the original dimension that doesn't consider the oritention. Currently we don't have the requirement to get the original dimension of a photo. So no need to provide another function now.

.github/workflows/tests.yml
```diff
@@ -75,7 +75,12 @@ jobs:
           photoview/api \
           /app/scripts/test_all.sh
         docker cp test:/app/api/coverage.txt ./api/
-        docker cp test:/app/api/test-api-coverage-report.xml ./api/
+
+    - name: Archive code coverage results
+      uses: actions/upload-artifact@v4
+      with:
+        name: api-${{ matrix.database }}-code-coverage-report
+        path: ./api/coverage.txt
```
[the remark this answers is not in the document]
googollee: Fixed.

.github/workflows/tests.yml
```diff
@@ -75,22 +75,20 @@ jobs:
...
-    - name: Upload test execution results
-      uses: codecov/test-results-action@v1
-      if: ${{ steps.test.conclusion == 'success' || steps.test.conclusion == 'failure' }}
+    - name: Upload coverage
+      uses: codecov/codecov-action@v5
+      if: ${{ steps.test.conclusion == 'success' }}
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         flags: api-${{ matrix.database }}
-        files: ./api/test-api-coverage-report.xml
```
kkovaletp, line 92: @googollee, could you please explain this fix?
- What root cause of the inconsistent coverage results did you find after your investigation?
- Why do you remove the `junit XML report` and the corresponding test results upload step? We need it to keep reporting the data to the https://app.codecov.io/gh/photoview/photoview/tests/master functionality.
- Is the artifact added for manual analysis only, or is there an integration, reading it?
- How do you see that the issue is fixed now?

api/scanner/media_encoding/executable_worker/magickwand.go
```diff
@@ -68,32 +55,12 @@ func (cli *MagickWand) EncodeJpeg(inputPath string, outputPath string, jpegQuali
...
-	}
-	// Reset EXIF orientation to 1 (top-left) since image is now properly oriented
-	if err := wand.SetImageOrientation(imagick.ORIENTATION_TOP_LEFT); err != nil {
-		return fmt.Errorf("ImagickWand set orientation for %q error: %w", inputPath, err)
-	}
-
-	// If the original image is rotated by 90 degrees, swap width and height for thumbnail generation
-	if originalWidth != wand.GetImageWidth() && originalHeight != wand.GetImageHeight() {
-		width, height = height, width
-	}
```
kkovaletp, line 96: @googollee, why don't you need this dimensions swap in the case of a rotated image?
When I tested it, I had to swap them because they were calculated before the rotation, and so, are incorrect after the rotation > the thumbnail is squished without that. 
Did you manually test your new implementation?

kkovaletp: I guess, to simplify the code, you can call the thumbnail dimensions calculation from the `GenerateThumbnail()` and drop its `width, height` input parameters. Then the dimensions are calculated for the auto-rotated image, and there will be no need for a swap.

api/scanner/scanner_test.go
```diff
@@ -204,6 +211,27 @@ func TestFullScan(t *testing.T) {
...
+			if err := db.Preload("MediaURL").Where("title = ?", filename).Find(&media).Error; err != nil {
+				t.Fatalf("can't find media with name %q: %v", filename, err)
+			}
+
+			thumbnail, err := media.GetThumbnail()
+			if err != nil {
+				t.Fatalf("can't get thumbnail of media %q: %v", filename, err)
+			}
+
+			if thumbnail.Width >= thumbnail.Height {
```
kkovaletp, line 230: @googollee, please extend this test by:
- adding a horizontal (landscape) properly oriented image and checking that the auto-orientation didn't change its orientation (original, large, and thumbnail)
- checking that the 'original' image dimensions are not affected by the auto-orientation, and going to the DB right away, as they are from the original file
- checking that the `large` JPG image (if generated, for TIFF) dimensions are affected in the same way as the thumbnail ones

These verifications are important, as I saw defects in those cases while I was working on the fix

api/test_utils/flags/flags.go
kkovaletp, line 1: @googollee, please describe how the new flags implementation works?
I see, you just add it to import and don't use it later in tests. Also, you removed the previous code to continue on error instead of panic, but this one doesn't contain such a flag...

googollee: [They](https://github.com/photoview/photoview/blob/124b6d34a93298bd84cb3a317855bd2f55af6a7d/api/test_utils/integration_setup.go#L21) are not new flags. The problem of coverage is caused by providing [these flags](https://github.com/photoview/photoview/blob/124b6d34a93298bd84cb3a317855bd2f55af6a7d/api/test_utils/integration_setup.go#L21) when unit tests don't define/require them. `test_util` imports a lot of libraries. Some unit tests, as `executable_worker`, can't import it due to the looping dependency problem.

The new `flags` library doesn't have any other dependencies besides stdlibs. All libraries can import it for those flags if they don't need to call `TestMain()`.

api/scanner/media_encoding/executable_worker/magickwand.go
googollee: > why don't you need this dimensions swap in the case of a rotated image?

Now `IdentifyDimension()` returns the right dimension, so it's not necessary to check it again when generating thumbnails.

In photoview, we should always process with the correct dimensions. It's good to get the right one from beginning.

> you can call the thumbnail dimensions calculation from the GenerateThumbnail() and drop its width, height input parameters.

Maybe. I don't want to change the old process now. One step, one time.

.github/workflows/tests.yml
googollee: > What root cause of the inconsistent coverage results did you find after your investigation?

See the comment with `flags` library.

> Why do you remove the junit XML report and the corresponding test results upload step? We need it to keep reporting the data to the https://app.codecov.io/gh/photoview/photoview/tests/master functionality.

Since codecov can handle the coverage report of go, it's better to process go report directly.

I didn't know codecov test web before. I'll check if it supports go report directly. If not, I'll rollback to JUnit report.

> Is the artifact added for manual analysis only, or is there an integration, reading it?

It's useful for debugging the coverage problem.

> How do you see that the issue is fixed now?

Much better. See `api/scanner/scanner_tasks/processing_tasks/process_photo_task.go` coverage is up to 100% now, no more tests adding.

But `api/scanner/media_encoding/executable_worker/magickwand.go` coverage drops. This code should be covered by `scanner_test.go`. I'm still checking it.

kkovaletp: > Since codecov can handle the coverage report of go, it's better to process go report directly.

In that case, you can explicitly tell `codecov-action` to use the `./api/coverage.txt` and `test-results-action` - the `./api/test-api-coverage-report.xml` files. I hope that this disables the reports autodiscovery and won't allow the use of the XML report for reading the coverage.

I'll wait for your final investigation results, but maybe after the issue with the flags is fixed consistently, we won't need to do the modifications for both codecov steps...

api/scanner/media_encoding/executable_worker/magickwand.go
kkovaletp: Ok, sure. Just make sure that it works stably and covers all the cases.
Once you're done with all your investigation and testing, and push the final code to the PR, please share the Docker image, so I can test it as well.

googollee:
Close as this pr mixing orientation test and codecov fix. I'll do them separately.

PR #1246, opened 2025-07-11
Small Dockerfile improvements + a cleaner image rotation condition

kkovaletp:
These improvements were highlighted when I merged upstream to my forked repo:
- some small improvements for Dockerfile
- a cleaner and straightforward condition to detect image rotation to swap the target dimensions for the thumbnail

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,8 +45,6 @@ ARG TARGETPLATFORM
 # See for details: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
-WORKDIR /app/api
-
 ENV GOPATH="/go"
 ENV PATH="${GOPATH}/bin:${PATH}"
 ENV CGO_ENABLED=1
@@ -57,19 +55,16 @@ RUN chmod +x /app/scripts/*.sh \
     && source /app/scripts/set_compiler_env.sh
 
 COPY scripts/install_*.sh /app/scripts/
-# Split values in `/env`
-# hadolint ignore=SC2046
 RUN chmod +x /app/scripts/*.sh \
-    && export $(cat /env) \
+    && set -a && source /env && set +a \
     && /app/scripts/install_build_dependencies.sh \
     && /app/scripts/install_runtime_dependencies.sh
 
+# hadolint ignore=DL3022
 COPY --from=photoview/dependencies:latest /artifacts.tar.gz /dependencies/
-# Split values in `/env`
-# hadolint ignore=SC2046
-RUN export $(cat /env) \
+WORKDIR /dependencies
+RUN set -a && source /env && set +a \
     && git config --global --add safe.directory /app \
-    && cd /dependencies/ \
     && tar xfv artifacts.tar.gz \
     && cp -a include/* /usr/local/include/ \
     && cp -a pkgconfig/* ${PKG_CONFIG_PATH} \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -57,19 +55,16 @@
...
     && /app/scripts/install_build_dependencies.sh \
     && /app/scripts/install_runtime_dependencies.sh
 
+# hadolint ignore=DL3022
 COPY --from=photoview/dependencies:latest /artifacts.tar.gz /dependencies/
-# Split values in `/env`
-# hadolint ignore=SC2046
-RUN export $(cat /env) \
+WORKDIR /dependencies
+RUN set -a && source /env && set +a \
```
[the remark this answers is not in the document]
kkovaletp: fixed

api/scanner/media_encoding/executable_worker/magickwand.go
```diff
@@ -90,8 +90,8 @@ func (cli *MagickWand) GenerateThumbnail(inputPath string, outputPath string, wi
 		return fmt.Errorf("ImagickWand set orientation for %q error: %w", inputPath, err)
 	}
 
-	// If the original image is rotated by 90 degrees, swap width and height for thumbnail generation
-	if originalWidth != wand.GetImageWidth() && originalHeight != wand.GetImageHeight() {
+	// If the image was rotated by 90 degrees (dimensions swapped), adjust thumbnail dimensions
```
googollee: I prefer current way. When I read it, it clearly says `when "width != width, height != height" then swap them`. I need to focus it a while then notice the new way is `width == height, height == width`.

Dockerfile
```diff
@@ -45,8 +45,6 @@ ARG TARGETPLATFORM
 # See for details: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
-WORKDIR /app/api
```
googollee, line 48: It's better to have a default `WORKDIR` in the base image.

api/scanner/media_encoding/executable_worker/magickwand.go
kkovaletp: Yeah, the current way looks more natural; however, it might work incorrectly and hide potential errors:
Imagine, image dimensions somehow corrupted in a previous step, and now they don't match the original ones. With the current implementation, they will be swapped, but it is wrong. By setting this `if` verification, we want to detect the exact case of 90-degree image rotation, and to serve this purpose, the `width == height, height == width` condition is better. All other cases of dimensions mismatch should not go inside this `if`, as they don't mean the 90-degree rotation.

Dockerfile
kkovaletp: Yes, it is the final WORKDIR for this stage still. I've moved this line to line 78 - before the RUD directive, expecting this current dir set to the `/app/api`. Also, I set the `WORKDIR /dependencies` at line 65 to prepare the environment for the RUN directive in line 66, which expects the current dir to be `/dependencies`. This allows dropping corresponding `cd` shell commands from the RUN directives' chains, which is a bad practice in Dockerfile design, reported by Hadolint linter.

So, in short, the final stage's WORKDIR is still the `/app/api`, while now the recommended Dockerfile design is followed and 2 unnecessary `cd` commands removed.

api/scanner/media_encoding/executable_worker/magickwand.go
googollee: Please consider https://github.com/photoview/photoview/pull/1249. The better way of solving the rotation problem, is getting the correct width/height from the beginning. `photoview` is a photo gallary, not a photo processing library. It should correct the dimension with the exif tag in the whole process, but not only generating files.

kkovaletp: ok, sure. I've reverted the change in this file.

PR #1249, opened 2025-07-13
Update the orientation checking. Add orientation tests.

googollee:
## Summary by CodeRabbit

* **Refactor**
  * Simplified image loading and orientation logic for improved maintainability.

* **Tests**
  * Added tests to verify correct EXIF orientation extraction and thumbnail orientation handling.
  * Enhanced test suite to confirm thumbnail dimension orientation matches image metadata.

* **Chores**
  * Updated build workflow to always enable image pushing.

PR #1254, opened 2025-07-19
Set AutoOrient when reading files.

googollee:
Split from https://github.com/photoview/photoview/pull/1249. Test image: `ghcr.io/googollee/photoview:orient-fix`

## Summary by CodeRabbit

* **Refactor**
  * Streamlined image processing logic for JPEG encoding, thumbnail generation, and dimension identification, resulting in more consistent image orientation handling and improved maintainability.

* **Chores**
  * Updated workflow configuration to always enable Docker image pushing during builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,8 +12,8 @@ on:
     - cron: '18 1 * * 4'
 
 env:
-  # IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
-  IS_PUSHING_IMAGES: true # Switch on when pushing images with forked repos.
+  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
+  # IS_PUSHING_IMAGES: true # Switch on when pushing images with forked repos.
   DOCKER_REGISTRY: ${{ github.repository != 'photoview/photoview' && 'ghcr.io/' || '' }}
   DOCKER_USERNAME: viktorstrate
   DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}


kkovaletp:
@googollee, I've tested the new implementation, and mostly it works fine, but there is 1 bug in it: the "Original" image dimensions are also rotated, while the corresponding image is downloaded "as-is". As Photoview is the read-only app according to the source media library, we should show the "Original"'s metadata (including dimensions) "as is", even if it contains auto-recoverable mistakes, like an incorrect rotation.
It should work in the correct way in `master` after my PR is merged. I guess, we still need 2 different ways of getting image dimensions: 1 would use your new function with auto-rotation when needed, and the other one should just read them as they are in the source file.

googollee:
> As Photoview is the read-only app according to the source media library, we should show the "Original"'s metadata (including dimensions) "as is", even if it contains auto-recoverable mistakes, like an incorrect rotation.

I don't agree with that. Photoview should correct the data if any mistakes are found. For the dimension of the original file, the correct dimension is either the one after auto-rotation, or the dimension of how the file stores + orientation info. To keep it simple, I prefer to show the dimension after auto-rotation.

We could check if the rotation could be fixed after the next scan. If not, we should add some mechanism to fix it.

kkovaletp:
I see your point and I agree with that when we talk about the images, processed by us, and stored in the cache folder. But when we talk about the source media files, which are read-only to Photoview, we should show the metadata for them "as is". The reason is simple: when a user clicks on the "Original" row and downloads the image (the same one, as is stored in his original media library), it will be rotated incorrectly, and the dimensions shown by us for that image won't match what is really stored inside the file.
Well, I also agree that this is a tricky case, because if we refer to the standard photo viewer apps (like the "Preview" - part of MacOS), most of them would auto-correct the rotation and draw the image after applying the correction.

Maybe, the best way would be to implement some sort of a warning notification for such autorotated images next to their Original dimensions in the Sidebar, but it would require UI changes, as well as an integration (new API or new field in the existing API) to let UI know that it should show such a warning icon/tooltip... It makes sense to do as a separate PR, then.

What do you think?

googollee:
> when a user clicks on the "Original" row and downloads the image (the same one, as is stored in his original media library), it will be rotated incorrectly, and the dimensions shown by us for that image won't match what is really stored inside the file.

I don't get your point. The original file should be shown after auto-oriented, and the image size should be the size after auto-oriented. The file stores pixels in some way, and stores that way as EXIF data to identify how to read those pixels, which is orientation. When a user clicks on the "original" row and downloads the image, it will be shown as after auto-oriented, see how GitHub shows [the up arrow file](https://github.com/photoview/photoview/blob/master/api/scanner/test_media/orient/up_arrow_90cw_web.jpg).

It's not tricky. The photo viewer should refer to EXIF data to show an image. Preview is like this. Photoview should also be like this.

googollee:
Preview shows the file dimension, which is after auto-oriented:
image

Preview shows EXIF data and dimensions, which is not auto-oriented:
image

I'll say only these two ways are correct.

Since `photoview` is targeting simple usage and doesn't expose EXIF info on the UX, we should show the file dimension after auto-oriented, even with the "Original" file.

kkovaletp (approved):
Ok, agree. Approving the PR

kkovaletp:
Please don't forget to revert the build workflow change before merging ;-)

PR #1255, opened 2025-07-19
The new queue

googollee:
Fix [#1265](https://github.com/photoview/photoview/issues/1265)

Test image: `ghcr.io/googollee/photoview:new-queue-1208`.

## Summary by CodeRabbit

* **New Features**
  * Introduced a new, robust background job queue system for scanning albums, supporting concurrent processing and dynamic worker scaling.

* **Refactor**
  * Replaced previous scanner queue and periodic scanner components with a unified queue system for improved reliability and maintainability.
  * Updated GraphQL and server logic to use the new queue system.

* **Bug Fixes**
  * Enhanced error handling and logging for background scanning tasks.

* **Tests**
  * Added comprehensive tests for the new queue system.
  * Removed outdated tests related to the old scanner queue.

* **Chores**
  * Updated dependencies and improved test scripts with race condition detection.


scripts/test_api_coverage.sh
```diff
@@ -2,6 +2,7 @@
 set -euxo pipefail
 
 cd "$(dirname $0)/../api"
-go test ./... -v -database -filesystem -p 1 \
+go test ./... -v -p 1 -race \
```
googollee, line 5: @kkovaletp `-race` checks race condition errors when running tests, but it makes the test much slower, ~2x-3x slower. I want to keep it since race condition errors are hard to fix if we introduce any. It's better to find them out as early as possible.

kkovaletp: Sure, it makes sense.
But how can it find race conditions with this flag during normal test runs? Race conditions are hard to detect because they are not always reproducible, and to reproduce them, the code has to be executed in a way that simulates all the preconditions for the particular race condition. Is the GO test runner smart enough to find out all prerequisites for potential race conditions and simulate them at the test run? Or, does it just catch race conditions if they are reproduced, and report them in the test results in a better way than the default test failure output?

api/scanner/queue/common.go
```diff
@@ -0,0 +1,305 @@
...
+
+	var ticker *time.Ticker
+	if interval == 0 {
+		ticker = time.NewTicker(time.Second) // The interval does not matter since the ticker is stopped.
+		ticker.Stop()
+	} else {
+		ticker = time.NewTicker(interval)
+	}
+
+	ret := &commonQueue[Job]{
```
kkovaletp, line 57: @googollee, does the `ret` name stand for "return"?
I know that this is just a var name, but still, can you name it properly, based on what is inside?

api/scanner/queue/common.go
kkovaletp, line 1: @googollee, this is not a request for a change, but a topic to think about...
While I was reading the code of this module, I got a question/idea in my head:
Does it make sense to wrap some critical sequences of queue and data manipulations with transactions? This would ensure that if we get an error or panic somewhere in the middle of such a critical path, it will be rolled back consistently with no impact on the outside code or DB.
I know that the decision of implementing transactions or not depends on the code implementation flow and impact analysis of possible predicted and runtime errors on the app state and data consistency.
So, the question is: would we benefit from transactions in the queue implementation, and if yes, where does it make sense to add?

scripts/test_api_coverage.sh
googollee: It can't find out all prerequisites for potential race conditions.

Go injects codes with `-race` and prints logs if those code are accessed without proper barriers between goroutine. If a code doesn't run during tests, `go test` can't find anything. See [race detector](https://go.dev/doc/articles/race_detector) for details. Without `-race`, `go test` just checks results, but won't catch potential race conditions. That's why I improved `scanner_test` to cover most common cases.

api/scanner/queue/common.go
googollee: It stands for "return value". Since there is only one return (besides errors), it shows the final instance clearly. It's also used in some stdlib, like [LookupNetIP()](https://cs.opensource.google/go/go/+/master:src/net/lookup.go;drc=f77bba43aa223fc86fd223f3ea4ef60db8e0c583;l=267). But it's not a widely-adopted pattern.

Naming it as `queue` doesn't make sense, because this name doesn't have any context, just redeclares the type. Naming it as `returnValue` is too long. `ret` is a proper name here.

api/scanner/queue/common.go
googollee: It doesn't make sense to let a queue handle the transaction. It should be the job handler to think about if all tasks in a job should perform in a transaction. The queue is already complex, and should not expose too many details to the handler.

On the other side, not all tasks have effects on database, like generating thumbnails. A DB transaction can't delete thumbnails that are already generated. If we do, it requires that [ScannerTask](https://github.com/photoview/photoview/blob/0b157b330a29f0abac91a6bf98153cfe27480edf/api/scanner/scanner_task/scanner_task.go#L16) has a new method like `Rollback` or `Drop`. The whole workflow is another part of complexity, which should not tangle with the detail of a queue.

api/scanner/queue/common_test.go
```diff
@@ -0,0 +1,298 @@
...
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+)
+
+func init() {
+	// Avoid panic with providing flags in `test_utils/integration_setup.go`.
+	flag.CommandLine.Init("queue", flag.ContinueOnError)
+}
```
kkovaletp: @googollee, do we still need this function after the test execution logic was changed? I mean, that similar function was removed from other tests in the #1251.

kkovaletp:
> Fix #1208

@googollee, I guess, this PR doesn't address all the points in #1208, right?
It is fine if you'd like to split the issue fix into several PRs. But in this case, you shouldn't use closing keywords to link the issue with these PRs (except for the last one), because merging of such an intermediate PR would close the issue and we might lose it from our backlog... Now it is linked on the right-side panel in the Development section, and I don't see an option to unlink it. Please unlink it from your side.

Please comment on that issue and list the points from that issue, already been addressed by merged PRs or this PR.

api/scanner/queue/common_test.go
googollee: Good catch, fixed.

scripts/test_api_coverage.sh
kkovaletp: Ok, I see. 
Let's keep it in our tests for a while and see how it works.
I see that now API tests run ~11 min vs. ~6 min in master. This would be our longest job if Docker cache is used, but I'd say, it is still acceptable time. If we see them run 3x slower (18 min or longer), we'd need to selectively run tests with the `race` flag, as waiting for ~20 min per commit is not acceptable.

api/scanner/queue/common.go
kkovaletp: I didn't mean the DB or queue transactions in particular. Transactions might be implemented on a wide range of levels: 
- a full scan run as 1 big transaction
- a single job (of a particular type) as a transaction
- a specific GO function as a transaction
- a specific sequence of code lines within a function
- specific methods of integration interfaces (DB, filesystem, executable workers, etc.)
Instead, it was a more generic and strategic topic in the scope of the scanner queue to think and analyze from different perspectives and on all levels to uncover and discuss the potential benefits of wrapping some modules/code blocks with transactions.

I understand that the queue is already complex module, and adding transactions might make it even more complex and difficult to maintain. That is why it is better to analyze it first and find out whether the added complexity brings even bigger advantage for the particular cases.

googollee: No. A queue could not provide any transaction features. A queue is a manager of tasks, and all detail of a task should be avoided exposing to a manager. To implement the transaction, it requires knowing details of tasks, which a queue should not be aware of.

Another thing, `transaction` is a strong word. What does a `transaction` mean to file system? Since photoview creates multiple files during tasks (thumbnail, high resolution for raw files), there is no guarantee to make the process of an atomic transaction. If we hope to implement the `transaction`, we need to define what `transaction` means first.

kkovaletp: > What does a `transaction` mean

Sure, let's align on the term. For me, "transaction" is a way of implementing a sequence of operations, so that when it runs, all of them succeed and the result is applied consistently to the app state, or in the case that something went wrong, all the changes made within this sequence are reverted and the app is notified about the transaction failure, so the app state doesn't change (except for the notification, so the app can perform a corrective action if defined). Also, it should be possible to define which issues trigger transaction rollback (critical errors, panic, etc.) and which don't (warnings, non-critical errors, which we can accept or correct later and move on).

Here are some examples of transaction implementations according to the levels I mentioned before:
- a full scan run as 1 big transaction: if something triggers rollback, everything that was already done by the current scan run is reverted (all changes to the DB, filesystem). This would require having temporary stores for files, DB entries, and the app's internal variables (if there are any, updated by the scan flow and used by the app externally). The process would look like the scan flow writes everything to these temp stores, and if it is successful, the last step copies all the results from those temp stores to the corresponding permanent ones, clearing temp once after the successful copy.
- a single job (of a particular type): for example, we might wrap the process media task with a transaction. This would result in storing all the info about the processed media file in some cache + all related files created in a temp folder, and once everything is done successfully, we copy the files to our permanent cache folder and push cached metadata to the DB, clearing the temp stores on success.
- a specific GO function: for instance, the `RescaleWorkers()` function might be wrapped with a transaction, so that if we catch an exception while trying to start more workers (e.g., because of OS limits), we rollback previous changes and shut down already started workers during this function invocation. I know, it is not so beneficial, just an example)
- a specific sequence of code lines: for example, the lines 266-269 and 278-279 of the `common.go` file are wrapped with transactions, making sure that all queue modifications succeed or are not performed at all, so that we avoid a case, when `q.ongoing` and `q.backlog` are inconsistent because of incomplete execution of those lines.
- specific methods of integration interfaces (DB, filesystem, executable workers, etc.): this should be clearer: doing DB changes inside DB transactions; wrapping EXIF parser calls with a transaction, so that we get all EXIF info or nothing and avoid partial data, etc.

I hope this explanation clarifies what the `transaction` term means for me. Again, I don't insist on implementing any of these examples or anything else. They are just examples, unless you find them beneficial)

googollee: No. It doesn't explain what `transaction` you want. They are details of how to implement different kinds of transactions. As I said, all of these transactions are full of details of how a job (at different levels) runs. A queue should care about the details of how to maintain the inner status. They may be called `transactions`, but it doesn't have an exact meaning without a database. Invariant is a more common concept, and functions should maintain invariants before returning.

For other levels, a queue should not know the details of jobs.

Imagine you are a manager. What kind of coders do you want to work with?

1. A coder works like:
  - You give him a task
  - He does something and commits the code
  - You find the code is wrong
  - You roll the code back (and maybe need to learn how to roll back safely in Git)
  - Mark the task as failed
2. A coder works like:
  - You give him a task
  - He does something, finds it's wrong, rolls the code back, and tells you it fails
  - You mark the task as failed

googollee:
> Investigate which parts (tasks) of the job don't run in parallel and make them run in parallel if possible (for example, the initial folder scan task)

They are running in parallel now.

> Make sure that all the code uses the new Backend's logger instead of throwing the messages to the console directly or using other methods

Yes.

> Enhance the log messages with the worker context and processed item to make them useful and descriptive

Yes.

> Make sure that the task assignment to workers strategy is optimal, so that all workers are busy with similar loads and work effectively

Yes.

> Investigate and find the most optimal way to run image processing workers without spinning a new process for each media

Yes.

> Detect and properly fix all places in the code, potentially sensitive to race conditions (for instance, api/scanner/periodic_scanner/periodic_scanner.go, api/scanner/scanner_queue/queue.go)

Yes.

> Fix the null pointer dereference issue in line 49 of the api/scanner/scanner_tasks/processing_tasks/sidecar_task.go

It's not part of the queue. As there is no log of that part, I don't have any idea what the problem is.

> Cover the touched modules with good enough unit tests to make sure our future changes won't break anything.

I'll say currently, there are good enough unit tests. But there are always not enough for tests. We can add tests when necessary, like what we did with the orientation problem.

I'll say it's fixed most of the things, except the unclear problem of `sidercar_task`.

api/scanner/queue/common.go
kkovaletp: > It doesn't explain what `transaction` you want

Ah, I got your message the wrong way) I don't want a particular type or level of transaction. Instead, all this discussion was to discuss the concept in general and probably inspire you to implement it somewhere if needed.
The technical implementation could be any: `try - catch - finally` construction, `if` statements to verify the consistency, etc.
If you see no weak places and think that there is no need for such protective code, that is fine too.

kkovaletp:
> They are running in parallel now.

Hm, ok. I guess that you highlighted another misunderstanding because of my too generic description - sorry for that.
After reading the `findAllAlbumsJobs()` and `findUserAlbumsJobs()` in the `queue.go`, I see that it creates a job per user per existing album. So, yes, it runs in parallel from this perspective.

However, the initial folders discovery is still done in 1 thread inside the `scanner/scanner_user.go FindAlbumsForUser()` function. This means that if we run the 1st scan or add a lot of new folders in our source media location, all the workers would wait until 1 worker navigates through all those new folders and adds them as albums into our DB. Later, all the albums (including these new ones) would be picked up by the new `findUserAlbumsJobs()` function and scanned in parallel by all workers. Did I correctly understand the implementation?
Initial folder scan takes a noticeable amount of time for large media galleries (with many folders, nested in other folders). I've got complaints from users about this for the old queue, which is why I added this case as an example for this requirement.

If my understanding is correct, and the initial album detection runs in a single thread, are there some other similar cases that might fall out of your scope (face detection, blurhash generation, etc.)?

kkovaletp:
> > Fix the null pointer dereference issue in line 49 of the api/scanner/scanner_tasks/processing_tasks/sidecar_task.go
> 
> It's not part of the queue. As there is no log of that part, I don't have any idea what the problem is.

It was part of the static code analyzer report several months ago. The analyzer doesn't run for the GO code anymore, as its interpreter version is too old and doesn't support our GO version. There is no logged runtime error about this bug, as it was found without code execution.
If you want, I can ask our AI bot in a separate PR to verify that file and reevaluate the existence of null-pointer bugs in the code.

api/scanner/queue/common.go
googollee: > The technical implementation could be any: try - catch - finally construction

If we do, this would be in the job handler, but not in the queue, for example, expanding [`ScannerTask`](https://github.com/photoview/photoview/blob/66fecbd5acf58ffb320f3c3328e07e60147a9f23/api/scanner/scanner_task/scanner_task.go#L16) and calling new methods in [`ScanAlbum`](https://github.com/photoview/photoview/blob/66fecbd5acf58ffb320f3c3328e07e60147a9f23/api/scanner/scanner_album.go#L87).

I don't want to overdesign any pattern without a real request. A database transaction is not simple`try-cache-fianlly`. It is designed to keep the atomic behavior even handling the machine is rebooted while a transaction is running. That's why I said `transaction` is a strong word. I don't think there is any requirement for such strong consistency for the application now.

googollee:
> It was part of the static code analyzer report several months ago.

It would be better to create a new issue with the report as a bug, but not mix it with a refactoring issue.

googollee:
> This means that if we run the 1st scan or add a lot of new folders in our source media location, all the workers would wait until 1 worker navigates through all those new folders and adds them as albums into our DB.

Not really. Scanning the directory does not happen in a worker. It blocks the caller.

> Initial folder scan takes a noticeable amount of time for large media galleries

Are they complaining about the slow time of the whole first scan? Or are they complaining about the first photo showing slowly after the first scan starts?

> Initial folder scan takes a noticeable amount of time for large media galleries.

I can't say that it could benefit from concurrency. There is only one IO device, and no matter how many threads/goroutines there are, they are blocked by the one device.

kkovaletp:
> Are they complaining about the slow time of the whole first scan? Or are they complaining about the first photo showing slowly after the first scan starts?

About long scan time in general, and about the long time to wait for the initial folder discovery, where there are no noticeable signs that the job actually does something. Several users wrote that they even thought that it got stuck or didn't actually run, so they tried to start it once more manually or restart the Photoview server to "fix" the hung job.

> There is only one IO device, and no matter how many threads/goroutines there are, they are blocked by the one device.

If the media collection is stored on a single hard drive, then yes, there will be a bottleneck for concurrent goroutines. But I'd say that large collections are stored on multi-drive NAS in most cases (I'd say, almost all cases), and NAS is designed to serve at least hundreds of concurrent requests, so there will be a noticeable difference in speed in the case of parallel runs. Also, Linux filesystem is designed with caching capabilities, so even with the bottleneck on the IO device level, there still might be a benefit from running it in 2-3 parallel threads.

googollee:
> About long scan time in general

It's really hard.

> about the long time to wait for the initial folder discovery

It's possible to improve.

> ...  are stored on multi-drive NAS in most cases

A raid can be traded as one device. If you see some performance tests, the RAID cluster speed increases compared to a single disk. Those tests are all running in one thread.

I agree that we should improve the time from scanning start to the first folder showing.

I created sub-issues for different targets. This pr is fixed  #1265. The problem of 1st scan is #1264. Could you create a new sub-issue for the null pointer dereference issue?

kkovaletp:
> It's really hard.

I know, and that is why I tried to decompose those complaints, where I saw in most cases that the most critical issue is with folder discovery, as it doesn't expose any results, doesn't show any progress, just looks like a black box for users. It is fine if it can finish in some minutes and switch to the following more traceable steps. But for large collections, it might take several hours with no output (1 real user's scenario).
The 2nd priority issue from those decompositions is rescanning previously scanned media just to make sure that nothing has changed, and we can go on. This would be even more difficult to improve, as we cannot use the FS subscription functionality. We can store the last update time for each source media, and then quickly check if it is the same, and all corresponding files exist in the cache. But this is just an idea, which should be tested in the scope of another PR. I mention it here just because we're talking about user complaints.

> A raid can be traded as one device. If you see some performance tests, the RAID cluster speed increases compared to a single disk. Those tests are all running in one thread.

Well, it is way not that simple:
- There is UNRAID with independent drives
- Only 1 RAID group can be abstracted to a single device, and only for specific IO types (uncached data reads and writes), while other types will show differences in speed. While most home NAS servers are usually configured with 1 RAID group (RAID 5 or 6), some could be configured in RAID 10 or other advanced RAID configurations, which contain more than 1 RAID group and support parallel IO natively.
- Many SOHO-level NAS servers support SSD cache, and all of them use RAM to cache folder structure for exposed exports/shares. This means that even with 1 RAID group, we can query those NAS filesystems for the folder structure in parallel with minimal dependency on the disks' IO.

> Those tests are all running in one thread.

That is because they measure reads and writes through an Ethernet connection. For such IO scenarios, it is the best use case to run the IO in a single thread and minimize the impact of potential bottlenecks on Ethernet, RAID type, etc.
Our case is different: folder structure is typically in RAM and the app runs typically on the same host (natively or in Docker container - doesn't matter for this case). So, we should see benefits from parallel calls.

> Could you create a new sub-issue for the null pointer dereference issue?

Sure, I'll review those issues and create the missing one later. I'm pretty busy at work these weeks... Can't wait for a vacation))

googollee:
Let's discuss if we should improve the directory scanning in another thread. No matter what, this PR would not contain that part.

kkovaletp:
I did check everything in this PR except for pulling the image and testing it on my testing instance.
I'll do that in several days - don't know when exactly...

googollee:
Sure, take your time.

kkovaletp:
@googollee, finally, I got to this task and tested the new implementation with your image:
Now I see that the status messages are reported to the GUI, but they overwrite each other, not respecting their types, which causes progress bar corruption, and there is no 100% completion.

Please see the attached video: 

https://github.com/user-attachments/assets/ce653e3f-7a7d-4c36-bf01-ebe3e60d1769

I really hope that you can implement a better integration with the GUI notifications:
- Differentiate single-task processing notifications from those reporting the overall queue progress. So there will be 2 different messages. 
- Single-task processing notifications would still frequently update their own notification and progressbar - that is expected, while there is another notification with the status of the global queue execution progress.
- Single-task processing notifications can be excluded from the GUI notification integration if the separation is not possible
- Queue progress notification should be updated with the current active stage of the workflow (discovering albums, scanning albums, discovering media, scanning media, recognizing faces, generating blurhashes) and show the end-to-end progress bar, which finally should come to 100%
- Once workflow is completed (all blurhashes are generated and no tasks are in the queue), it would be great to show the "Scan job finished" notification, which should update the previous one with the queue progress.

BTW, you can see on the video a couple of errors about the new arrow images: "Error detecting faces in image (/photos/main/orient/up_arrow.*): error read faces: jpeg_mem_loader: unsupported pixel size". I know that this is unrelated to the Queue PR, but can you please also look at them separately: 
- Are they valid? If the files contain pixels with incorrect size, then the error is fine.
- If not, why only the face recognition module reports it? 
- Can we put MagicWand in the middle to apply autocorrections before an image goes to the face recognition module?

googollee:
The notification is more complex than I thought. It'll take time to figure out how to fix it.

googollee:
Since we can't get agreement, withdraw this PR first.

kkovaletp:
> Since we can't get agreement, withdraw this PR first.

The PR was fine except for the inconsistent notifications behavior. I thought that you accepted the fund issue with the notifications and just needed some more time to find a better solution...

PR #1266, opened 2025-07-27
Add jpeg-xl example file.

googollee:
I see JPEG XL is another new format that some photographers are talking about. See https://jpegxl.info/resources/jpeg-xl-test-page.html for details of this format.

We already include libjxl when building ImageMagick. Add examples in the test.

## Summary by CodeRabbit

* **Tests**
  * Updated test coverage to include an additional media file format in the expected results.


kkovaletp (approved):
Sure, let's have it)

PR #1276, opened 2025-08-18
Add scrollbar to people sidebar section

tyler-hagen:
If there are more than a few people recognized in a picture it is currently not possible to see all the faces on a non-mobile platform.

Screenshot 2025-08-17 at 10 52 27 PM

This is how it looks on chrome on macos after moving `overflow-x-auto` to the `ul` element and removing the height and margin modifications:

Screenshot 2025-08-17 at 11 01 38 PM

I'm not sure why the adjustments to margin/height were there (maybe to increase scroll area? - note that scrolling did work on mobile browser by dragging but no scroll bar was shown). I also think the scrollbar width was set to none to not have it take up space on a smaller device, but I believe that now the general guidance is to let the platform hide the scrollbar if it chooses to do so. I'm a total front end/styling noob though so if I'm totally off base please let me know.

## Summary by CodeRabbit

* Style
  * People section in the media sidebar now uses a 3-column grid instead of horizontal scrolling.
  * Avatars and details are stacked vertically and centered for clearer scanning.
  * Removed extra spacer/scroll area for a cleaner, more compact layout.
  * Improved spacing, alignment, and menu placement for more consistent, at-a-glance browsing of people.

### DIFF
--- a/ui/src/components/sidebar/MediaSidebar/MediaSidebarPeople.tsx
+++ b/ui/src/components/sidebar/MediaSidebar/MediaSidebarPeople.tsx
@@ -200,7 +200,7 @@ const MediaSidebarPeople = ({ media }: MediaSidebarFacesProps) => {
   const { t } = useTranslation()
 
   const faceElms = (media.faces ?? []).map((face, i) => (
-    <MediaSidebarPerson key={face.id} face={face} menuFlipped={i == 0} />
+    <MediaSidebarPerson key={face.id} face={face} menuFlipped={i % 3 == 0} />
   ))
 
   if (faceElms.length == 0) return null


tyler-hagen:
I'm not sure adding a test for this presentation-based change makes sense - I'd just be checking that the `ul` element has the overflow property set or something which isn't much of a value add.

kkovaletp:
@tyler-hagen, thanks for reporting this bug and proposing a fix for it!
I confirm that currently, we have an issue with an invisible horizontal scrollbar in the People section of the image sidebar. There is a way to scroll it to the right and see other faces only if your mouse has a horizontal scrolling wheel or if there is a touchpad.

Adding a visible horizontal scrollbar would fix the issue from the functional PoV, but I think that the UX concept here is wrong: the sidebar is designed to contain a lot of info, and so, it is ready for vertical scrolling to let users see all the info available for the media. Because of this, I don't see a reason for keeping all recognized faces in 1 row with additional horizontal scrolling inside the People section... Do you?
So, I'd propose to fix it in another, simpler, and more consistent way: if a picture has more than 3 recognized faces, split them by 3 per line and grow the People section vertically as long as needed to show all the faces with no nested scrollbar. This would keep the sidebar logic consistent.

What do you think?

kkovaletp:
@tyler-hagen, any updates so far?
Just checking in the case, you missed my previous reply...

tyler-hagen:
I think 3 per row makes sense. Here's what it looks like now:  

Screenshot 2025-08-24 at 2 50 05 AM

I also noticed the images aren't quite centered over the text below them so I updated that.

kkovaletp:
@tyler-hagen, great, thanks for implementing the change! Now it looks much more consistent with the sidebar UX.
Could you please review the https://github.com/photoview/photoview/pull/1276#pullrequestreview-3149227344 bot's high-level comment and implement those suggestions that you find beneficial?
Finally, please make sure (or ask the bot to verify it) that there is no dead code on the page after the layout changes (e.g., unneeded horizontal scrollbar or horizontal scrolling functionality)

tyler-hagen:
@kkovaletp I've pulled in all the bot's suggestions that I think are useful, and there's no dead horizontal scrolling code

kkovaletp:
@tyler-hagen, great, thanks!
I see that the bot provided 3 new comments - please read them as well.
I know that it looks like a frustrating back-and-forth process, but it is still better than nothing (as we don't have a UI maintainer on the project)

tyler-hagen:
Yeah I looked through them and implemented one of the three. The other two have to do with behavior on "narrow sidebars", but I've noticed that if I view the sidebar on a small device or make my browser smaller at some point the sidebar will just take up the whole screen. So for in order for those cases to happen, you'd have to be viewing the page on a device with ~300 pixels for it to be an issue which is such a narrow (no pun intended) edge case I don't think I should spend effort on it right now

PR #1281, opened 2025-08-19
Bump PostgreSQL image to version 17; add the auto-migrate preparation service

kkovaletp:
I've tested this setup in my forked repo and already run my Photoview instance with this config on PostgreSQL 17.

## Summary by CodeRabbit

* **New Features**
  * Added an optional, commented PostgreSQL auto-upgrade service template for in-place upgrades.

* **Documentation**
  * Added “ATTENTION to PostgreSQL users” notice, PostgreSQL 17 guidance, registry/update notes, demo link, supported databases/platforms section, expanded setup/getting-started and macOS/local dev docs, password URL-encoding examples, startup/run commands, FFmpeg hardware-accel guidance, and contributing note.

* **Chores**
  * Updated example Postgres image to v17, added optional depends_on autoupgrade guidance, and annotated ports with host:container notation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,16 +19,19 @@ feel free to join the Discord server [https://discord.gg/jQ392948u9](https://dis
 
 # ATTENTION to Docker users !!!
 
-We migrated to the new Docker registry <https://hub.docker.com/r/photoview/photoview> and all new images for the `master` tag, as well as future releases are going to be published there instead of the previously used registry. Old images will be still accessible in the old registry <https://hub.docker.com/r/viktorstrate/photoview>, so if you want to use 1 of those old images, you need to revert back to the old registry.
+We migrated to the new Docker registry <https://hub.docker.com/r/photoview/photoview>, and all new images for the `master` tag as well as future releases will be published there instead of the previously used registry. Old images will still be accessible in the old registry <https://hub.docker.com/r/viktorstrate/photoview>, so if you want to use one of those older images, revert to the old registry.
 
 Please update your `docker-compose.yml` file to use the new registry for the `photoview` image, as shown in the corresponding example of the compose file: <https://github.com/photoview/photoview/tree/master/docker-compose%20example>
 
 # ATTENTION to PostgreSQL users !!!
 
-We switched to the PostgreSQL 17 in the `master` branch. The PostgreSQL 17 will be the default recommended version
-of the PostgreSQL for the future release. If you use Photoview from this branch, please make sure to update
-your `docker-compose.yml` according to the `./docker-compose example/docker-compose.example.yml`.
-Don't forget to backup your DB before running the upgrade.
+We switched to PostgreSQL 17 on the `master` branch. PostgreSQL 17 will be the default recommended version
+for the next release. If you follow `master`:
+
+- For Docker users: please update your `docker-compose.yml` to match [docker-compose.example.yml](./docker-compose%20example/docker-compose.example.yml).
+- For direct host installations and external shared PostgreSQL instance usage scenarios: please upgrade your PostgreSQL database manually to version 17.
+
+Don't forget to back up your database before upgrading.
 
 ## Demo site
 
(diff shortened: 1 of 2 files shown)


docker-compose example/docker-compose.example.yml
```diff
@@ -156,9 +156,26 @@ services:
       retries: 5
       start_period: 3m
 
-  ## Uncomment the `postgres` service if PHOTOVIEW_DATABASE_DRIVER is set to `postgres` in the .env
+  ## Uncomment both `postgres` services if PHOTOVIEW_DATABASE_DRIVER is set to `postgres` in the .env
```
googollee: When this autupgrade service is required?

kkovaletp: Every major version upgrade requires running the DB content migration procedure, which requires both current and future PostgreSQL versions to co-exist. It is a tricky requirement for Docker containers, which is why this image was created.
It is safe to run it as a preparation step before the PostgreSQL container starts, as if no migration is needed, it exits with 0 code in a second. But if it is needed, it does the job automatically with no user interaction, and the PostgreSQL container would wait till the migration is running.

kkovaletp: https://hub.docker.com/r/pgautoupgrade/pgautoupgrade - here are the details about the new image

googollee: Will it handle other old versions, like 14 or 15?

It's a one-shot upgrade. I don't like to consider such cases in a common launch script. If users choose PostgreSQL/MySQL, they should have the knowledge to maintain it themselves.

We could provide some link like [pgupgrade](https://www.postgresql.org/docs/current/pgupgrade.html) or [pgautoupgrade](https://github.com/pgautoupgrade/docker-pgautoupgrade) in the doc and scripts.

kkovaletp: > Will it handle other old versions, like 14 or 15?

Yes, it handles any version since 9.5 automatically.

> It's a one-shot upgrade. I don't like to consider such cases in a common launch script.

Yes, till the next major version is released. In that case, everybody would need to remember how to do the upgrade again. It is not so straightforward a task without this image.

> If users choose PostgreSQL/MySQL, they should have the knowledge to maintain it themselves.

Well, our users are not necessarily High-Tech enthusiasts. They might host Photoview just to serve their need. I don't see any negative impact from this migration image, but I see how it makes the migration process easier for users with less expertise in administration.

Could you please describe your PoV in detail: what negative aspects do you see in this solution?

googollee: All those scripts are maintained by us in the future. Since we don't test it during releasing, they may fail silently. 

I hope to draw a line about what kind of support we should provide, and what we should not. Do we drop the support of PostgreSQL 9?

I want to support only for direct dependencies, which are services of databases. The upgrade of database is not direct dependency. Maybe `autoupgrade` is suitable for common cases. Does `autoupgrade` cover all cases? For those indirect dependency, we can depend on the community.

kkovaletp: Thanks for the detailed description.

> All those scripts are maintained by us in the future. Since we don't test it during releasing, they may fail silently

Well, you're right. However, I'd say that in this case we should limit our scope of maintenance only to our Docker-Compose setup and migration from PostgreSQL 16 (which we set in our [compose example](https://github.com/photoview/photoview/blob/f79d379e38241c687c55e9f899e30661fd9f115a/docker-compose%20example/docker-compose.example.yml#L136) for the 2.4.0) to the current version 17. As this image is only part of the compose and we don't tell anything about the direct host deployments, that would be out of the scope and completely handled by users. The same is for older PostgreSQL versions.
The migration from 16 to 17 was tested by me in my testing (small) and production (not so large as some of our users have, but still bigger and more diverse). And, as my production runs on Postgres, I'm going to test all future version upgrades, which anyway will go through us 1st, because after the new major version is released, we need to set it in the compose for both images and push to `master`.

> I hope to draw a line about what kind of support we should provide, and what we should not. Do we drop the support of PostgreSQL 9?

Make sense. I'll add the Readme change to this PR once we agree on the scope and the way we want to ship the new Postgres version. I thought that we support only the latest major Postgres version + 1 previous major version (just in case someone missed the new version release (like me) or forgot to update the compose file after the new version is added to our repo). PostgreSQL releases are not so frequent, so limiting to the last 2 major versions looks enough to me...

> Does `autoupgrade` cover all cases?

As far as I see, it should cover all our Docker-based deployments, because we actually utilize very basic PostgreSQL functionality and our DB size is small (even for large source media libraries). I see only the exception if a user shares the same Postgres container between several different products, running several different DBs on the same Postgres engine. In this case, we cannot take responsibility, as this is not a recommended setup. But from a technical PoV, the autoupgrade image should upgrade our DB correctly. The question would be on the other services' DBs...

> For those indirect dependency, we can depend on the community.

Sure, we can, but in that case, I'd expect many users will stay on the old Postgres version just to avoid the complexity. Which would force us to keep the support for old Postgres versions (which shouldn't cause us many problems because of good backward compatibility on the Postgres side and our very basic functionality usage in the product), but still, this is something that we might want to avoid.

googollee: OK. Could you update the comment of `autoupgrader`? Add what it is, when it should be enabled, and say it's optional in the normal case?

Thank.

kkovaletp: Done, also added explanations about supported platforms and DBs, and fixed many warnings for Markdown composition in the Readme

README.md
```diff
@@ -15,17 +15,27 @@ You configure Photoview to look for photos and videos within a directory on your
...
+# ATTENTION to PostgreSQL users !!!
+
+We switched to PostgreSQL 17 on the `master` branch. PostgreSQL 17 will be the default recommended version
+for the next release (PostgreSQL 16 remains supported). If you follow `master`:
+
+- For Docker users: please update your `docker-compose.yml` to match [docker-compose.example.yml](./docker-compose%20example/docker-compose.example.yml).
+- For direct host installations or external/shared PostgreSQL instances: please upgrade your PostgreSQL server and database manually to version 17 following PostgreSQL guidance.
+
+Don't forget to back up your database before upgrading.
+
```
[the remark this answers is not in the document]
kkovaletp: I don't want to go so deep in the scope of this warning message. Target users would go to the compose file and find the detailed description of the auto-upgrader service. It should be enough for the version upgrade.

README.md
```diff
@@ -19,16 +19,19 @@ feel free to join the Discord server [https://discord.gg/jQ392948u9](https://dis
...
 
-We switched to the PostgreSQL 17 in the `master` branch. The PostgreSQL 17 will be the default recommended version
-of the PostgreSQL for the future release. If you use Photoview from this branch, please make sure to update
-your `docker-compose.yml` according to the `./docker-compose example/docker-compose.example.yml`.
-Don't forget to backup your DB before running the upgrade.
+We switched to PostgreSQL 17 on the `master` branch. PostgreSQL 17 will be the default recommended version
+for the next release. If you follow `master`:
+
+- For Docker users: please update your `docker-compose.yml` to match [docker-compose.example.yml](./docker-compose%20example/docker-compose.example.yml).
+- For direct host installations and external shared PostgreSQL instance usage scenarios: please upgrade your PostgreSQL database manually to version 17.
```
googollee: Add a link of https://www.postgresql.org/docs/current/upgrading.html.

README.md
```diff
@@ -195,7 +196,7 @@ Possible ways of securing a self-hosted service might be (but not limited to):
 1. Configure a **Firewall** on your local network's gateway and allow only the intended type of incoming traffic to pass.
 2. Use **VPN** to provide external access to local services.
 3. Setting up a **Reverse proxy** in front of the service and forwarding all the traffic through it, exposing HTTPS port with strong certificate and cipher suites to the Internet. This could be one of the next products or something else that you prefer:
-   - [Traefic Proxy](https://doc.traefik.io/traefik/)
+   - [Traefik Proxy](https://doc.traefik.io/traefik/)
    - [Caddy](https://caddyserver.com/docs/getting-started)
    - [NGinx Proxy Manager](https://nginxproxymanager.com/guide/)
```
googollee: Typo Nginx.

googollee (approved):
Thanks for updating README. Besides some small fixes, LGTM.

README.md
kkovaletp: Thanks, already fixed.

README.md
kkovaletp: Sure, added to both Readme and Compose.

PR #1293, opened 2025-08-27
Remove internal exif 

### LINKED ISSUE #1291: Investigate the need in keeping the internal EXIF parser as part of the product
After we decreased the list of supported platforms and significantly improved our toolset, do we still need the `api/scanner/exif/exif_parser_internal.go` in the product?
What are the real-world scenarios and use cases for it, which the external EXIF parser cannot cover/handle?

If there are no mandatory/exclusive scenarios for this code, I'd propose dropping it from the codebase, as it is less reliable and requires much more maintenance from our side.

googollee:
Fix #1291.

## Summary by CodeRabbit

* **Refactor**
  * Standardized EXIF metadata extraction on the external ExifTool parser. The application now requires ExifTool to be available at runtime; startup will fail if it isn’t.
* **Chores**
  * Removed an unused EXIF-related dependency.
* **Tests**
  * Updated tests to rely solely on the external ExifTool-based parser.


kkovaletp:
@googollee, looks good, but please review the bot's comments

googollee:
Just remove the library first. The usage of exittool should be updated in the next PR.

kkovaletp:
Ok, sure, just don't forget about these findings if some of them are valid (even if unrelated to the PR goal)

googollee:
Others will be traced by #1295

Issue #1295, opened 2025-08-27
Improve the code of calling external tools


googollee:
- [ ] Collect all external tools in one sub-package, `externaltools`.
- [ ] Provide a unique initialization and cleanup.
- [ ] Better comments about those external tools.

PR #1297, opened 2025-08-28
Move `exif` under `externaltools`.

googollee:
Part of #1295

- Move `exif` under `externaltools`.
- Add cleanup function.
- Add tests.
- Remove dataloader. It's much simple without this loader.

## Summary by CodeRabbit

* **Refactor**
  * Replaced legacy EXIF subsystem with a new external-tool based EXIF parser, added explicit init/cleanup and centralized EXIF saving during media scans.

* **Bug Fixes**
  * Improved validation and sanitization of GPS and numeric EXIF fields; safer behavior and clearer errors when EXIF tooling is unavailable.

* **Tests**
  * Added unit and integration tests covering parsing, GPS edge-cases, sanitization, initialization/cleanup, and parse-failure handling.

* **Documentation**
  * Added package documentation for external tools and initialization patterns.

* **New**
  * Added a structured parse-failure reporting utility.


googollee:
I don't want to change too much old code in one PR, but... Well, I'll fix them later.

api/scanner/externaltools/exif/exiftool.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	// GPS coordinates - longitude
+	rawLongitude, err := fileInfo.GetFloat("GPSLongitude")
+	if err == nil {
+		longitude = &rawLongitude
+	}
+
+	if latitude == nil || longitude == nil {
+		return 0, 0, fmt.Errorf("no valid GPS data")
```
kkovaletp: @googollee, 0, 0 - are real GPS coordinates. I think that it is incorrect to return them in case of error (even while this is an existing practice of some vendors https://en.wikipedia.org/wiki/Null_Island). I'd prefer to set them to nil, as before

api/server.go
```diff
@@ -61,7 +62,11 @@ func main() {
 		log.Panicf("Could not initialize periodic scanner: %s", err)
 	}
 
-	exif.InitializeEXIFParser()
+	cleanup, err := exif.Initialize(ctx)
```
kkovaletp: @googollee, isn't it better to name the cleanup var in a specific way to the content: `cleanupExif` or similar? Just to not mix it with other modules' cleanups now and in the future...

api/test_utils/integration_setup.go
```diff
@@ -46,6 +50,12 @@ func IntegrationTestRun(m *testing.M) {
 	faceModelsPath := PathFromAPIRoot("data", "models")
 	utils.ConfigureTestFaceRecognitionModelsPath(faceModelsPath)
 
+	cleanup, err := exif.Initialize(ctx)
```
kkovaletp: @googollee, the same naming question here as well

api/scanner/externaltools/exif/exiftool.go
googollee: Before it returns (nil, nil) as an error, because it doesn't have a way to say it's error or not. If there is an error, then it should return `error`.  For a valid GPS (0,0), the function should return with (0, 0, nil).

It's a common pattern when returning with `error`. Since there is an error, a caller should not use the value, even it's a valid one.

api/server.go
googollee: Fixed.

api/test_utils/integration_setup.go
googollee: Fixed.

api/scanner/externaltools/exif/exif_test.go
```diff
@@ -0,0 +1,25 @@
...
+	cleanup, err := Initialize(t.Context())
+	if err != nil {
+		t.Fatalf("Initialize() error: %v", err)
+	}
+	defer cleanup(t.Context())
+
+	if _, err := Parse("./test_data/bird.jpg"); err != nil {
+		t.Fatalf("Parse() returns an error: %v", err)
+	}
+}
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/externaltools/exif/exif.go
```diff
@@ -0,0 +1,35 @@
...
+		return nil, err
+	}
+
+	log.Info(ctx, "Found exiftool")
+
+	return func(ctx context.Context) {
+		if err := globalExifParser.Close(); err != nil {
+			log.Error(ctx, "Cleanup exiftool error:", err)
+		}
+	}, nil
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/externaltools/tools.go
```diff
@@ -0,0 +1,4 @@
+// package externaltools contains a group of tools outside go runtime.
+// They are provided by the runtime environment. Some of them must be initiazlied before using, and be cleaned up properly when finishing.
+// The dependency of packages under externaltools must be as less as possible, to avoid cycle dependencies.
+package externaltools
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/externaltools/exif/exif.go
```diff
@@ -0,0 +1,34 @@
...
+		return nil, err
+	}
+
+	log.Info(nil, "Found exiftool")
+
+	return func() {
+		if err := globalExifParser.Close(); err != nil {
+			log.Error(nil, "Cleanup exiftool error:", err)
+		}
+	}, nil
```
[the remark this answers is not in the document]
googollee: Not necessary.

api/test_utils/integration_setup.go
```diff
@@ -46,6 +47,12 @@ func IntegrationTestRun(m *testing.M) {
 	faceModelsPath := PathFromAPIRoot("data", "models")
 	utils.ConfigureTestFaceRecognitionModelsPath(faceModelsPath)
 
+	exifCleanup, err := exif.Initialize()
+	if err != nil {
+		log.Fatalf("init exif error: %v", err)
+	}
+	defer exifCleanup()
+
```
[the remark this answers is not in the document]
googollee: Fixed with Panicf.

api/scanner/externaltools/exif/exiftool.go
```diff
@@ -0,0 +1,198 @@
...
+			break
+		}
+
+		dateTime, err = time.Parse(layoutWithOffset, date)
+		if err == nil {
+			retEXIF.DateShot = &dateTime
+			foundExif = true
+			break
+		}
+	}
```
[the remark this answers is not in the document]
googollee: We don't want to store UTC time. The exif time should keep the original timezone.

Also exif time is not rfc3339. It should not use `z` for time zone.

api/scanner/scanner_tasks/exif_task.go
```diff
@@ -13,14 +16,48 @@ type ExifTask struct {
...
+	if err := tx.Model(&media).Association("Exif").Replace(exif); err != nil {
+		return fmt.Errorf("failed to save media exif to database: %w", err)
+	}
+
+	if exif.DateShot != nil && !exif.DateShot.Equal(media.DateShot) {
+		media.DateShot = *exif.DateShot
+		if err := tx.Save(media).Error; err != nil {
+			return fmt.Errorf("failed to update media date_shot: %w", err)
+		}
 	}
```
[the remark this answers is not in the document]
googollee: Not really, Exif should keep the original time zone.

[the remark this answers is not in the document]
googollee: Fixed.

api/server.go
```diff
@@ -53,6 +53,12 @@ func main() {
 		log.Panicf("Could not migrate database: %s\n", err)
 	}
 
+	exifCleanup, err := exif.Initialize()
+	if err != nil {
+		log.Panicf("Could not initialize exif parser: %s", err)
+	}
+	defer exifCleanup()
+
```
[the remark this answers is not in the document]
googollee: No. The graceful shutdown should only trigger shutting down the http server, but not cleanup other libraries. The cleanup of other libraries should after the server closed.

api/scanner/externaltools/exif/exiftool.go
kkovaletp: > Since there is an error, a caller should not use the value

I see 2 logical issues with this approach:
- The code "invents" a valid default value, while the purpose of the function is to read the value from the media and verify its validity. This means that the code implicitly performs an action that isn't expected of it.
- The code makes an assumption about the output processing on the callers' side and transfers responsibility of (not)using valid outputs to the callers.

On the other hand, sending `nil, nil, Error` looks much more consistent and clear. If the code is unable to read the valid data, it returns the commonly recognized set of invalid values together with an explanation of why they are invalid. No assumptions made: even if a caller doesn't look at the error, the returned data itself is invalid and unusable, so even if it is tried to be used, there will be no incorrect media placement on the map.

api/scanner/externaltools/exif/exiftool.go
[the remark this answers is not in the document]
kkovaletp: @googollee, actually, I checked this case in my production DB and double-checked the `date_shot` field values in my test instance's DB in `media` and `media_exif` tables, and I see all of them in the server's timezone (+0200 and +0300), while I have many media files from our community members' defect reports and being shot in many timezones.

This means that the current EXIF dates handling (which is unchanged from the `master` branch) has a critical inconsistency:
- If we respect the original timezone from the media, we should save the date-time with that timezone to the DB and display it in the UI, use for sorting, etc.
- If we store the date-time in the server's timezone, we should convert the date-time from the original timezone and not just store it as-is. Using the UTC timezone in this case doesn't look so wrong here...

api/scanner/externaltools/exif/exiftool.go
googollee: It's not a good pattern to consider the value when there is an error happened. It's the caller responsibility to check errors. If caller doesn't respect to the return (error is a part of return), there are many ways to misuse the results.

Returning pointers to simple values is also not a good pattern. It implies callers should validate the pointer before using values. It implies some valid results like (nil, 10.0, nil) or (10.0, nil, nil).

> The code "invents" a valid default value

Please respect all return, including the error.

> The code makes an assumption about the output processing on the callers' side and transfers responsibility of (not)using valid outputs to the callers.

Sure. It's the responsbility to check error if a function returns one.

kkovaletp:
> I don't want to change too much old code in one PR, but... Well, I'll fix them later.

@googollee, sure, please create corresponding issues to handle those future fixes and mention them in the corresponding threads, so those links are added to the issues, and later we can trace them to the points, they were detected

api/scanner/externaltools/exif/exiftool.go
kkovaletp: ok, accepting your points. Anyway, while the errors are being processed, no `0,0` values are going to be stored in the DB, so no fake data is being reported to users. This is fine for me.

kkovaletp (approved):
LGTM.
But I didn't review the `exiftool.go` and its test carefully, assuming that the code was copied from the old external parser with no or minimal changes - the way, GitHub shows the diff for these 2 files (old files removal and new files creation with 100% new code) doesn't let me to see the diff((

googollee:
> doesn't let me to see the diff

That thing also made me sad. I tried to get that diff by having `git mv` as [one commit](https://github.com/photoview/photoview/pull/1297/commits/8ac51428215525d9622bed813d2c4fb83fca0364), and it did show diff with [the code change commit](https://github.com/photoview/photoview/pull/1297/commits/d37ebf7a036cc5cc68c0cf61a4f3b299c316878f). But that diff was lost in the overall diff.

If you want, you can check [the code change commit](https://github.com/photoview/photoview/pull/1297/commits/d37ebf7a036cc5cc68c0cf61a4f3b299c316878f) for the diff. It's not the final version but not far away.

api/scanner/externaltools/exif/exiftool.go
```diff
@@ -51,182 +25,174 @@ func sanitizeEXIF(exif *models.MediaEXIF) {
...
-		log.Printf(
-			"Incorrect GPS data in the %s Exif metadata: %s, %s, (expected latitude '-90'..'90' / longitude '-180'..'180'). Ignoring GPS data.",
-			mediaPath, latStr, longStr)
-		return nil, nil
+	if math.Abs(*latitude) > 90 || math.Abs(*longitude) > 180 {
+		latStr := fmt.Sprintf("%f", *latitude)
+
+		longStr := fmt.Sprintf("%f", *longitude)
+
+		return 0, 0, fmt.Errorf("incorrect GPS data: latitude %s should be (-90, 90), longitude %s should be (-180, 180)", latStr, longStr)
```
kkovaletp: @googollee, after you excluded the `mediaPath`, how can a user understand for which file this error is reported?

api/scanner/externaltools/exif/exiftool.go
```diff
@@ -51,182 +25,174 @@ func sanitizeEXIF(exif *models.MediaEXIF) {
...
 
-	// Get description
-	description, err := fileInfo.GetString("ImageDescription")
-	if err == nil {
-		foundExif = true
-		newExif.Description = &description
-	}
+func (p *ExifParser) Close() error {
+	return p.exiftool.Close()
+}
```
kkovaletp: @googollee, why did you move these init/cleanup code from the beginning of the file to this position? As far as I know, it is not typical (but it doesn't impact the execution)

api/scanner/externaltools/exif/exiftool.go
```diff
@@ -51,182 +25,174 @@ func sanitizeEXIF(exif *models.MediaEXIF) {
...
-	expProgram, err := fileInfo.GetInt("ExposureProgram")
-	if err == nil {
-		foundExif = true
-		newExif.ExposureProgram = &expProgram
+		dateTime, err = time.Parse(layoutWithOffset, date)
+		if err == nil {
+			retEXIF.DateShot = &dateTime
+			foundExif = true
+			break
+		}
```
kkovaletp: @googollee, the 2nd block overrides the results of the 1st block... What is the point? It looks strange...

api/scanner/externaltools/exif/exiftool_test.go
kkovaletp, line 1: @googollee, depending on whether you plan to fix the timezone inconsistency issue in this PR or in a new PR, please add the test for the `dateShot` field and cover several common cases with/without the timezone set in EXIF metadata

api/scanner/externaltools/exif/exiftool.go
googollee: `mediaPath` is not part of `extractValidGPSData`. If callers want to add that in the error message, they can wrap this error like `fmt.Errorf("parse %q error: %w", mediaPath, err)` when returning or logging.

api/scanner/externaltools/exif/exiftool.go
googollee: I hope to group all functions related to `ExifParser` together.

api/scanner/externaltools/exif/exiftool.go
googollee: This part is [updated](https://github.com/photoview/photoview/pull/1297/files#diff-b43e270a6feb300944a2e4e26c83d67b106a1868930cd9e9ab6981b6203a9f7eR171). For this logic, it won't overwrite, as the first layout will break the loop of `createDateKey` and won't come to the 2nd layout.

api/scanner/externaltools/exif/exiftool.go
kkovaletp: I searched in the final code for such wrapping and didn't find it. Can you please ensure that the final error message clearly points to the file in which the error was caught?

googollee: Well, before I'm thinking if I should pop this warns on the web, but it's hard to link with send notification.

With the latest commit, it logs warns in `Parse()`.

api/scanner/externaltools/exif/exiftool_test.go
googollee: There is a `stripped.jpg`, which doesn't contain any exif. It loads `dateShot` with file modified date from the file system, which contains a timezone. You can verify that with exiftool against that file.

api/scanner/externaltools/exif/exif.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/photoview/photoview/api/graphql/models"
+	"github.com/photoview/photoview/api/log"
+)
+
+var globalExifParser *ExifParser
+var globalInit sync.Once
+
```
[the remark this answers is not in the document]
googollee: Not necessary.

api/scanner/externaltools/exif/exiftool_test.go
```diff
@@ -0,0 +1,259 @@
...
+	"fmt"
+	"math"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/barasher/go-exiftool"
+	"github.com/photoview/photoview/api/graphql/models"
+	"github.com/stretchr/testify/assert"
+)
```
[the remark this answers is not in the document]
googollee: Not necessary. Especially they are test files.

api/scanner/externaltools/exif/exiftool_test.go
```diff
@@ -0,0 +1,259 @@
...
+	for _, img := range images {
+		t.Run(fmt.Sprintf("%s", path.Base(img.path)), func(t *testing.T) {
+			exif, failures, err := parser.ParseExif(img.path)
+			if len(failures) != 0 {
+				t.Errorf("parse failures: %v", failures)
+			}
+
+			img.assert(t, exif, err)
+		})
+	}
```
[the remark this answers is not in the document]
googollee: Not necessary. We are Go 1.25 already.

api/scanner/externaltools/exif/exiftool.go
```diff
@@ -0,0 +1,212 @@
...
+func (p *ExifParser) ParseExif(mediaPath string) (*models.MediaEXIF, ParseFailures, error) {
+	fileInfos := p.exiftool.ExtractMetadata(mediaPath)
+	if l := len(fileInfos); l != 1 {
+		return nil, nil, fmt.Errorf("invalid file infos with %q, len(fileInfos) = %d", mediaPath, l)
+	}
+
+	fileInfo := fileInfos[0]
+	if err := fileInfo.Err; err != nil {
+		return nil, nil, err
+	}
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/externaltools/exif/exiftool.go
kkovaletp: Having it in the log is enough for now.
I feel that we would need to refactor the notifications architecture/interface, as you find it too limited and inconsistent...
After it is done (if my assumption is correct), we can review all the API/scanner errors and warnings, and decide which ones to show as notifications.

Resolving the thread.

api/scanner/externaltools/exif/exiftool_test.go
kkovaletp: I see.
I think that this is a fallback source for exiftool if nothing is stored in the EXIF metadata.

After our discussion about timezone inconsistency, I guess that you'd prefer to have a separate PR for it, as it is pretty complex. Right? 
Then feel free to resolve this thread.

googollee: I feel helpless with this diff. Let me update the timezone in another PR. Thanks.

PR #1305, opened 2025-09-06
Build development UI code when needed

kkovaletp:


kkovaletp:
@coderabbitai, review the PR and write the description

Dockerfile
```diff
@@ -9,12 +9,14 @@ ENV NODE_ENV=${NODE_ENV}
 WORKDIR /app/ui
 
 COPY ui/package.json ui/package-lock.json /app/ui/
-RUN if [ "$NODE_ENV" = "production" ]; then \
+# NPM 10.x is the latest supported version for Node.js 18.x
+RUN npm install --global npm@10 \
```
googollee, line 13: I'm surprised that they didn't upgrade the `node:18` image to `npm@10`. Do you know if there is a rule of `node` image providing `npm`?

Dockerfile
```diff
@@ -28,15 +30,15 @@ ARG UI_PUBLIC_URL=/
...
-    export COMMIT_SHA="-=<GitHub-CI-commit-sha-placeholder>=-"; \
-    export REACT_APP_BUILD_COMMIT_SHA=${COMMIT_SHA}; \
-    export VERSION="${VERSION}-${TARGETARCH}"; \
-    export REACT_APP_BUILD_VERSION=${VERSION}; \
-    npm run build -- --base="${UI_PUBLIC_URL}"
+RUN export REACT_APP_BUILD_DATE="$(date -u +'%Y-%m-%dT%H:%M:%S+00:00(UTC)')"; \
+    export REACT_APP_BUILD_COMMIT_SHA="-=<GitHub-CI-commit-sha-placeholder>=-"; \
+    export REACT_APP_BUILD_VERSION="${VERSION}-${TARGETARCH}"; \
+    export REACT_APP_API_ENDPOINT="${REACT_APP_API_ENDPOINT}"; \
+    npm run build"$( [ "$NODE_ENV" = "production" ] || echo ":dev" )" -- --base="${UI_PUBLIC_URL}"
```
googollee: How about `npm run build$( [ "$NODE_ENV" != "production" ] && echo ":dev" ) -- ...`? The condition matching the branch is more clear to me.

Also the quote is not necessary as it's already quoted by `$(...)`.

Dockerfile
kkovaletp: The reasons are:
- The Node 18 is End of Life, so the base image is updated last time 4 months ago and is not going to be updated anymore: https://hub.docker.com/layers/library/node/18/images/sha256-eb29363371ee2859fad6a3c5af88d4abc6ff7d399addb13b7de3c1f11bdee6b9
- There is a rule to ship the NPM version, which was officially packed with that Node version. So the image contains 10.8.2, while the latest NPM@10 is currently resolved to 10.9.3

I'm not sure if they will maintain the NPM version 10 and keep releasing minor/patch versions, but even if not, using the latest available version is better than staying with the default one

Dockerfile
kkovaletp: Sure, I've updated the condition.

I'd like to keep the quoting to make Hadolint happy) I've removed the nested quotes around the `:dev` and tested both branches of the condition locally

Dockerfile
googollee: Wow, it's already node 24. It's ok to use npm@10 now. We should find a way to use the latest node version.

kkovaletp: It is impossible for the upstream repo because of a huge number of other dependencies being incompatible with the newer Node versions.

I have updated a lot of them in my forked repo, but still stuck with a short list of key dependencies, requiring some code changes for the version upgrade. To do that, I need to fix some more bugs, and integrate better logging in the UI code to better catch failures and understand their causes during version upgrades...

I want to ship what I already have, as this would be a long story)

PR #1311, opened 2025-09-08
Remove the dependency on the serving UI from the API URI base path logic

kkovaletp:
## Summary by CodeRabbit

* **New Features**
  * Added configurable API endpoint via PHOTOVIEW_API_ENDPOINT (defaults to /api) for flexible deployment paths.

* **Chores**
  * Docker healthcheck now targets the configured API endpoint to support custom base paths.
  * GraphQL playground and API routing now resolve under the configured API base path; minor UI log message capitalization fixed.
  * API endpoint handling improved to better accept and normalize absolute or relative endpoint URLs with validation.

* **Tests**
  * Updated tests to expect API-prefixed media URLs.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -135,7 +135,7 @@ func main() {
 	}
 }
 
-func setupGracefulShutdown(http_server *http.Server) {
+func setupGracefulShutdown(svr *http.Server) {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
 
@@ -149,7 +149,7 @@ func setupGracefulShutdown(http_server *http.Server) {
 		periodic_scanner.ShutdownPeriodicScanner()
 		scanner_queue.CloseScannerQueue()
 
-		if err := http_server.Shutdown(ctx); err != nil {
+		if err := svr.Shutdown(ctx); err != nil {
 			log.Printf("Server shutdown error: %s", err)
 		} else {
 			log.Println("Shutdown complete")
--- a/api/utils/endpoints.go
+++ b/api/utils/endpoints.go
@@ -8,18 +8,12 @@ import (
 	"strings"
 )
 
-const apiPrefix = "/api"
+const defaultIP = "127.0.0.1"
+const defaultPort = "4001"
+const defaultAPIPrefix = "/api"
 
 func ApiListenUrl() *url.URL {
-	const defaultIP = "127.0.0.1"
-	const defaultPort = "4001"
-
-	// Reuse the normalization logic so absolute/relative values are both supported.
-	// If EnvAPIEndpoint contains a full URL, use its path component.
 	apiPath := ApiEndpointUrl().Path
-	if apiPath == "" {
-		apiPath = apiPrefix
-	}
 
 	var listenAddr string
 
@@ -54,7 +48,7 @@ func ApiListenUrl() *url.URL {
 func ApiEndpointUrl() *url.URL {
 	apiEndpointStr := EnvAPIEndpoint.GetValue()
 	if apiEndpointStr == "" {
-		apiEndpointStr = apiPrefix
+		apiEndpointStr = defaultAPIPrefix
 	}
 
 	apiEndpointURL, err := url.Parse(apiEndpointStr)
@@ -65,7 +59,7 @@ func ApiEndpointUrl() *url.URL {
 
 	// If absolute URL with empty path (e.g. "https://host"), default to /api for backward compatibility.
 	if apiEndpointURL.Scheme != "" && apiEndpointURL.Host != "" && apiEndpointURL.Path == "" {
-		apiEndpointURL.Path = apiPrefix
+		apiEndpointURL.Path = defaultAPIPrefix
 	}
 	// Ensure relative paths start with a leading slash.
 	if apiEndpointURL.Scheme == "" &&


kkovaletp:
@googollee, when you review the PR, please try to understand the original logic of setting up the dependency between serving UI and setting the default API URI: I didn't get the use case for it and found it too complex to manage, so I removed it and set the `/api` to be always the default URI.

Dockerfile
```diff
@@ -143,7 +144,7 @@ ENV PHOTOVIEW_MEDIA_CACHE=/home/photoview/media-cache
 EXPOSE ${PHOTOVIEW_LISTEN_PORT}
 
 HEALTHCHECK --interval=60s --timeout=10s --start-period=10s --retries=2 \
-    CMD curl --fail http://localhost:${PHOTOVIEW_LISTEN_PORT}/api/graphql \
+    CMD curl --fail http://localhost:${PHOTOVIEW_LISTEN_PORT}${PHOTOVIEW_API_ENDPOINT}/graphql \
         -X POST -H 'Content-Type: application/json' \
         --data-raw '{"operationName":"CheckInitialSetup","variables":{},"query":"query CheckInitialSetup { siteInfo { initialSetup }}"}' \
     || exit 1
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, while you report the valid case, you're wrong in saying `The server listen prefix is fixed to “/api”`: The `/api` is the default URI if nothing is explicitly provided by the user. If the user sets the "PHOTOVIEW_API_ENDPOINT" env variable in the compose file to something else, it will change the URI to the string from that variable.

Do you see a better way to keep the helthcheck working after a custom value is set for the "PHOTOVIEW_API_ENDPOINT" and still be valid for the default use case with no custom value provided?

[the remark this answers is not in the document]
kkovaletp: @coderabbitai, thanks for the explanation. Does it make sense to implement something similar to:
```
apiEndpointStr := EnvAPIEndpoint.GetValue()
	if apiEndpointStr == "" {
		apiEndpointStr = apiPrefix
	}
```
for the `ApiListenUrl()`?
It looks like this will align both cases and make the setup consistent, but please verify that it won't break the case of running the server behind a reverse proxy or other common hosting deployment specifics.

api/server.go
```diff
@@ -135,7 +135,7 @@ func main() {
 	}
 }
 
-func setupGracefulShutdown(server *http.Server) {
+func setupGracefulShutdown(http_server *http.Server) {
```
googollee: Since the type is `http.Server`, use `server` as the name is clear enough here. Please keep it short.

kkovaletp: I've renamed the var here because it masks the package name, which refers to the `package server ("github.com/photoview/photoview/api/server")`. The bot in my repo has raised an issue for this case.
Maybe you can suggest a better var name, different from the package name?

kkovaletp:
@googollee, it looks like the bot is satisfied now with the PR's code, so you can do your review.

api/server.go
googollee: Since it's a local variable, I prefer a sort name. `svr` or even `s` are good enough.

api/graphql/models/media_test.go
```diff
@@ -105,7 +105,7 @@ func TestMediaGetThumbnail(t *testing.T) {
 	video := models.Media{
 		Title: "video-test.mp4",
 		Path:  "path/test.mp4",
-		Type:  models.MediaTypePhoto,
+		Type:  models.MediaTypeVideo,
```
googollee, line 108: I'm a little bit curious about why it was a photo before and didn't fail the test.

api/utils/endpoints.go
```diff
@@ -1,57 +1,60 @@
...
 	"net/url"
-	"path"
 	"strconv"
+	"strings"
 )
 
+const apiPrefix = "/api"
+
 func ApiListenUrl() *url.URL {
+	const defaultIP = "127.0.0.1"
```
googollee: Could you group those const values together? Also `apiPrefix` is `defaultAPIPrefix` actually.

api/utils/endpoints.go
```diff
@@ -1,57 +1,60 @@
...
 
-	shouldServeUI := ShouldServeUI()
-
-	apiPrefix := "/"
-	if shouldServeUI {
-		apiPrefix = "/api"
+	// Reuse the normalization logic so absolute/relative values are both supported.
+	// If EnvAPIEndpoint contains a full URL, use its path component.
+	apiPath := ApiEndpointUrl().Path
+	if apiPath == "" {
```
googollee: I see this logic is repeated [below](https://github.com/photoview/photoview/pull/1311/files#diff-a1decf6e4146e7454952c2a5dabb498b8bed149be1c12a16cf2d7e062038b8d2R55). Since `ApiEndpointUrl()` is exposed, I suggest to remove this check and rely on `ApiEndpointUrl()` to handle the default value.

api/graphql/models/media_test.go
googollee: Ah, ok. It doesn't check type below. Good catch.

kkovaletp: I don't know) I was surprised as well...
Probably the test is a false positive, or it doesn't care about the type at all...

api/utils/endpoints.go
```diff
@@ -1,57 +1,54 @@
...
 	var listenAddr string
 
 	listenAddr = EnvListenIP.GetValue()
 	if listenAddr == "" {
-		listenAddr = "127.0.0.1"
+		listenAddr = defaultIP
+	}
+	// Validate listenAddr is a valid IP (IPv4 or IPv6)
+	if parsedIP := net.ParseIP(listenAddr); parsedIP == nil {
+		log.Panicf("%q must be a valid IPv4 or IPv6 address: %q", EnvListenIP.GetName(), listenAddr)
```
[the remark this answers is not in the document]
kkovaletp: This env var is named "PHOTOVIEW_LISTEN_IP". It is clear that the IP is expected. Stop suggesting to allow hostnames!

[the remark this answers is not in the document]
googollee: You can rename it as `listenIP` if you want.

googollee (approved):

Issue #1134, opened 2024-11-21
Migrate Photoview from Environment Variables configuration to a config file approach


kkovaletp:
**Is your feature request related to a problem? Please describe.**
Not a problem, just an inconvenience: we have many environmental variables and more to come. The current approach of `.env` file to manage them is not the most convenient way.

**Describe the solution you'd like**
It would be great to have a dedicated config file, formatted in YAML (preferably, as it is already used in `docker-compose`) or TOML, as it provides powerful features and visually clear and easily read structure.

**Additional context**
This topic was originally discussed in the https://github.com/photoview/photoview/pull/1130 PR.

mafrosis:
FWIW I prefer ENV vars as it's one less file to consider.

All config for photoview is current in the `environment` block of my `docker-compose.yaml`. This simplicity is desirable for someone who runs many services in docker.

kkovaletp:
Good point, @mafrosis. I'd agree that having a list of environment variables is easier for docker service management.
We would need to carefully review all the pros and cons before making the decision here...

PR #1315, opened 2025-09-11
HTTP server access log with optional debug and logging to a rotated file

kkovaletp:
## Summary by CodeRabbit

* **New Features**
  * Optional file-based access logging with rotation, standardized request lines, sensitive-query redaction, graceful startup/shutdown of logging, and built-in CORS handling.

* **Documentation**
  * Examples updated with commented logging options, host log-volume hints, and host log-directory setup notes.

* **Chores**
  * Release image now creates a default access-log directory and exposes environment variables to configure access logging and the UI artifact path; log-rotation dependency added.

### DIFF
--- a/api/server/logging.go
+++ b/api/server/logging.go
@@ -23,10 +23,14 @@ import (
 type contextKey string
 
 var (
+	logMutex         sync.RWMutex
 	logFile          io.WriteCloser
 	logWriter        io.Writer = os.Stdout
-	logMutex         sync.RWMutex
 	logGlobalContext context.Context
+	sensitiveKeys    = []string{
+		"access_token", "token", "auth", "authorization", "apikey", "api_key",
+		"password", "passwd", "secret", "signature", "session", "jwt", "code",
+	}
 )
 
 // InitializeLogging sets up the logging system with optional file output
@@ -174,11 +178,6 @@ func sanitizeURI(u *url.URL) string {
 	queryString := cloneUrl.Query()
 	if len(queryString) == 0 {
 		return cloneUrl.RequestURI()
-	}
-
-	sensitiveKeys := []string{
-		"access_token", "token", "auth", "authorization", "apikey", "api_key",
-		"password", "passwd", "secret", "signature", "session", "jwt", "code",
 	}
 
 	for name := range queryString {


api/server.go
```diff
@@ -71,6 +71,10 @@ func main() {
 		log.Panicf("Could not initialize face detector: %s\n", err)
 	}
 
+	if err := server.InitializeLogging(); err != nil {
+		log.Printf("Warning: Could not initialize access logging: %s", err)
+	}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, the server can work with no logger initialized, so it is not a critical feature. That is why I set it to the warning level. Please analyze the code and confirm that.

api/server/logging.go
```diff
@@ -2,95 +2,410 @@
...
+	if logLevel == "" {
+		logLevel = "info"
+	}
+
+	// Default to console output
+	logWriter = os.Stdout
+
+	// If log path is configured, open file and create multi-writer
+	if logPath := utils.AccessLogPath(); logPath != "" {
+		file, err := os.OpenFile(logPath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, as you can see in the code, sensitive data is masked in debug mode, and no request/response details are printed in normal mode. On the other hand, this log file is created inside the container and mapped to the host with the intention of allowing it to be read by others from the host. So this is done intentionally with no risks.

api/utils/environment_variables.go
```diff
@@ -65,6 +74,19 @@ func (v EnvironmentVariable) GetBool() bool {
...
+	value := os.Getenv(string(v))
+	if value == "" {
+		return 0
+	}
+	i, err := strconv.Atoi(value)
+	if err != nil {
+		return 0
+	}
+	return i
+}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, why do you think that the generic getter function should validate for negative values? they might be valid for some env vars... This is strange suggestion...

googollee:
@kkovaletp could you describe the case that you want to store logs in a file? Since we are using docker for development, exposing logs to stdout/stderr and using Docker to collect logs are good enough for me. It's hard to make a log right, including speed and retention. I tend not to do it.

api/utils/environment_variables.go
```diff
@@ -84,3 +106,27 @@
...
+	return EnvAccessLogPath.GetValue()
+}
+
+func AccessLogMaxSize() int64 {
+	// Default: 10MB in bytes
+	if size := EnvAccessLogMaxSize.GetInt(); size > 0 {
+		return int64(size) * 1024 * 1024 // Convert MB to bytes
+	}
+	return 10 * 1024 * 1024
+}
```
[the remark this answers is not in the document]
kkovaletp: It is up to the user, as there is no way to accidentally set it too large.

api/utils/environment_variables.go
```diff
@@ -84,3 +106,27 @@
...
+
+func AccessLogMaxFiles() int {
+	// Default: keep 5 rotated files
+	if files := EnvAccessLogMaxFiles.GetInt(); files > 0 {
+		return files
+	}
+	return 5
+}
```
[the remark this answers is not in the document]
kkovaletp: It is up to the user to control the number of files.

kkovaletp:
> could you describe the case that you want to store logs in a file?

Sure. There are several points from my PoV to implement it:
- The most important thing is to store the debug log when it is enabled. The reason is that STDOUT contains both server and access logs mixed, and with the debug enabled, it fills up so fast that it is difficult to efficiently analyze it. Of course, we can redirect it to a file on the Docker level, but it will still contain a mix of server + access logs. And this way is limited to the Docker use case, while the direct host deployment would deal with the syslog, which is also not so convenient to filter and analyze.
- The debug log file would be useful not only for development purposes, but also to investigate issues reported by users. That is why I implemented filtering out sensitive content and masking it with placeholders.
- A separated access log in a file is a good common practice for web servers, and most of them provide this ability enabled by default for historical and audit purposes.

> It's hard to make a log right, including speed and retention

I agree. I'll try to implement the log auto-rotation using an existing library with the hope that it already handles it in the most efficient way. I don't plan to expose the debug log enabling capability (the log level env var) to our examples, so that we'd limit this feature to us and for the specific issue investigations on the user side. I think that with the debug enabled, it should be acceptable to see some performance degradation and increased load. But as it is going to be an exceptional case and for a relatively short time, I don't see a big problem here. Logging the access log to a file in the Info level should not impact performance, as there is not so much data per event.

Do I miss something?

api/server/logging.go
```diff
@@ -2,95 +2,410 @@ package server
...
+		logWriter = io.MultiWriter(os.Stdout, file)
+		log.Info(context.Background(), "Access logging enabled to file", "logfile", logPath)
+	}
+
+	if logLevel == "debug" {
+		log.Warn(context.Background(), "Debug access logging enabled")
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, thanks for reporting this inconsistency. Yes, that part wasn't implemented yet.
Please carefully analyze the current logging implementation in the context of the project, and research the most optimal solution for the automatic access logfile rotation using the 3 user-provided config options (from the PHOTOVIEW_ACCESS_LOG_PATH, PHOTOVIEW_ACCESS_LOG_MAX_SIZE, PHOTOVIEW_ACCESS_LOG_MAX_FILES). The logfile rotation should work robustly, reliably, and efficiently from the server performance point of view, as the access logging would be 1 of the constantly running code modules on the server, and it shouldn't noticeably impact the server's performance or lead to overfilling the log folder because of delayed rotation too much (or cause any other runtime issues)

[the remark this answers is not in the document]
kkovaletp: @coderabbitai, you've posted only the last part of the reply, so I don't see your suggestion and the reply is useless.
Please send it again, making sure to include the complete content and format it in the yaml format as a workaround for your bug of losing part of the message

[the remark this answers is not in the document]
kkovaletp: @coderabbitai, now I see only the last several words of your answer, and it looks like you have completely ignored my requirement to format your complete answer in the YAML format!
Do your job and prepare the answer to my original question with all my requirements met!!!

api/server/logging.go
```diff
@@ -2,95 +2,434 @@ package server
...
-	return &statusResponseWriter{
+// Enhanced status response writer that captures headers
+type debugStatusResponseWriter struct {
+	http.ResponseWriter
+	status          int
+	hijacker        http.Hijacker
+	capturedHeaders http.Header
+	bodyBuffer      *bytes.Buffer
+	bodySize        int64
+}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, please carefully analyze the code and explain the real use cases, when these new functions would be needed (for both Debug and Info levels if there is a difference).
Then, please give me the exact, consistent, and robust code replacements to implement and integrate them into the existing code in the most efficient way.
Finally, before posting the response, make sure that it contains the complete content you want to reply to me, and format your response message in YAML format.

api/server/logging.go
```diff
@@ -2,97 +2,491 @@ package server
...
+					logBuf.WriteString(fmt.Sprintf("Body: %s\n", string(bodyBytes)))
+				} else {
+					logBuf.WriteString(fmt.Sprintf("Body: [binary content, %d bytes]\n", r.ContentLength))
+				}
+
+				// Reset buffer for the actual handler
+				buf.Reset()
+				io.Copy(&buf, tee)
+				r.Body = io.NopCloser(&buf)
+			}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, this is the 2nd time you have reported a serious issue for this code and suggested a fix. Please carefully analyze the current implementation and design, develop a solution, consistent with the other code of the project, reliable/robust, which fixes all the issues and doesn't introduce new ones, while doing the job it was originally designed to do.
Before submitting your reply, please make sure to include the complete content of your reply and format it in YAML format

api/server/logging.go
```diff
@@ -2,97 +2,491 @@
...
+	w.status = status
+	w.ResponseWriter.WriteHeader(status)
 }
 
-func (w *statusResponseWriter) WriteHeader(status int) {
+func (w *debugStatusResponseWriter) WriteHeader(status int) {
 	w.status = status
+	for k, v := range w.ResponseWriter.Header() {
+		w.capturedHeaders[k] = v
+	}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, please explain in detail why this is a bug, and show me a few examples of the incorrect headers processing, which might happen because of the current implementation.
Before submitting your reply, please make sure to include the complete content of your reply and format it in YAML format

googollee:
I still don't want to introduce this feature, since there are complex logic and we don't have time to maintain them. In this PR, there are nearly 1k new lines code. I believe there will be more if we go to this direction.

> The most important thing is to store the debug log when it is enabled.

For this, I'd like to add a flag to enable/disable access log. For debugging, we can disable access log while developing.

Another way is that we could mark different kind of log with a `kind` type. Then it's easy to grep those log while developing.

kkovaletp:
@googollee, ok, I see your point. Then this feature will go to my forked repo once I finish the development: I plan to review the code end-to-end to ensure consistency and effectiveness of the implementation, refactor it to split it between several files, and cover it with unit tests.
Since you're not interested in it I'll not ask you to participate in the development of this feature anymore, but please go quickly through the current implementation and tell me if you notice some particular things, which might be still useful even without the complete code merge: since I already spent time on it, if there are some useful things, I'll create a separate PR for them.

api/server/logging.go
```diff
@@ -2,97 +2,515 @@ package server
...
+			log.Error(context.Background(), "Failed to close log file", "error", err)
+		}
+		logWriter = os.Stdout
+		logFile = nil
+	}
+}
+
+// Thread-safe log writing
+func writeLog(format string, args ...interface{}) {
+	logMutex.Lock()
```
googollee: nits: The locking may slow the program if it outputs a lot of logs.

api/server/logging.go
```diff
@@ -2,97 +2,515 @@ package server
...
+					}
+				}
+			}
+
+			// Log request body (with size limit for safety)
+			if r.Body != nil && r.ContentLength > 0 {
+				var bodyBytes []byte
+				var readErr error
+
+				bodyBytes, readErr = io.ReadAll(io.LimitReader(r.Body, maxLogBodyBytes))
```
googollee: For a middleware, it's not a good idea to read all body and put it back. It consumes a lot of memory, and it may break some protocols silently.

api/server/logging.go
```diff
@@ -2,97 +2,515 @@ package server
...
+					}
+				}
+			} else if r.Body != nil && r.ContentLength == 0 {
+				logBuf.WriteString("Body: WARN [is not empty, but Content-Length: 0]\n")
+			}
+
+			logBuf.WriteString("===^ INCOMING REQUEST ^===\n")
+
+			// Use debug writer with capture capabilities
+			debugWriter := newDebugStatusResponseWriter(&w)
```
googollee: It's also not a good idea. Covering all http cases is hard.

I also don't agree the original `statusResponseWriter`. Since it's already existed, it's not a high priority to change.

api/server/logging.go
```diff
@@ -2,97 +2,515 @@ package server
...
-		requestText := fmt.Sprintf("%s%s", r.Host, r.URL.Path)
-		durationText := color.Sprintf("@c%s", elapsed)
+	if !looksTextual {
+		return !isBinaryData(body)
+	}
+	return looksTextual
+}
 
-		fmt.Printf("%s %s %s %s %s\n", date, statusText, requestText, durationText, userText)
+func isBinaryData(data []byte) bool {
```
googollee: It's better to use MIME Type to determine the content.

api/server/logging.go
```diff
@@ -2,97 +2,515 @@ package server
...
+	return writer
+}
+
+func (w *simpleStatusResponseWriter) WriteHeader(status int) {
 	w.status = status
 	w.ResponseWriter.WriteHeader(status)
 }
 
-func (w *statusResponseWriter) Write(b []byte) (int, error) {
+func (w *debugStatusResponseWriter) WriteHeader(status int) {
```
googollee: It's better to group methods with a type, but not interleaving them.

api/utils/environment_variables.go
```diff
@@ -17,6 +19,17 @@ const (
 	EnvFaceRecognitionModelsPath EnvironmentVariable = "PHOTOVIEW_FACE_RECOGNITION_MODELS_PATH"
 )
 
+// Logging
+const (
```
googollee, line 26: I'd like to introduce config file somehow.

api/utils/environment_variables.go
```diff
@@ -65,6 +78,19 @@ func (v EnvironmentVariable) GetBool() bool {
...
 
+// GetInt returns the environment variable as an integer (defaults to 0 if not defined or invalid)
+func (v EnvironmentVariable) GetInt() int {
+	value := os.Getenv(string(v))
+	if value == "" {
+		return 0
+	}
+	i, err := strconv.Atoi(value)
+	if err != nil {
+		return 0
```
googollee, line 91: Maybe, logging the error?

api/server/logging.go
```diff
@@ -2,97 +2,515 @@ package server
...
+	} {
+		if strings.Contains(n, pat) {
+			return true
+		}
+	}
+	return false
 }
 
-type statusResponseWriter struct {
+type simpleStatusResponseWriter struct {
```
googollee: If you want to record the `status`, just implement `WriteHeader()` and `Write` methods. All other hiding methods should delegate to embeding `http.ResponseWriter` directly.

Same as debug one.

api/server/logging.go
```diff
@@ -2,97 +2,515 @@ package server
...
+					logBuf.WriteString(fmt.Sprintf("Response Body: [binary content, %d bytes]\n", len(responseBody)))
+				}
+			} else if debugWriter.bodySize > maxLogBodyBytes {
+				logBuf.WriteString(fmt.Sprintf(
+					"Response Body: [large content, %d bytes - not logged]\n",
+					debugWriter.bodySize,
+				))
+			} else if debugWriter.bodySize > 0 {
+				logBuf.WriteString("Response Body: [content was written but not captured]\n")
+			}
```
googollee: This code is way too complicated. It's just a list of all possible scenarios without any thought.

kkovaletp:
@googollee, thanks for your comments. Now I see how bad this code is and why you think that it is not ready for merge.
I'll think about how to improve it and research if it is possible to use a fully-functional web server to serve our API, so that we can rely on it instead of implementing our own web server from scratch.

googollee:
I didn't mean the access log is a bad idea. I'm still thinking access log with method/path/status etc, but not body contents. It may have a special tag like `kind=access` for filtering, or a flag to enable/disable access logs.

Since photoview is not a common gateway, I don't want to colorize logs because first it's not critial, second it may not works when collecting log to other storage (like [Grafana Loki](https://grafana.com/oss/loki/)).

kkovaletp:
> access log with method/path/status etc, but not body contents

I had to develop the body logging for my debug purposes, as without reading the body, it wasn't possible to understand where the communication between my UI and the API breaks. So, the body interception is really a corner case of the access log feature, but there are cases where it is absolutely needed.

> I don't want to colorize logs because first it's not critial, second it may not works when collecting log to other storage (like [Grafana Loki](https://grafana.com/oss/loki/)).

Completely agree. It was already implemented, but I had many problems when I tried to clear those colors for logging to the file. It looks nice in the shell, but I'm not sure how useful it is in real life with log forwarding and Docker console to access them during development...

> research if it is possible to use a fully-functional web server to serve our API

I asked ChatGPT, and it looks like there is no possibility to use a reverse proxy to serve API without a minimal HTTP server (same as we have now), but its recommendation was to migrate to gRPC, which would remove all the HTTP overhead from our code. Also, I understand that this is not applicable to our current architecture with a single image to serve both UI and API. Am I correct?

googollee:
> but its recommendation was to migrate to gRPC, which would remove all the HTTP overhead from our code.

Yes, it will remove all HTTP overhead and introduce another giant part of gRPC overhead. I don't want to move to gRPC because HTTP works better with the internet than gRPC. HTTP GET is more friendly with CDN for catching, while gRPC uses POST for all kinds of requests.

I suppose that in the GraphQL layer, it should already have some feature to dump request/response. I suggest to check that direction.

api/server/logging.go
kkovaletp: @coderabbitai, please reevaluate the code. Since we use the `lumberjack` dependency for writing and managing the logfile, do we still need this mutex to be part of the `writeLog()`? As far as I understand, the `lumberjack` dependency has its own mutex inside. Do we need it for console logging? 
If we do need it still, please carefully analyze the code and search for all relevant info in the web to suggest a consistent and robust solution to optimize the code performance and mitigate the bottleneck, while ensuring its functional stability for all supported use cases.
Please make sure to include the complete content in your reply, and format it in YAML.

api/server/logging.go
kkovaletp: @coderabbitai, please reevaluate/analyze the code and search for all relevant info on the web to suggest a consistent and robust solution to log the request and response body in a way (or on a level), ensuring optimal memory consumption and not breaking the workflow. Is there a dump request/response on the GraphQL level? Are there any other ways?
Please make sure to include the complete content in your reply, and format it in YAML.

api/server/logging.go
kkovaletp: @coderabbitai, please reevaluate/analyze the code and search for all relevant info on the web to suggest a consistent and robust solution to detect the type of data based on MIME type.
Please make sure to include the complete content in your reply, and format it in YAML.

api/server/logging.go
kkovaletp: I've removed the mutex from the `writeLog()`

api/server/logging.go
kkovaletp: I've added the bot's suggested code and did additional refactoring on top of it.

api/server/logging.go
```diff
@@ -2,97 +2,482 @@ package server
...
+	"sync"
 	"time"
+	"unicode/utf8"
 
 	"github.com/photoview/photoview/api/graphql/auth"
-	"github.com/wsxiaoys/terminal/color"
+	"github.com/photoview/photoview/api/log"
+	"github.com/photoview/photoview/api/utils"
+	"gopkg.in/natefinch/lumberjack.v2"
 )
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, if some imports are missing, the Go code compilation would fail. But all the tests are passed, and the Docker image is built. Please double-check your suggestion, as it looks to be wrong.

api/utils/environment_variables.go
kkovaletp: Implemented the error logging

api/server/logging.go
kkovaletp: Can you please explain in detail what is not optimal or wrong with the current `statusResponseWriter()`?
I'll try to refactor both if I know what is wrong and what is expected.

api/server/logging.go
kkovaletp: I thought that keeping the corresponding standard and debug functions together is a good grouping.
But ok, I've moved all the debug functions after the standard ones.

api/utils/environment_variables.go
kkovaletp: Do you mean a config file as an alternative to the list of env vars?
Well, that might be useful for some cases, but this would be a different story. We have the #1134 to track this improvement.

api/server/logging.go
```diff
@@ -2,97 +2,443 @@ package server
...
+					// Restore body for downstream handlers
+					if r.ContentLength <= maxLogBodyBytes {
+						r.Body = io.NopCloser(bytes.NewReader(bodyBytes))
+					} else {
+						// Large body: restore read portion + remaining unread portion
+						// Calculate remaining bytes to read
+						remainingBytes := r.ContentLength - int64(len(bodyBytes))
+						remaining := io.LimitReader(r.Body, remainingBytes)
+						r.Body = io.NopCloser(io.MultiReader(bytes.NewReader(bodyBytes), remaining))
+					}
```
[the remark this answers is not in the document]
kkovaletp: Irrelevant, as that code is removed

kkovaletp:
@googollee, please take a look again at this PR: I decided to remove the debug logging from it, as it became too complex and impacted the product maintainability a lot.

I hope that now the PR looks much more solid and valuable.
Please let me know if you're interested in merging it, so I'll generate unit tests to cover the new code.

googollee:
I still don't want to introduce log rotation. Through years, the best practise of log rotation is using some daemon management, like systemd or dockerd.

kkovaletp:
Ok, then I'll close it and use it for my forked repo only

PR #1329, opened 2025-10-08
Efficient content encoding and secure request processing on HTTP server side

kkovaletp:


api/routes/spa.go
```diff
@@ -27,26 +76,205 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+	// Verify the file exists
+	if _, err := os.Stat(fullPath); os.IsNotExist(err) {
+		return false
+	}
+
+	// Parse Accept-Encoding header
+	acceptEncoding := r.Header.Get("Accept-Encoding")
+	if acceptEncoding == "" {
+		return false
+	}
```
kkovaletp: @coderabbitai, please enhance this parser and implement filtering out all encodings with `q=0` weight. Please make sure that the code stays consistent with the other code and does the job efficiently and robustly

kkovaletp: @coderabbitai, this solution is too complex, and it does a lot more than was asked.
I need a simple, robust, and consistent code just to detect and filter out encodings with the `q=0` and `q=0.0`, if there are any, from the `acceptEncoding` before it is used. If there are no `q` params, this code shouldn't do anything (which would be the main execution flow, as `q` params are not so often passed for encodings).

api/routes/spa.go
```diff
@@ -27,26 +76,205 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+			// Index file doesn't exist - this is a serious configuration error, not a regular 404
+			log.Error(r.Context(), "Error: index.html not found", "index.html path:", indexPath)
+			http.Error(w, "Application index file not found", http.StatusInternalServerError)
+			return
+		}
+
+		// Set cache headers for index.html (short cache, must revalidate)
+		w.Header().Set("Cache-Control", "public, max-age=3600, must-revalidate")
+		http.ServeFile(w, r, indexPath)
 		return
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, please carefully analyze the code of this module in the scope of the project and reevaluate this case. The index file should be served (with priority on pre-compressed type if supported by the client):
- `/` requested (starting the session and querying the host's root)
- The index file is requested explicitly
- A valid URL path is in the request (inside the static path, and with no `..`), but the requested file doesn't exist (meaning, this is the React routing request and should be served by the index file)

Do I miss something? If not, please make sure that all the cases are handled correctly and processed consistently. If there is an issue reported by you in this thread, please suggest a consistent and robust solution that wouldn't break any scenarios.

api/routes/spa.go
```diff
@@ -27,26 +76,245 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+		log.Error(
+			r.Context(),
+			"requested path is outside of static path",
+			"static path", absStaticPath,
+			"requested path", r.URL.Path,
+			"error", err,
+		)
+		http.Error(w, "Invalid request URI", http.StatusBadRequest)
+		return
+	}
```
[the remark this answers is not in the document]
kkovaletp: The app will never send such a path with the `..` pattern. Also, this is not an all-purpose HTTP server, so while the current implementation correctly protects from the path traversal hack, I'm fine with the side effect.

api/routes/spa.go
kkovaletp, line 1: @coderabbitai, please carefully analyze this file in the scope of the project, as well as the current testing framework and previously created unit tests for other PRs. Please cover this code with new unit tests according to the requirements defined for this project, and make sure that these tests provide good coverage, while being consistent, robust, and maintainable.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, the application's server runs on Linux exclusively. Please remember that. Please confirm that even if clients from Windows browsers call the app, it will work fine.

kkovaletp:
@googollee, please review this PR. It was inspired by my long-developing "Separate UI image" epic.
The main purpose of this PR is to remove unnecessary real-time compression from all server endpoints and leave it only for the GraphQL endpoint, where it actually makes sense. Our media files are already compressed according to the format, as well as album downloads. The static UI content is pre-compressed at the image build time and served according to the "Accept-Encoding" values from clients.

I also added some security enhancements, as was suggested by the bot, and I think that they are useful, but I hope that they won't make a false impression of making this HTTP server secure:
- They are too simple to cover all possible attack vectors
- Most of them are applied to the SPA handler only, and others are still not covered. And I don't want to duplicate them for all handlers, cause this would increase maintainability overhead, but won't make the app secure enough.
So, from the security point of view, the only reliable way to make the server more-or-less secure is to serve UI by a generic-purpose Web server and proxy API calls to this simple self-developed HTTP server from that Web server securely (which would be part of my "Separate UI image" epic eventually).

Saying that, please let me know if you'd like to keep these SPA security enhancements or drop them and focus on the efficient content encoding.

api/routes/spa.go
```diff
@@ -16,7 +20,52 @@ type SpaHandler struct {
...
+const (
+	ctxKeyFullPath contextKey = "fullPath"
+	ctxKeyRelPath  contextKey = "relPath"
+)
+
 func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
+	staticPathAbs, err := filepath.Abs(staticPath)
+	if err != nil {
+		log.Error(context.Background(), "static path is not valid", "static path", staticPath, "error", err)
+		return SpaHandler{}
```
googollee: Please return an error instead of logging it. The server should fail with any errors when initializing the handler.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
 // on the SPA handler. If a file is found, it will be served. If not, the
 // file located at the index path on the SPA handler will be served. This
 // is suitable behavior for serving an SPA (single page application).
+// Pre-compressed files (.br, .zst, .gz) are served if the client supports
+// them, otherwise the original file is served.
 func (h SpaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	// get the clean path to prevent directory traversal
-	servePath := path.Clean(r.URL.Path)
+	if h.staticPath == "" || h.indexPath == "" {
```
googollee: You can remove these checks if `NewSpaHandler()` returns an error.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+
+	// Special case: root path should serve index.html
+	if relPath == "" {
+		h.serveIndexHTML(w, r)
+		return
+	}
+
+	ctx := context.WithValue(r.Context(), ctxKeyFullPath, fullPath)
+	ctx = context.WithValue(ctx, ctxKeyRelPath, relPath)
+	r = r.WithContext(ctx)
```
googollee: Suggest to use `r = r.WithContext(log.WithAttrs(ctx, ctxKeyFullPath, fullPath, ctxKeyRelPath, relPath)`. If so, any logs with this context will print with `ctxKeyFullPath` and `ctxKeyRelPath`.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
 func (h SpaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	// get the clean path to prevent directory traversal
-	servePath := path.Clean(r.URL.Path)
+	if h.staticPath == "" || h.indexPath == "" {
+		log.Error(r.Context(), "staticPath or indexPath not configured on SpaHandler")
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+
+	absStaticPath, _ := filepath.Abs(h.staticPath)
```
googollee: This should be done in `NewSpaHandler()`.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+	// Try to serve pre-compressed file
+	if h.canServePrecompressed(w, r) {
+		return
+	}
+
+	// Fallback to original SPA handler logic
+	h.serveOriginal(w, r)
+}
+
+func (h SpaHandler) canServePrecompressed(w http.ResponseWriter, r *http.Request) bool {
```
googollee: This name is strange. It's not **can**, it's *serve** the request if possible. Maybe `tryServePrecompressed()`?

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+	fullPath := getFullPath(r.Context())
```
googollee: Why passing variables from context? Please pass them through function parameters.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+	// Don't compress already-compressed files
+	if isCompressedFormat(filepath.Ext(fullPath)) {
+		return false
+	}
+
+	// Try to serve pre-compressed variant
+	return h.servePrecompressedFile(w, r)
+}
+
+func (h SpaHandler) serveOriginal(w http.ResponseWriter, r *http.Request) {
```
googollee: Why passing variables from context? Please pass them through function parameters.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+
+	// Fallback to original SPA handler logic
+	h.serveOriginal(w, r)
+}
+
+func (h SpaHandler) canServePrecompressed(w http.ResponseWriter, r *http.Request) bool {
+	fullPath := getFullPath(r.Context())
+
+	// Check if the original file exists
+	if _, err := os.Stat(fullPath); os.IsNotExist(err) {
```
googollee: This check should be in `ServeHTTP()`.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+
+	// Check whether a file exists at the given path
+	_, err := os.Stat(fullPath)
 	if os.IsNotExist(err) {
-		// file does not exist, serve index.html
-		http.ServeFile(w, r, filepath.Join(h.staticPath, h.indexPath))
+		// File does not exist, serve index.html (SPA routing)
+		h.serveIndexHTML(w, r)
 		return
 	} else if err != nil {
```
googollee: No need `else` branch here.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+	h.setCacheHeaders(w, getRelPath(r.Context()))
+
+	// Use http.FileServer to serve the static file with no compression
 	http.FileServer(http.Dir(h.staticPath)).ServeHTTP(w, r)
 }
+
+// servePrecompressedFile attempts to serve a pre-compressed variant of the file
+// Returns true if a pre-compressed file was served, false otherwise
+func (h SpaHandler) servePrecompressedFile(w http.ResponseWriter, r *http.Request) bool {
+	fullPath := getFullPath(r.Context())
```
googollee: Why passing variables from context? Please pass them through function parameters.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+
+	// Verify the file exists
+	if _, err := os.Stat(fullPath); os.IsNotExist(err) {
```
googollee: Why does it again?

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+		name      string
+		extension string
+	}{
+		{"br", ".br"},
+		{"zstd", ".zst"},
+		{"gzip", ".gz"},
+	}
+
+	for _, enc := range encodings {
+		if strings.Contains(acceptEncoding, enc.name) {
```
googollee: To simple branches, please `if !strings.Contains() { continue }`.

api/routes/spa.go
```diff
@@ -27,26 +76,246 @@ func NewSpaHandler(staticPath string, indexPath string) SpaHandler {
...
+			precompressedPath := fullPath + enc.extension
+			if stat, err := os.Stat(precompressedPath); err == nil && !stat.IsDir() {
```
googollee: To simple branches, please `if stat, err := os.Stat(); err != nil || stat.IsDir() { continue }`

But, why could it be a directory if we provides all precompressed files?

api/routes/spa.go
kkovaletp: I did that initially, but since there is a chain of function calls from the `ServeHTTP()`, I had to pass the same vars through all the chain, and it didn't look good from the code structure PoV. Then, I created global vars, but got the critical bug report from the bot about impossible concurrency with those global vars, as they will be rewritten by each thread, and there will be a mess.

So, I've added them to the request's context, so they are still globally accessible within the same request, but don't impact any other requests. Is there a negative side of this implementation? I thought that this is a standard way in Go for such cases...

api/routes/spa.go
kkovaletp: Same as above. I'll do it in the same way as the one approved for the previous case (once we clarify and agree on it)

api/routes/spa.go
kkovaletp: Actually, I moved all checks to the `ServeHTTP()` and removed this intermediate function.

api/routes/spa.go
kkovaletp: As part of handling the https://github.com/photoview/photoview/pull/1329#discussion_r2416299476 suggestion, I moved everything to the `ServeHTTP()` and removed this intermediate function

api/routes/spa.go
kkovaletp: Same as the https://github.com/photoview/photoview/pull/1329#discussion_r2416295349 discussion. Once we agree on the solution, I'll apply it consistently

api/routes/spa.go
kkovaletp: Correct, it must exist in any case here. Removed this dead code

api/routes/spa.go
kkovaletp: > But, why could it be a directory if we provides all precompressed files?

Well, the code doesn't guarantee that explicitly, so I added this check for consistency. However, I agree that there is almost no chance of happening, because there should be no folders with the `fullPath + enc.extension` name:
- In the image, we control this, and unless a user manually modifies the content of the `/app/ui` folder, it is impossible
- In a direct installation scenario, the user is responsible for everything, but still, it would be strange to modify anything inside the pre-built UI folder.

So, let me know if you want me to drop this `IsDir()` verification or keep it.

api/routes/spa.go
kkovaletp: Done

api/routes/spa.go
kkovaletp: Done

api/routes/spa.go
kkovaletp: Done

api/routes/spa.go
kkovaletp: Done

api/routes/spa.go
kkovaletp: Done

api/routes/spa.go
kkovaletp: Done

api/routes/spa.go
```diff
@@ -16,37 +21,289 @@ type SpaHandler struct {
...
+		}
+
+		// Check if this encoding has q=0 or q=0.0
+		normalized := strings.ReplaceAll(part, " ", "")
+		if strings.HasSuffix(normalized, ";q=0") || strings.HasSuffix(normalized, ";q=0.0") {
+			continue
+		}
+
+		filtered = append(filtered, part)
+	}
```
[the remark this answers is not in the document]
kkovaletp: I don't see real-world cases for this. If somebody intentionally sends those `0.00(0)` values to test the app, I don't care about the result. Real browsers won't send such values for effective 0 value.

api/routes/spa.go
googollee: Sure, it's a chain, and you should pass those variables but not put parameters in the context. The context should not be used to pass parameters. It may be one of case below:

- There is a concept, and need to be wrapped as a structure/object.
- They belong to a concept, and should be attached to a structure/object.
- They are parameters, and should be passed down to all functions.

kkovaletp: @googollee, I've implemented the direct function parameter passing. Please review.

kkovaletp:
@googollee, I've implemented the requested changes some time ago. Is there a chance you can review them and potentially approve the PR, once you have a few minutes?

api/routes/spa.go
```diff
@@ -16,37 +20,258 @@ type SpaHandler struct {
...
+	if os.IsPermission(err) {
+		return SpaHandler{}, fmt.Errorf("no permission to access index path %s", indexPathAbs)
+	}
+	if err != nil {
+		return SpaHandler{}, fmt.Errorf("error accessing index path %s: %w", indexPathAbs, err)
+	}
+	if stat.IsDir() {
+		return SpaHandler{}, fmt.Errorf("index path %s is a directory, must be a file", indexPathAbs)
+	}
```
googollee, line 35: How about a new help function like `validDir(dir string) error` and use it to check `indexPath` and `staticPathAbs`? Then the main function could be cleaned as:

```go
func NewSpaHandler(staticPath string, indexPath string) (SpaHandler, error) {
  if err := validDir(staticPath); err != nil {
    return fmt.Errorf("....", staticPath, err)
  }
  if err := validDir(indexPath); err != nil {
    return fmt.Errorf("...", indexPath, err)
  }

  return SpaHandler{
		indexPath:  indexPath,
		staticPath: staticPathAbs,
	}, nil
}
```

BTW, I don't know why you change `staticPath` to `staticPathAbs` but not `indexPath`. I guess both should work properly. For the simplicity, I keep using original value here.

api/routes/spa.go
```diff
@@ -16,37 +20,258 @@ type SpaHandler struct {
...
 // is suitable behavior for serving an SPA (single page application).
+// Pre-compressed files (.br, .zst, .gz) are served if the client supports
+// them, otherwise the original file is served.
 func (h SpaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	// get the clean path to prevent directory traversal
-	servePath := path.Clean(r.URL.Path)
+	relPath := filepath.Clean(r.URL.Path)
+	relPath = strings.TrimPrefix(relPath, "/")
+	fullPath := filepath.Join(h.staticPath, relPath)
+
```
googollee, line 52: Suggest adding here:

```
r = r.WithContext(log.WithAttr(r.Context(), "status_path", h.staticPath, "requested_path", r.URL.Path)
```

Then when logging later, it could be simplified as:

```
log.Error(r.Context(), "error message", "error", err)
```

api/routes/spa.go
kkovaletp: > BTW, I don't know why you change `staticPath` to `staticPathAbs` but not `indexPath`. I guess both should work properly.

@googollee, this is because I need to pass the relative path to the `setCacheHeaders()`. Of course, it is possible to use `strings.Contains()` instead of `strings.HasPrefix()`, but in theory, it can introduce a bug if the searched substring is not on the right place of the Abs path (for instance, if user located the Photoview files on the host somewhere under his own `assets` folder). The current implementation is safer.

kkovaletp:
@googollee, finally, all suggestions addressed, and the code is ready for your final review

PR #1343, opened 2025-11-21
Switch the WatchTower image source to the maintainable fork

kkovaletp:
The WatchTower project has been unmaintained for years and recently became incompatible with the latest Docker Engine version.
But there is [the well-maintained fork](https://github.com/nicholas-fedor/watchtower), with many GitHub stars and regular PRs. So, here is the PR to switch to it.

## Summary by CodeRabbit

* **Chores**
  * Updated Docker Compose example to change the automatic-update service and add prominent warnings about potential silent issues from automatic version upgrades; users are advised to periodically review logs and monitor for unexpected behavior.

### DIFF
--- a/docker-compose example/docker-compose.example.yml
+++ b/docker-compose example/docker-compose.example.yml
@@ -102,6 +102,8 @@ services:
       # - "${HOST_PHOTOVIEW_MEDIA_FAMILY}:/photos/Family:ro"
 
   ## Watchtower upgrades services automatically (optional)
+  ## WARNING: the automatic version upgrades may silently generate unexpected problems,
+  ##   so please periodically check the logs
   watchtower:
     image: nickfedor/watchtower:latest
     hostname: watchtower


kkovaletp:
@googollee, please review this PR

googollee:
I don't think it's a best practice to use Docker image. Could we remove this from the example file?

kkovaletp:
> I don't think it's a best practice to use Docker image

What do you mean? Please give me some details, as I don't understand this sentence in the context of this PR.

googollee:
I don't think it's a good practice to update image automatically. It causes some unexpect problem like https://x.com/gwenshap/status/1990942970682749183. Even `photoview` is not used in a strictly production environment. 

I don't like recommending such solutions if users are unaware of them. If they are, they should know how to set it up.

kkovaletp:
ok, I see.

This is a big holy war in the DevOps and the Administration communities:
- One part says that we shouldn't automatically update versions before we read and test the new version properly
- Another part says that we have to update versions ASAP to apply bug fixes and close vulns

Each paradigm makes sense, but:
1st one requires A LOT of attention and engineering hours to cover endless upgrades of hundreds (and even thousands) of new versions released, as nobody uses just a few apps and services - in the real world, we have many of them, each of which might have many of its own dependencies, which are still managed by end users.
2nd one sometimes results in unexpected bugs and other issues, like the one you've shared.

So, in practice, if the default strategy is the 1st one, it quickly transforms into "never upgrade while it works". For us, it is bad because:
- Many users stay on the old versions, which we don't want to support (cause we have no time for that)
- Less feedback on our recent merges, so more bugs (which we might introduce and didn't catch before the merge) stay in the `master` for a long time, making it difficult to investigate and fix when finally found (like the date-time issue).
- If I remember correctly, we had just a few issues, caused by or related to WatchTower, so I don't see a big negative impact from it.

Finally, answering your question:
- The WatchTower is an optional service, which is documented and can be easily turned off by users.
- But if you still want it, I can comment it out in the compose example, or even remove it (while I still think that it gives us more benefits than problems)

Please let me know your final decision, and I'll follow, as this is the 2nd time we have had this discussion.

googollee:
Well, OK. I agree that it's more annoying that users don't upgrade the server. Please add a warning in the comment that the automatic upgrade may silently generate unexpected problems.

kkovaletp:
Done

PR #1299, opened 2025-08-30
Fix CR3 image format processing

kkovaletp:
I've added the CR3 file sample to the repo's test data and included it in the file formats test.
@googollee, please step in and fix the issue. The branch is in the main repo, so you don't need to clone any other repos.

PR #1345, opened 2025-11-21
Use `exiftool` to get MIME type, to support Canon CR3 RAW files.

### LINKED ISSUE #1307: CR3 RAW files produce `unsupported media type` error
**Describe the bug**

Raw files from my Cannon EOS R camera are marked as unsupported media type.

```
photoview | 2025/09/07 10:31:07 Unsupported media type "application/octet-stream" for file: /<filepath>/<image>.CR3
```

Previously we have used Darktable to develop RAW images to produce the high resolution JPEG.
This process was slow, and Darktable was a large dependency in terms of size.

However, most RAW files have a full sized JPEG embedded within.
We can potentially use `exiftool` (which we are already using for exif data) to extract these JPEGs.

```sh
$ exiftool -b -JpgFromRaw IMAGE.CR3 > IMAGE.JPG
```

When I run this command on my raw files, a full size jpeg version of the RAW image with the original nice cannon colors is produced.

**To Reproduce**
1. Download this [sample.zip](https://github.com/user-attachments/files/22194627/sample.zip) containing a .CR3 file from my camera
2. Add it to Photoview and scan
3. The above error is printed and the image does not appear in Photoview

**Expected behavior**

A jpeg version should be produced and the image should appear in the app.

**Your environment:**

 - Are you using Docker? yes
 - Database: mariadb:lts
 - Browser: Firefox
 - Version: photoview/photoview:sha-ce01f28

googollee:
Fix #1307

Test image: `ghcr.io/googollee/photoview:cr3-1307`

## Summary by CodeRabbit

* **New Features**
  * Added recognition/support for CR3 (Canon RAW) images.

* **Improvements**
  * Media-type detection now uses EXIF-based MIME lookups with clearer unknown-type handling.

* **Tests**
  * Added/updated MIME-type tests; several tests moved to integration mode and test harness adjusted for isolation.

* **Chores**
  * Removed libmagic integration and related install steps plus an unused dependency.
  * CI pipeline changed to always push Docker images.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,8 +12,8 @@ on:
     - cron: '18 1 * * 4'
 
 env:
-  # IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
-  IS_PUSHING_IMAGES: true # Switch on when pushing images with forked repos.
+  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
+  # IS_PUSHING_IMAGES: true # Switch on when pushing images with forked repos.
   DOCKER_REGISTRY: ${{ github.repository != 'photoview/photoview' && 'ghcr.io/' || '' }}
   DOCKER_USERNAME: viktorstrate
   DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}


api/scanner/externaltools/exif/exiftool.go
```diff
@@ -278,3 +287,17 @@ func calculateOffsetFromGPS(fileInfo *exiftool.FileMetadata, date *time.Time) (*
...
+		return "", fmt.Errorf("invalid parse %q exif: %w", mediaPath, err)
+	}
+
+	mime, err := fileInfo.GetString("MIMEType")
+	if err != nil {
+		fmt.Errorf("invalid parse %q mimetype: %w", mediaPath, err)
+	}
+
+	return mime, nil
+}
```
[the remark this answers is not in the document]
googollee: Fixed.

api/scanner/externaltools/exif/exif.go
```diff
@@ -60,3 +60,20 @@ func Parse(filepath string) (*models.MediaEXIF, error) {
...
+	if globalExifParser == nil {
+		return "", fmt.Errorf("no exif parser initialized")
+	}
+
+	mime, err := globalExifParser.ParseMIMEType(filepath)
+	if err != nil {
+		return "", err
+	}
+
+	return mime, nil
```
kkovaletp, line 77: @googollee, this is not an issue, but I don't understand the code construction here: why do we need to explicitly verify the `err` and conditionally return different combinations of results?
Doesn't the following code work exactly the same, but is much clearer?

`return globalExifParser.ParseMIMEType(filepath)`

As far as I understand, the function `ParseMIMEType()` always returns a string and an error - exactly as our expected return values, so we can just pipe it to the `return` directly. Then, it is the caller's responsibility to correctly process the result.

api/scanner/externaltools/exif/exif_test.go
```diff
@@ -43,3 +45,24 @@ func TestParse(t *testing.T) {
...
+	filename := "./test_data/bird.jpg"
+
+	mime, err := MIMEType(filename)
+	if err != nil {
+		t.Fatalf("MIMEType() returns an error: %v", err)
+	}
+
+	if mime == "" {
+		t.Errorf("MIMEType(%q) should not return an empty string", filename)
+	}
```
kkovaletp, line 67: @googollee, you've checked 2 negative cases. Please also check the positive: since you give the real image to the function, you can expect the exact MIME type to be returned.

kkovaletp: PS: ok, I found the exact MIME type checking in the `api/scanner/externaltools/exif/exiftool_test.go`, so there is no need for testing the same twice

api/scanner/externaltools/exif/test_data/IncorrectGPS.jpg
kkovaletp, line 1: @googollee, I see the file size change, but GitHub doesn't show the exact diff.
What was changed and why?
This particular image covers the case of GPS coordinates being out of the boundaries case. I'd like to make sure that this wasn't changed.

api/scanner/media_type/media_type.go
```diff
@@ -3,10 +3,25 @@ package media_type
...
+// This function is thread-safe.
+func GetMediaType(f string) MediaType {
+	mime, err := exif.MIMEType(f)
+	if err != nil {
+		log.Warn(nil, "GetMediaType() error.", "error", err, "file", f)
+		return TypeUnknown
+	}
+
+	return mediaType(mime)
+}
```
kkovaletp, line 23: @googollee, well, this approach will work for most of the media files, but as far as I understand, the `exiftool` just reads the MIME type from the EXIF metadata, so if it is absent or clear, nothing is returned, falling back to the `TypeUnknown`.

I think that it is much safer to have a fallback type detection based on the file content, not the optional EXIF metadata. Does the `MagickWand` have such functionality, or should you revert the `gomagic` removal? 
If the `MagickWand` can do this, I guess that it is better, as we'd have fewer dependencies in the project, and it might get us better coverage of different formats. But this is my assumption. What do you think?

Also, we might need to call `ffmpeg` for videos without the EXIF metadata in the same way. Since differentiating between videos and images at this time is an overhead, I'd suggest running the `exiftool`, `MagicWand` or `gomagic`, and `ffmpeg` in a sequence, breaking once we have a result, different from the `TypeUnknown` or `octet-stream`.

api/scanner/media_type/media_type_test.go
```diff
@@ -1,13 +1,66 @@
...
+
+		{"mp4.mp4", TypeMP4},
+		{"ogg.ogg", TypeOGG},
+		{"mpeg.mpg", TypeMPEG},
+		{"webm.webm", TypeWEBM},
+
+		{"avi.avi", mediaType("video/x-msvideo")},
+		{"mkv.mkv", mediaType("video/x-matroska")},
+		{"quicktime.mov", mediaType("video/quicktime")},
+		{"wmv.wmv", mediaType("video/x-ms-wmv")},
```
kkovaletp, line 43: @googollee, please add here an image and a video with no EXIF metadata to cover such scenarios as well.
While this is a corner case, I expect that our users might have some old media files in their collections with no EXIF metadata - from years where it wasn't automatically added to the files or lost during many file copy operations between different FS (not all filesystems support file metadata streams), or because of privacy concerns.

api/scanner/externaltools/exif/exif.go
googollee: It's exactly same as `return globalExifParser.ParseMIMEType(filepath)`. I hope to show the handle of errors explicitly in the code. It says clearly how errors are processed and there is no meaningful value when errors are happened.

api/scanner/externaltools/exif/test_data/IncorrectGPS.jpg
googollee: The file size should not be changed. Before this file has a execute flag `+x`. I removed that flag with `chmod -x` because I don't think the file is executable.

api/scanner/media_type/media_type_test.go
googollee: Since most of files are as small as possible, they don't contain any exif data, even no Exif data block in the file. Example:

```
$ exiftool jpeg.jpg 
ExifTool Version Number         : 13.25
File Name                       : jpeg.jpg
Directory                       : .
File Size                       : 160 bytes
File Modification Date/Time     : 2025:08:20 10:08:49+02:00
File Access Date/Time           : 2025:11:21 17:09:59+01:00
File Inode Change Date/Time     : 2025:08:20 10:08:49+02:00
File Permissions                : -rw-r--r--
File Type                       : JPEG
File Type Extension             : jpg
MIME Type                       : image/jpeg
JFIF Version                    : 1.01
Resolution Unit                 : None
X Resolution                    : 1
Y Resolution                    : 1
Image Width                     : 1
Image Height                    : 1
Encoding Process                : Baseline DCT, Huffman coding
Bits Per Sample                 : 8
Color Components                : 1
Image Size                      : 1x1
Megapixels                      : 0.000001

$ exiftool bmp.bmp 
ExifTool Version Number         : 13.25
File Name                       : bmp.bmp
Directory                       : .
File Size                       : 30 bytes
File Modification Date/Time     : 2025:08:20 10:08:49+02:00
File Access Date/Time           : 2025:11:21 17:09:11+01:00
File Inode Change Date/Time     : 2025:08:20 10:08:49+02:00
File Permissions                : -rw-r--r--
File Type                       : BMP
File Type Extension             : bmp
MIME Type                       : image/bmp
BMP Version                     : OS/2 V1
Image Width                     : 1
Image Height                    : 1
Planes                          : 1
Bit Depth                       : 24
Image Size                      : 1x1
Megapixels                      : 0.00000
```

api/scanner/media_type/media_type.go
googollee: Since real media contains videos, I'll say `exiftool` covers video mime too.

According to https://github.com/photoview/photoview/pull/1299#issuecomment-3239493719, `MagicWand` doesn't provide feature of identifying MIME type.

googollee: I tried with `ffprobe`. It doesn't output MIME Type also.

api/scanner/externaltools/exif/test_data/IncorrectGPS.jpg
kkovaletp: Oh, now I see:
GitHub shows the `100755 → 100644` in the diff, and I thought that it was the file size, while in fact, those were file permissions.

api/scanner/media_type/media_type.go
kkovaletp: Thanks for the reference. I forgot about that conversation.
Also, I did some deeper investigation of how the tool identifies MIME types, and it looks like we don't need any fallback solution: the tool really parses the file header (part of the file's main data stream) and doesn't rely on the EXIF metadata.

api/scanner/media_type/media_type_test.go
kkovaletp: Thanks. Actually, after I understood how it detects the MIME type, there is no need to add files with or without EXIF metadata, as it is unrelated to this functionality and not used.

kkovaletp:
@googollee, I did the testing of this PR using your image, and in general, it works well.

I've noticed a few issues, which I'd like to have addressed before the merge:
- The new error message doesn't look good for an end user: `2025/11/22 11:18:28 ERROR IsPathMedia (/photos/main/.DS_Store): unknown media type (/photos/main/.DS_Store)`. Can we drop the 1st part "IsPathMedia (/photos/Viber/.DS_Store): " - it duplicates the 2nd part and doesn't add any value.
- The `exiftool` is now one of our key data crawlers, but there is no INFO row during the Photoview init with its version:
```
2025/11/22 11:16:43 INFO Found magickwand worker: ImageMagick 7.1.2-8 Q16-HDRI x86_64 2300ae543:20251025 [https://imagemagick.org⁠](https://imagemagick.org/) version=1810
2025/11/22 11:16:43 INFO Found executable worker: ffmpeg version="ffmpeg version 7.1.2-Jellyfin Copyright (c) 2000-2025 the FFmpeg developers" codec=h264
2025/11/22 11:16:43 INFO Found ffprobe path=/usr/local/bin/ffprobe version="ffprobe version 7.1.2-Jellyfin Copyright (c) 2007-2025 the FFmpeg developers"
```
Can you please add a similar INFO with the exiftool path and version?
- Previously, non-media files were silently skipped, and now the tool reports an error for each of them. I think that the new behavior is better for users, but it also helped me to find an issue: the scanner invokes exiftool before processing ignore files, so even if we have some known to be unsupported by Photoview files, and we've correctly mentioned them in the corresponding `.photoviewignore` files, we still are going to be spammed by those error during each scan job run. Can you please fix this order of invocations as well?

The example from the 1st issue in this list is about such files. When I got such errors for the 1st time, I created `/photos/Viber/.photoviewignore` and `/photos/main/.photoviewignore` on my Mac with the `**/.DS_Store` inside (also tried with `.DS_Store`), but still get the same errors each time I run the scan.

googollee:
There is `exiftool` log, see https://github.com/photoview/photoview/blob/cfc49a3733e1914a6a2e89a9dbea2582af11feed/api/scanner/externaltools/exif/exif.go#L24. Since the library doesn't require the absolute path and the version, I tend not to log that info.

I didn't touch the logic of `.photoviewignore`. If there is any problem with it, it doesn't belong to this PR. Please create a new issue and provide logs about it.

googollee:
Cleaned logs.

kkovaletp:
> There is `exiftool` log

Oh, sorry. Didn't notice it)

> I didn't touch the logic of `.photoviewignore`. If there is any problem with it, it doesn't belong to this PR. Please create a new issue and provide logs about it.

I searched for the `.photoviewignore` in the code and found it only in the `scanner/scanner_user.go`. It looks like the current implementation expects only folders to be ignored, not the particular files (but I might be wrong). 
It looks like my previous assumption about the wrong order of operations during the scan job execution was wrong. The issue is that the ignore functionality is incomplete and doesn't correctly handle the case of ignoring files.
I've created the #1346

kkovaletp (dismissed):
LGTM

kkovaletp:
@googollee, I've approved the PR, but please don't forget to revert the changes in the build workflow before the merge

googollee:
> The new error message doesn't look good for an end user: 2025/11/22 11:18:28 ERROR IsPathMedia (/photos/main/.DS_Store): unknown media type (/photos/main/.DS_Store). Can we drop the 1st part "IsPathMedia (/photos/Viber/.DS_Store): " - it duplicates the 2nd part and doesn't add any value.

I thought again about this. The `unknown media type` error is a normal case for users, so it should not be displayed in the UI. It should be logged, which the code already does.

When `photoview` was using the `magic` library, it identified the `.DS_Store` file as `application/octet-stream`, which didn't generate `unknown media type`. So it doesn't show anything in the UI. Now `exiftool` doesn't care about all files, so it doesn't generate a mime type for `.DS_Store`, and turns it to `unknown media type`. It should not be an error for users here.

Please try the latest image again with the new logic.

kkovaletp:
@googollee, I've retested the image, and now it works fine for most of the files. I've downloaded samples for many different formats and scanned all of them in addition to my regular testing dataset. Here is what I found:
- I see the following log message for many unsupported files: "Unsupported media type "" for file: /photos/main/samples/images/sample1.erf", but I don't see them anymore for my ".DS_Store" files - they are somehow silently ignored now. I've removed my "/photos/main/.photoviewignore", but it didn't help. Why are they tracked differently?
- Since we already detect MIME types for all found files and log unsupported media type for unsupported ones, isn't it better to return the industry standard fallback type "application/octet-stream" instead of the empty string?
- Some formats of images and videos have "application/*" MIME types, while we ignore them in the code (we accept only "image/*" and "video/*" MIME types):
```
Unsupported media type "application/vnd.adobe.photoshop" for file: /photos/main/samples/images/sample_640×426.psd
Unsupported media type "application/x-shockwave-flash" for file: /photos/main/samples/video/sample_640x360.swf
```
- We cannot successfully parse the MIME type of the HEVC format, which is pretty popular: "Unsupported media type "" for file: /photos/main/samples/video/sample_640x360.hevc". I've downloaded several different file examples to make sure that the issue is not caused by an invalid file. Also, all of them can be played by the VLC player, so they are not corrupted. I tried to search for a way to detect its MIME type, but didn't. Maybe you can suggest something... If not, this is not a blocker for us anyway, but better to make sure. Here are my samples:

[video.zip](https://github.com/user-attachments/files/23833513/video.zip)

And also, please check the bot's nitpick comment for this PR. It looks valid...

googollee:
All the answer is the behavior of `exiftool`. 

- When `photoview` was using the `magic` library, it identified the `.DS_Store` file as `application/octet-stream`, which didn't generate `unknown media type`. So it doesn't show anything in the UI. Now `exiftool` doesn't care about all files, so it doesn't generate a mime type for `.DS_Store`, and turns it to `unknown media type`. It should not be an error for users here. For `erf` files, I guess `exiftool` returns a unsupported MIME type. I don't have such files so I can't verify it.
- I tend to differ the unknown type and unsupported type. So I'd like to keep the mime type empty if we can't identify it.
- I tend to believe `psd` is an application file, but not a photo. It's a private format of photoshop and it's not suitable for showing. `swf` is an application file too, not a video. It contains interactive behaviors. I don't know how ffmpeg converts it to preview video.
- `exiftool` doesn't support `hevc` now. See:

```sh
$ exiftool bbb.hevc 
ExifTool Version Number         : 13.25
File Name                       : bbb.hevc
Directory                       : .
File Size                       : 21 kB
File Modification Date/Time     : 2025:11:29 13:38:24+01:00
File Access Date/Time           : 2025:11:29 13:38:24+01:00
File Inode Change Date/Time     : 2025:12:02 11:23:24+01:00
File Permissions                : -rw-r--r--
Error                           : Unknown file type
```

kkovaletp:
> * .DS_Store

Ok, this is not a big deal. I'm fine with the current behavior.

> * I'd like to keep the mime type empty if we can't identify it.

Please correct me if I'm wrong, but as far as I understand, the `application/octet-stream` literally means that the content is an unknown binary stream of data that the current tool cannot identify. So, if my understanding is correct, returning the `application/octet-stream` as a fallback type follows the common practice. But if you're sure that it is better for us to return an empty string, I'd agree with you.

> * `psd` is an application file

Sure, it is a photo editing project for Adobe Photoshop, and usually it is not a final image format, but I know that many photo editors prefer to store their `psd` projects in case of future edits and/or exports - similarly to the RAW photos. That is why I asked about its support.
Of course, Photoshop is not the only professional photo editor tool: we have GIMP (which I use) and other tools, each of which stores projects in its own format, so if we start supporting such projects, we might shortly fall into a list of such formats, so your concern about supporting them is valid, and I'd agree with you if you'd say that this is out of our scope.

> * `swf` is an application file too, not a video. It contains interactive behaviors

Isn't it similar to the `gif`, which we support? The capability to convert it with the `ffmpeg` tool is a valid concern. Can you please try? If that is not supported, then it is fine to skip it as well.

> * `exiftool` doesn't support `hevc` now

Yes, this is what I noticed too. The question is, can we somehow overcome this limitation without dirty hacks? The HEVC format is becoming pretty popular, and modern phones and cameras start offering it as a video format for the recorded videos, as an alternative to the `mp4`, so I'd expect more user requests for it eventually.

googollee:
> But if you're sure that it is better for us to return an empty string, I'd agree with you.

If it's UNKONWN, we know it's not supported by `exiftool`. If it's UNSUPPORTED, we know it's not a video or photo.

> Isn't it similar to the gif, which we support?

`swf` is not `gif`. You can find some games in `swf`, but never for `gif`. See [here](https://archive.org/details/banned-from-equestria-daily-1.5). Besides interactivation, no cameras/recorders/phones generate `swf` files. I tend to not support it since it's photo gallery application, not a common video player.

> The HEVC format is becoming pretty popular.

You've confused container formats with video codec, and the raw codec format. `hevc` files in `video.zip` are raw hevc format files, which can't contain audios. We already supports container formats, like `mp4`, `mov`, `mkv`, `ogv`. These container format may contain HEVC video stream. Since we support those container formats through ffmpeg, we already support hevc stream.

I don't think raw hevc is a popular format. And it will never be. The `hevc` format is good for researching the codec, but not good for sharing/storing video. All cameras/phones/recorders generates files with a container format, for containing both video and audio (and other metadata).

kkovaletp:
@googollee, thanks for your explanation.
I have no objections now)

Please revert the build workflow, and I'll approve the PR

.github/workflows/build.yml
```diff
@@ -12,8 +12,8 @@ on:
     - cron: '18 1 * * 4'
 
 env:
-  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
-  # IS_PUSHING_IMAGES: true # Switch on when pushing images with forked repos.
+  # IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
+  IS_PUSHING_IMAGES: true # Switch on when pushing images with forked repos.
```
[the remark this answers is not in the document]
googollee: Rolled back.

googollee:
Done.

api/scanner/externaltools/exif/exiftool.go
```diff
@@ -278,3 +288,20 @@ func calculateOffsetFromGPS(fileInfo *exiftool.FileMetadata, date *time.Time) (*
...
 	return &offset, nil, nil
 }
+
+func (p *ExifParser) ParseMIMEType(mediaPath string) (string, error) {
+	fileInfo, err := p.loadMetaData((mediaPath))
+	if err != nil {
+		return "", fmt.Errorf("invalid parse %q exif: %w", mediaPath, err)
+	}
+
+	mime, err := fileInfo.GetString("MIMEType")
```
trap000d, line 298: Here the problem appears:
By default `exiftool` returns descriptions, not tags, see here: https://exiftool.org/faq.html
E.g.:
```
exiftool thumbnail_191005~1_tif_yyuLmXHr.jpg | grep MIME
MIME Type                       : image/jpeg
```
Note the space! 
To return `MIMEType` string you should run `exiftool -s FILENAME`:
```
exiftool -s thumbnail_191005~1_tif_yyuLmXHr.jpg | grep MIME
MIMEType                        : image/jpeg
```
That caused a bunch of errors in logs:
```
WARN GetMediaType() error. error="invalid parse \"/path/to/album/20170514T200812.JPG\" exif: write |1: broken pipe" file="/path/to/album/20170514T200812.JPG"
Unsupported media type "unknown" for file: /path/to/album/20170514T200812.JPG
```
empty database and thumbnail cache.

googollee: Could you provide the problem file in the issue?

kkovaletp: @trap000d, this is a really strange case, as I've tested a lot of times with different file formats, and I never saw this issue before.
What is your deployment type? Docker or directly on the host? If the 2nd, what is your OS?
I see no difference in your examples of the `exiftool` calls with and without the `-s` key.

trap000d: > @trap000d, this is a really strange case, as I've tested a lot of times with different file formats, and I never saw this issue before. 

> What is your deployment type? 

Directly on the host. O/S - Arch Linux. 

Exiftool - from official repo. 

Photoview - my own build based on corresponding AUR package.
https://github.com/trap000d/aur/tree/master/photoview-git

> I see no difference in your examples of the `exiftool` calls with and without the `-s` key.

You parse string 'MIMEType'

Without `-s` exiftool returns 'MIME type' (two words separated by space).
With `-s` exiftool returns 'MIMEType' (one word)

So I've rolled back to commit c5fc5002 and restored database and cache from backup.

kkovaletp: Thanks for the clarification. Now I see.

> Directly on the host. O/S - Arch Linux.

Well, there are many things that might impact the behavior... Would it be possible to run our `master` image with mounted media collection and scan it there - just to exclude the possible impact from the host OS and the tools?

> You parse string 'MIMEType'

Ah, yes. I didn't notice it, as I analyzed the returned values)
But somehow it works inside the image without the `-s` key... That is strange...

trap000d: Here I could only speculate about `go-exiftool` module behavior on different platforms. I've not tried to look inside it yet. For digging through such many abstraction layers I need enough of free time, which I unfortunately don't have at the moment.

trap000d: For Docker image test. Currently I've running photoview on Odroid-M1. This little thing is definitely better than e.g. Raspberry Pi, however I'm not sure if it will be able to take off with container on the board. Maybe I'll give it a try on desktop (also with Arch)...

trap000d: As you use Debian Trixie in container, I assume there are might be some differences between Debian and Arch builds and default behavior. E.g. it's possible to override set of standard arguments with config files.

With the following `~/.ExifTool_config`:
```
# Specify default exiftool command-line arguments
# - inserted before all other options except -config
# - applies to all -execute'd commands
@Image::ExifTool::UserDefined::Arguments = (
    '-s' # Show tags instead of descriptions
);

```
it adds `-s` and shows all tags as tags.
I'll test it later then let you know if this override works or not. There might be a chance to catch another bug :)

trap000d: Well, on Arch desktop I see the same issue. And it seems, the problem not in exiftool arguments, but in interprocess communication. I.e `exif: write |1: broken pipe` indicates the error.

I've strace'd running photoview process and found that all calls to exiftool end up with broken pipe:

```
newfstatat(AT_FDCWD, "/media/Pictures/20130206T133647.JPG", {st_mode=S_IFREG|0664, st_size=4892865, ...}, AT_SYMLINK_NOFOLLOW) = 0
newfstatat(AT_FDCWD, "/media/Pictures/20130206T133647.JPG", {st_mode=S_IFREG|0664, st_size=4892865, ...}, 0) = 0
write(11, "-j\n", 3)                    = -1 EPIPE (Broken pipe)
 | 00000  2d 6a 0a                                          -j.              |
--- SIGPIPE {si_signo=SIGPIPE, si_code=SI_USER, si_pid=323034, si_uid=947} ---
rt_sigreturn({mask=[]})                 = -1 EPIPE (Broken pipe)
write(11, "/media/Pictures/20130206"..., 53) = -1 EPIPE (Broken pipe)
--- SIGPIPE {si_signo=SIGPIPE, si_code=SI_USER, si_pid=323034, si_uid=947} ---
rt_sigreturn({mask=[]})                 = -1 EPIPE (Broken pipe)
write(2, "2025/12/12 19:01:31 WARN GetMedi"..., 212) = 212
write(2, "2025/12/12 19:01:31 Unsupported "..., 116) = 116
write(2, "2025/12/12 19:01:31 INFO Check t"..., 105) = 105
n
```
It looks like exiftool is called with `-j`, which means json format

At the moment I assume it all laying a bit deeper and could only speculate either it's:

1. Permissions (I start it via systemd so it might be another systemd quirk)
2. Kernel (6.12.60 LTS on one machine and 6.17.9 on another, 6.12 in Trixie - irrelevant)
3. Go build flags (I copypasted them from original PKGBUILD with no reviewing)

I agree, it makes sense to run Docker image to see what happens. Tomorrow...

trap000d: OK, looks like it's systemd. I just started photoview as ordinary user from command line, and all works well. So I suppose systemd drops process privileges.

UPD. I can confirm, it's systemd, but I'm still have no clear idea what exactly happens.

kkovaletp: Great debugging!
Let's try asking the bot... Maybe he knows some useful tips...

@coderabbitai, please read this conversation thread and carefully analyze all the problem symptoms and debugging results discovered so far. Please search the web for all relevant info and suggest the most likely root causes and further debugging steps to confirm or reject each. Also, if you have pretty high confidence, you can suggest mitigation/fix steps for those root causes.

trap000d: Few more updates in regarding to systemd behavior. Honestly, I have a little idea why this happens.
`/usr/bin/vendor_perl` is an Arch 'feature' - is added to $PATH (otherwise photoview won't start)

If I start it via systemd unit, it creates only a single photoview process:
```
[root@desktop]# systemctl status photoview.service                                                                                                                                                           
● photoview.service - Photoview - Photo gallery for self-hosted personal servers                                                                                                                                  
     Loaded: loaded (/etc/systemd/system/photoview.service; disabled; preset: disabled)                                                                                                                           
     Active: active (running) since Fri 2025-12-12 22:15:14 NZDT; 7s ago                                                                                                                                          
 Invocation: 35de076e241645bdb7777939a8b2e009                                                                                                                                                                     
       Docs: https://photoview.github.io/docs                                                                                                                                                                     
   Main PID: 1180290 (photoview)                                                                                                                                                                                  
      Tasks: 11 (limit: 37599)                                                                                                                                                                                    
     Memory: 47.9M (peak: 49.8M)                                                                                                                                                                                  
        CPU: 310ms                                                                                                                                                                                                
     CGroup: /system.slice/photoview.service                                                                                                                                                                      
             └─1180290 photoview
```

If I wrap it with a shell script (or start from command line), then an exiftool process is spawned and then everything works just fine:
```
[root@desktop]# systemctl status photoview
● photoview.service - Photoview - Photo gallery for self-hosted personal servers
     Loaded: loaded (/etc/systemd/system/photoview.service; enabled; preset: disabled)
     Active: active (running) since Fri 2025-12-12 22:24:32 NZDT; 15min ago
 Invocation: 4ff2f636dcad43c493e68a48faf7357a
       Docs: https://photoview.github.io/docs
   Main PID: 101414 (photoview.sh)
      Tasks: 18 (limit: 8831)
     Memory: 634.3M (peak: 639M)
        CPU: 11min 27.039s
     CGroup: /system.slice/photoview.service
             ├─101414 /bin/bash /var/lib/photoview/photoview.sh
             ├─101441 photoview
             └─101856 perl /usr/bin/vendor_perl/exiftool -stay_open True -@ - -common_args -n

```

googollee: `bash` will load `~/.bashrc`, which may set `PATH` to look up binaries. IIRC, `go-exiftool` relies on `PATH` env to run `exiftool`. Could you paste logs when the problem happens?

trap000d: In brief, I've tried everything suggested by @coderabbitai (most of things - before, few remaining - just checked right now)
In particular, I haven't spotted any signs of spawned exiftool process in strace logs, changes in service file don't make any difference.

I assume we're at the point of 'If it Still Failing'

As I've mentioned, I overridden it by starting photoview in shell script...

trap000d: > @trap000d, excellent detective work! You've confirmed the root cause: **the exiftool binary can't be found because systemd's minimal PATH doesn't include `/usr/bin/vendor_perl`**.
> ## The Fix
> 
> The `Environment=PATH=...` directive I suggested earlier needs to include the **full path including `/usr/bin/vendor_perl`**. Here's the corrected fix for your systemd service file:
> 
> ```ini
> [Service]
> ExecStart=/usr/lib/photoview/photoview
> WorkingDirectory=/usr/lib/photoview
> User=photoview
> Group=photoview
> 
> # CRITICAL: Include the directory where exiftool is located on Arch
> Environment="PATH=/usr/local/bin:/usr/bin:/bin:/usr/bin/vendor_perl"
> 
> # Other recommended settings
> LimitNOFILE=65536
> ```
> 
> After updating:
> 
> ```shell
> sudo systemctl daemon-reload
> sudo systemctl restart photoview.service
> ```
> 
> ## Verification Steps
> 
>     1. **Confirm exiftool location on your system:**
>        ```shell
>        which exiftool
>        # Should show: /usr/bin/vendor_perl/exiftool
>        ```
> 
>     2. **Check that the photoview user can execute it:**
>        ```shell
>        sudo -u photoview /usr/bin/vendor_perl/exiftool -ver
>        ```
> 
>     3. **After restart, verify exiftool child process spawns:**
>        ```shell
>        systemctl status photoview.service
>        # Should show both photoview AND perl exiftool processes in CGroup
>        ```
> 
> 
> ## Alternative: Use Absolute Path in Environment
> 
> If the above still doesn't work, you can also explicitly tell the exiftool where to find the binary by using an absolute path wrapper:
> 
> ```ini
> Environment="EXIFTOOL_HOME=/usr/bin/vendor_perl"
> Environment="PATH=/usr/local/bin:/usr/bin:/bin:/usr/bin/vendor_perl"
> ```
> 
> ## Why Previous Suggestions Didn't Work
> 
> My earlier `Environment="PATH=/usr/local/bin:/usr/bin:/bin"` suggestion was missing `/usr/bin/vendor_perl`, which is an Arch-specific location. On Debian (used in Docker images), exiftool is at `/usr/bin/exiftool`, so it works with the standard PATH. This is why the Docker image works fine but your host installation doesn't.
> ## Potential Code Improvement
> 
> @googollee, there's a potential improvement we could make to the code: when `NewExiftool()` fails (likely because exiftool binary isn't found), we should log a more descriptive error at initialization time rather than letting it fail later with "broken pipe" errors. Looking at `api/scanner/externaltools/exif/exif.go`, the `Initialize()` function could add better error context:
> 
> ```go
> func Initialize() error {
>     parser, err := NewExifParser()
>     if err != nil {
>         return fmt.Errorf("failed to initialize exif parser (is exiftool installed and in PATH?): %w", err)
>     }
>     // ...
> }
> ```
> 
> This would make it immediately obvious if exiftool can't be found, rather than discovering it later with cryptic pipe errors.
> 
> Let me know if this fixes your issue!
> 🧠 Learnings used


Nope, I've added PATH to service file 10 hours ago. It still fails:
```
[Unit]
Description=Photoview - Photo gallery for self-hosted personal servers
Documentation=https://photoview.github.io/docs
After=network.target

[Service]
ExecStart=/usr/lib/photoview/photoview
WorkingDirectory=/usr/lib/photoview
User=photoview
Group=photoview
IgnoreSIGPIPE=no

ReadWritePaths=/var/lib/photoview
Environment="HOME=/var/lib/photoview"
Environment=PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl
EnvironmentFile=/etc/photoview.env
CacheDirectory=photoview/media_cache

[Install]
WantedBy=multi-user.target

```

trap000d: > `bash` will load `~/.bashrc`, which may set `PATH` to look up binaries. IIRC, `go-exiftool` relies on `PATH` env to run `exiftool`. Could you paste logs when the problem happens?

There is no ~/.bashrc for `photoview` user (it's a system one with no HOME and `/bin/nologin` as shell), PATH in systemd service file is set for pointing to `/usr/bin/vendor_perl`.
AI bot hasn't suggested anything which I've not tried already.

Except their recommendation to add some more informative error messages. If I'd started debugging, I would do the same :)

googollee: Is there any log about the failure of launching `exiftool`?

trap000d: Nothing in logs. Looks lite there is no attempt to start exiftool at all.

For testing a hypothesis of custom PATH's  I've copied `exiftool` to /usr/bin, and it also won't spawn if photoview is started via systemd.

I'm wondering if it could be be a wild combination of systemd, perl and go-exiftool?

In short, when I start photoview from command line/shell script, child exiftool process appears as:
```
perl /usr/bin/vendor_perl/exiftool -stay_open True -@ - -common_args -n
```
It looks like go-exiftool starts it as `perl exiftoolBinary` - and it would fail even if `exiftool` is in /usr/bin
Somehow in shell it points to the right location, but from systemd it doesn't expand to path.
I suppose if you want to catch it, we need to add some debug output to see what it's trying to start.

```
[photoview@desktop photoview]$ perl exiftool
Can't open perl script "exiftool": No such file or directory

[photoview@desktop photoview]$ perl /usr/bin/exiftool | head
NAME
    exiftool - Read and write meta information in files

SYNOPSIS
  Reading
    exiftool [*OPTIONS*] [-*TAG*...] [--*TAG*...] *FILE*...

  Writing
    exiftool [*OPTIONS*] -*TAG*[+-^<]=[*VALUE*]... *FILE*...

```

P.S. I've found an override, but this issue is appearing so curiously puzzling that I can't stop digging deeper :)

trap000d: Ok. Caught it after stracing of systemd unit.

```
ExecStart=/usr/bin/strace --absolute-timestamps=precision:us --daemonize --follow-forks --output=/tmp/strace.log /usr/lib/photoview/photoview
```

It appears systemd doesn't export variables, so all children processes of main process would start in a 'clean' environment. Effectively, exiftool in Arch (and, I assume, in other distributions with non-standard location) can't be found in a standard location. Solution is to add PATH line to `/etc/photoview.env`

```
PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/bin/vendor_perl
```
Note: it MUST be a full PATH because systemd ignores $ variables, so expansion like `PATH=/usr/bin/vendor_perl:$PATH` doesn't work.

kkovaletp: Great that you've found the root cause!
I remember this issue from a few different cases of running services and scheduled tasks on Linux.
I thought you already mitigated this issue in your setup, as it is pretty common. That is why I didn't mention it earlier.
Anyway, congrats on finding a fix!

PR #1353, opened 2025-12-13
Improving the GraphQL auth workflow

kkovaletp:
## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened auth and WebSocket flows with standardized "invalid authorization token" responses, improved logging, and safer handling for missing, malformed, or unavailable tokens.

* **New Features**
  * WebSocket now sends stored auth token when available to support real-time auth.
  * Pagination now raises an explicit error when required paginate data is missing.

* **Refactor**
  * Safer dataloader/context handling using typed context keys.

* **Tests**
  * Added comprehensive tests for WebSocket auth, middleware, context user utilities, and GraphQL endpoint.

### DIFF
--- a/api/graphql/auth/auth.go
+++ b/api/graphql/auth/auth.go
@@ -17,6 +17,7 @@ var ErrUnauthorized = errors.New("unauthorized")
 var bearerRegex = regexp.MustCompile("^(?i)Bearer ([a-zA-Z0-9]{24})$")
 
 const INVALID_AUTH_TOKEN = "invalid authorization token"
+const INTERNAL_SERVER_ERROR = "internal server error"
 
 // A private key for context that only this package can access. This is important
 // to prevent collisions between different context uses
@@ -35,7 +36,7 @@ func Middleware(db *gorm.DB) func(http.Handler) http.Handler {
 				loaders := dataloader.For(r.Context())
 				if loaders == nil {
 					log.Error(r.Context(), "Dataloader not available in HTTP context")
-					http.Error(w, "internal server error", http.StatusInternalServerError)
+					http.Error(w, INTERNAL_SERVER_ERROR, http.StatusInternalServerError)
 					return
 				}
 
@@ -105,7 +106,7 @@ func AuthWebsocketInit() func(context.Context, transport.InitPayload) (context.C
 		loaders := dataloader.For(ctx)
 		if loaders == nil {
 			log.Error(ctx, "Dataloader not available in websocket context")
-			return nil, nil, errors.New("internal server error")
+			return nil, nil, errors.New(INTERNAL_SERVER_ERROR)
 		}
 
 		user, err := loaders.UserFromAccessToken.Load(*token)


api/graphql/auth/auth.go
```diff
@@ -72,31 +81,37 @@ func UserFromContext(ctx context.Context) *models.User {
...
 		}
 
 		user, err := dataloader.For(ctx).UserFromAccessToken.Load(*token)
-		// user, err := models.VerifyTokenAndGetUser(db, *token)
 		if err != nil {
-			log.Printf("Invalid token in websocket: %s\n", err)
-			return nil, fmt.Errorf("invalid authorization token")
+			log.Printf("Error loading user from token (websocket): %s\n", err)
+			return nil, nil, errors.New(INVALID_AUTH_TOKEN)
+		}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, please carefully analyze the external code mentioned by you and tell me how probable those cases are, leading to panic.
If they are possible, please suggest the exact code replacements to handle them consistently and reliably.

api/graphql/auth/auth.go
kkovaletp, line 1: @coderabbitai, please analyze the code changes in this PR (this file and the `api/graphql/endpoint/graphql_endpoint.go`) in the context of the UI code implementation in the product.
Please make sure that the UI code is fully compatible with these changes and doesn't require updates, directly related to these changes.
If you see some inconsistencies, which would lead to communication errors and/or unexpected behavior, please clearly describe them and the cases in which they would cause issues. Also, suggest consistent and reliable code replacements to fix them.

api/graphql/auth/auth.go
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, thanks for the fix. Please investigate why the dataloader context might not be passed to the WebSocket InitFunc and how to mitigate those cases (if we should). Please provide the exact code replacements for each reasonable mitigation, making sure that they are consistent, solid, and reliable.

api/graphql/auth/auth.go
```diff
@@ -72,31 +81,43 @@ func UserFromContext(ctx context.Context) *models.User {
...
 		}
 
 		// put it in context
 		userCtx := context.WithValue(ctx, userCtxKey, user)
 
 		// and return it so the resolvers can see it
-		return userCtx, nil
+		// Return nil for the InitPayload acknowledgment (no custom ack payload needed)
+		return userCtx, nil, nil
 	}
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, please reevaluate your review comment. It looks like this is not an issue at all, as if I remove the `*transport.InitPayload` from the returned values, I get an incompatibility error in the `graphql/endpoint/graphql_endpoint.go` line 34. Also, the current implementation passes all the unit tests.

kkovaletp:
@googollee, finally, I've resolved all bot's comments.
Please review.

api/graphql/auth/auth.go
kkovaletp, line 1: @coderabbitai, please carefully analyze this PR in the scope of the product code (UI and API).
Please confirm that the auth workflow works fine for the cases when there is valid communication with unauthorized UI sessions (like notifications on the login page, when no user token is expected).

api/graphql/auth/auth.go
```diff
@@ -72,31 +88,43 @@ func UserFromContext(ctx context.Context) *models.User {
...
-			return nil, err
+			log.Error(ctx, "Invalid bearer format (websocket)", "error", err)
+			return nil, nil, err
+		}
+
+		loaders := dataloader.For(ctx)
+		if loaders == nil {
+			loaders = dataloader.NewLoaders(db)
+			ctx = context.WithValue(ctx, dataloader.GetLoadersKey(), loaders)
 		}
```
googollee, line 110: I suggest returning an error here if dataloader doesn't exist. If dataloader doesn't exist, it means `dataloader.Middleware()` doesn't run and there is no loader injected. Blindly creating a new one just repeats the logic of `dataloader.Middleware()`. Since all middleware is constructed by the code, it should depend on `dataloader.Middleware()` but not repeat it again.

And please remove `dataloader.NewLoader()`. All loader should be constructed by `dataloader.Middleware()`.

kkovaletp: Good point! 
Done

api/graphql/auth/auth.go
```diff
@@ -72,31 +89,43 @@ func UserFromContext(ctx context.Context) *models.User {
 	return raw
 }
 
-func AuthWebsocketInit(db *gorm.DB) func(context.Context, transport.InitPayload) (context.Context, error) {
-	return func(ctx context.Context, initPayload transport.InitPayload) (context.Context, error) {
+func AuthWebsocketInit() func(context.Context, transport.InitPayload) (context.Context, *transport.InitPayload, error) {
```
googollee, line 92: I'm surprised this function has not been used before. Thanks for fixing it.

kkovaletp: Sure, I was surprised as well)

PR #1355, opened 2025-12-17
fix cookie operations for authentication

### LINKED ISSUE #891: Weird behaviour when inputing passwords with a trailing backslash
### (Case 1)
Frontend logic removes (at least) trailing backslashes from a password input when accessing a protected share. 
I had a quick look, and it's probably due to this:
https://github.com/photoview/photoview/blob/8cdc4cd136dfd5e3785df6b209a8befac8d7e1cb/ui/src/helpers/authentication.ts#L20-L25

The resultant behavior is:
1. Create a share with a password (let's say password is only [a-zA-Z]+) 
2. Access the share as an unathenticated user
3. Input the password with a trailing backslash (or backslashes)
4. You are authenticated and share is shown

Expected behavior:
user is not authenticated and not authorized to the share.

### (Case 2)
I also checked the behavior when the password itself had a trailing backslash. For example, when password is `a\`, using a password that matches `a\\*` does not authorize to the share, sometimes throwing an internal server error. When the share password is simply `\`, then inputing `\` results in:
```
2023/09/06 23:50:04 Did not find auth-token cookie
runtime error: invalid memory address or nil pointer dereference

goroutine 2257323 [running]:
runtime/debug.Stack()
...
```
### also
Sometimes (??) when inputing password+trailing slashes I'm authorized to the share, but the image doesn't display (link to the image is unauthorized). In that case though, there are exif settings visible, so there is still info leak:
image

Expected behavior:
just normal password stuff

**Your environment:**
Nothing too useful, Docker-compose, Ubuntu server, postgres

version: undefined? 
I pull from `viktorstrate/photoview:master`, `4f1126b4d5b3`
image

emadalam:
Following has been implemented:

- added [`js-cookie`](https://github.com/js-cookie/js-cookie) library for setting authentication cookies
- fix issue with password protected shares that contained passwords with special characters
- improved test coverage

See original issue for details. Fixes #891 

## Summary by CodeRabbit

* **Chores**
  * Added js-cookie dependency for improved cookie management infrastructure.

* **Tests**
  * Added comprehensive test coverage for authentication and password handling in share links, including token validation and password storage scenarios.


ui/src/helpers/authentication.ts
```diff
@@ -1,25 +1,47 @@
+import Cookies from 'js-cookie'
+
+const COOKIE_DEFAULT_OPTIONS = {
+  path: '/',
+  sameSite: 'Lax',
+} as Partial<Cookies.CookieAttributes>
+
+const AUTH_TOKEN_MAX_AGE_IN_DAYS = 14
```
kkovaletp, line 8: @emadalam, this value was hardcoded to 14 days before your change.
Can we get it from the API? Can we configure it on the server with an env var?
Since this is an old implementation, this question is optional for this PR, but nice to improve it if it is easy to do in a reliable way.

ui/src/helpers/authentication.ts
```diff
@@ -1,25 +1,47 @@
...
+  const cookieName = SHARE_TOKEN_COOKIE_NAME(shareToken)
+
+  Cookies.set(cookieName, password, COOKIE_DEFAULT_OPTIONS)
+}
+
+export function clearSharePassword(shareToken: string) {
+  const cookieName = SHARE_TOKEN_COOKIE_NAME(shareToken)
+
+  Cookies.remove(cookieName)
 }
```
kkovaletp, line 41: @emadalam, I see this new function. It looks useful, but where is it actually used in the product code?
I guess it should be used.

ui/src/helpers/authentication.ts
emadalam: Yes, it was a hardcoded value and to make it configurable was out of scope of this PR.

As I mentioned before my implementation, I'm only doing a fix with minimal code changes that doesn't touch other parts of the existing setup. Trying to make this value configurable would have certainly touched many parts of the app.

ui/src/helpers/authentication.ts
emadalam: For now it's just in line with the other `clearTokenCookie` function to keep everything consistent in this module. I mainly added this for now for the sake of completeness of test suits for this module as previously there were no tests.

It's used in the test suite to test auth token and share password functionality to be working as expected.

ui/src/helpers/authentication.ts
kkovaletp, line 1: @coderabbitai, please analyze this PR carefully in the scope of the product end-to-end shares management flow.
Do you see some inconsistencies in this PR? Does it cover all possible cases? Is the auth workflow fully aligned with the API? Will this implementation work for different product deployment types (simple Docker deployment for local network access, direct installation on the host, running both behind a reverse proxy to let users from the Internet access the site and shares, integrations with CDNs, etc.)? Please note that we need to focus only on the PR changes and potentially fix issues in the functionality that was working in the `master` and broke in this PR. If something didn't work in the `master` branch as well, it is out of this PR scope (but good to know for future fixes).

kkovaletp: @emadalam, can you please take a look at the following type mismatch:

> 4\. **Type mismatch**: `AlbumSharePage` expects `password: string | null`, but `getSharePassword` returns `string | undefined`. TypeScript should flag this if it's problematic, but it's minor.

And would you like to create separate PR(s) to consistently handle the other 3 bot's observations from the previous message in this thread?

emadalam: All those bot comments are irrelevant in the given context.

1. Adding `secure: true` to the cookie will break the deployments. This is a self hosted app and many would run it in their local homelab that's not always served over https. Making that assumption will break the auth flow.
2. Same thing here, many people access it on localhost or their internal ip, domain is irrelevant.
3. Existing behaviour, not changed as that requires more effort to show a logout button for share link page, increasing the scope drastically.
4. Existing behaviour, typescript would have complained on compilation, which isn't the case.

> Observations (Not Regressions)
> These issues existed in master

If you read carefully, the PR preserves entire existing behaviour from master and only fixes the core issue without producing any regressions or unintended side-effects elsewhere.

kkovaletp: Thanks for the clarification about point 4. As you confirm that this is not an issue, we can go with it.
For all other 3 points, I didn't say that those are issues for your PR. Instead, I thought about them as potential improvements, implemented in a smart way in new PRs:
1. Adding the `secure` option can be driven by a configuration variable, disabled by default
2. The domain setting also should be optional, unset by default
3. If manual logout is the only use case for the cleanup share cookies, it is not so important, but I thought there might be some other use cases (switching between different shares from the same browser, etc)

kkovaletp (approved):

ui/src/helpers/authentication.ts
emadalam: Definitely, the first 2 points can be made configurable as an improvement after careful examination of the whole architecture 🖖

Issue #1163, opened 2025-01-26
Duplicate images if person exists multiple times in the same picture


marc7s:
**Describe the bug**
If the same person has been identified multiple times in the same picture, an error is shown in the browser console of duplicate keys, and that picture will be displayed twice on the `/people` page for that person. This is a corner case, but something that I could never track down during my work on #1162 until now. During the development, I would just select random people and merge them together to test the feature, without caring that I was merging different people. This would sometimes (when for that person, there was at least one image with faces tied to the same `face_group_id` present multiple times in the same picture).

The real world scenario would be panorama images, where you moving around during the time it is being taken to appear multiple times in the same picture. Otherwise I'm not sure there are many cases where this would occur. However, it is still a bug and should be fixed. Especially since it will help with future development work on the face merging.

I have not looked into this issue so I cannot determine whether it is a frontend issue or backend issue, or both. However, my guess would be a backend issue since I think it is related to the SQL query returning duplicates, then causing the frontend to display elements using a key that should be unique but is not since there are duplicates.

**To Reproduce**
1. Have an image with multiple people in, say Person A and Person B
2. Merge Person B with Person A
3. Go to the page for Person A (which now also contains Person B since they are merged)
4. The image in step 1 will appear twice and you will have errors in the browser console

**Expected behavior**
The same face appearing multiple times in a single image should be handled without errors.

**Screenshots**
<img width="625" alt="Image" src="https://github.com/user-attachments/assets/15c8b7d5-08f7-43f2-9621-d2c857f9ec87" />

**Your environment:**

 - Are you using Docker? [yes]
 - Database [current `master` branch, using mySQL]
 - Browser [chrome]
 - Version [current `master` branch]

kkovaletp:
@marc7s, thanks. This is a good catch!
Actually, there are several valid reasons for this case to happen, for example:
- twins (and I don't expect our face recognition engine to distinguish them)
- a photo of a person in front of a banner with the same person on it (mostly relevant for famous people)
- collages and photo series
- etc

So, it would be good to have this bug fixed and make sure that Photoview accepts and correctly handles the case when the same face appears multiple times on the same photo.

KarlGoodMan:
<img width="2485" height="1387" alt="Image" src="https://github.com/user-attachments/assets/a62cdaac-2336-45a1-b620-f997f49755c9" />
@kkovaletp I’ve been able to reproduce this bug. I’m still investigating the root cause, but I’d like to try addressing it.

[...]

PR #1357, opened 2025-12-21
Fix multi-face photo duplication during photo group merge (Bug #1163).

KarlGoodMan:
The duplication occurs because multi-face photos can belong to multiple face groups and this was not accounted for during the merge, resulting in multiple database records pointing to the same photo. Deduplication has been added to the merge transaction to resolve this issue.
database

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup during face-group consolidation to remove duplicate image associations so each media item keeps a single image-face entry, ensuring data consistency.

* **Tests**
  * Added an automated test that verifies consolidation results in unique media associations and asserts duplicate image-face records are filtered.

### DIFF
--- a/api/graphql/resolvers/faces_test.go
+++ b/api/graphql/resolvers/faces_test.go
@@ -1,0 +1,106 @@
+package resolvers
+
+import (
+	"context"
+	"testing"
+
+	"github.com/photoview/photoview/api/graphql/auth"
+	"github.com/photoview/photoview/api/graphql/models"
+	"github.com/photoview/photoview/api/scanner/face_detection"
+	"github.com/photoview/photoview/api/test_utils"
+)
+
+func TestCombineFaceGroups(t *testing.T) {
+	test_utils.FilesystemTest(t)
+	db := test_utils.DatabaseTest(t)
+	face_detection.InitializeFaceDetector(db)
+	pass := "1234"
+	user, err := models.RegisterUser(db, "test_user", &pass, true)
+	if err != nil {
+		t.Fatal("register user error:", err)
+	}
+	db.AutoMigrate(&models.ImageFace{}, &models.FaceGroup{}, &models.Media{}, &models.Album{})
+	tests := []struct {
+		name string
+		dest int
+		src  []int
+	}{
+		{
+			name: "merge multiple combinations with duplicates",
+			dest: 1,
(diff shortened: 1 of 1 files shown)


api/graphql/resolvers/faces.go
```diff
@@ -192,7 +192,18 @@ func (r *mutationResolver) CombineFaceGroups(ctx context.Context, destinationFac
...
+			Where("face_group_id = ?", destinationFaceGroup.ID).
+			Group("media_id")
+
+		err := tx.Where("face_group_id = ?", destinationFaceGroup.ID).
+			Where("id NOT IN (?)", subQuery).
+			Delete(&models.ImageFace{}).
+			Error
+		if err != nil {
+			return err
+		}
```
kkovaletp, line 206: @KarlGoodMan, Please don't manually edit files, generated by GQLgen:

> // This file will be automatically regenerated based on the schema, any resolver
> // implementations
> // will be copied through when generating and any unknown code will be moved to the end.
> // Code generated by github.com/99designs/gqlgen version v0.17.84

Instead, please modify the corresponding schema file

KarlGoodMan: @kkovaletp As noted in the file header: "any resolver implementations will be copied through when generating." This means gqlgen will safely preserve the logic I wrote here even when the schema is updated and the code is regenerated. I'm only implementing the business logic, not changing the boilerplate.

kkovaletp: @googollee, please review this PR

googollee, line 206: The code looks good. It's better to add some test cases to cover the new code. The test cases can only cover this change, but not other existing logic.

@KarlGoodMan Do you want to do that in this PR or shall we do it later?

KarlGoodMan: @googollee The added code performs a SQL operation. To verify it, I plan to simulate two datasets with overlapping entries and perform a merge to test the logic.

Please let me know if there are any other scenarios I should cover—I’m happy to include them in this PR.

kkovaletp:
@googollee, please take a look at this PR after the tests are committed.
If everything is good, please approve and merge it.

googollee (approved):
LGTM

Issue #1364, opened 2026-01-03
media.date_shot not updated from EXIF data due to order-of-operations bug


jdevora:
**Describe the bug**
 When photos are scanned, the `media.date_shot` field is not being updated with the EXIF date, causing photos to be sorted by file modification time instead of the actual shot date in the timeline view.

**To Reproduce**
  1.- Take an old photo and change the file date (either using touch or by editing the EXIF metadata).
  2.- Import it as a new photo (use a different filename or import it from a different directory) to force an EXIF scan
  3- The application orders the photo in the timeline based on the file modification date instead of the EXIF date.

**Expected behavior**
The expected behavior is that `media.date_shot` is populated from the EXIF timestamp, and the file date is only used as a fallback.

**Screenshots**

**Your environment:**
Feel free to remove and add what you think might be useful.

 - Are you using Docker? yes 
 - Database tested with  mariadb  and sqlite
 - Version [e.g. 2.0.0]

**Additional context**
After taking a look at the code, it looks like the issue is an order-of-operations bug.
The assignment happens after the database save, so it only updates the in-memory object, not the database.

In `exif_task.go` (lines 66–71), we just need to move the line:
 `media.DateShot = *exifData.DateShot`
before 
  `tx.Save(media)`

kkovaletp:
@jdevora, thanks for reporting the issue, and sorry for the delay. I've just checked the code you refer to:

https://github.com/photoview/photoview/blob/f5e1a515590f8e1438292e8c138e57a506ff183e/api/scanner/scanner_tasks/exif_task.go#L66-L71

Actually, the order of operations here is correct:
- The DB save operation goes 1st, and if it is unsuccessful, the code stops and keeps both the DB and the running app consistent
- The app's in-memory object is updated after the DB successful save, so we're sure that both data sources are in a consistent state again.

Please double-check your log for any relevant errors, as it looks like the root cause of your issue is somewhere else:
- DB save was unsuccessful
- The EXIF metadata field with the date shot is not in the list of fields, we read: https://github.com/photoview/photoview/blob/f5e1a515590f8e1438292e8c138e57a506ff183e/api/scanner/externaltools/exif/exiftool.go#L110-L176

jdevora:
> [@jdevora](https://github.com/jdevora), thanks for reporting the issue, and sorry for the delay. I've just checked the code you refer to:
> 
[...]

PR #1365, opened 2026-01-03
Bugfix faulty date shot update

jdevora:
As described in issue #1364, there is a bug that prevents media.date_shot from being populated correctly.

This PR includes a small code change that fixes the issue.

It also includes an additional change that allows the information sidebar to be closed by clicking the “i” icon again.

If this second change makes the PR too broad, I’m happy to remove it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure image shot dates from EXIF are correctly persisted during media processing.

* **New Features**
  * Timeline gallery sidebar now toggles: clicking the active image closes the sidebar for smoother navigation.

* **Style**
  * Media thumbnail icons are now partially visible on small screens to improve discoverability on touch devices.

### DIFF
--- a/ui/src/components/photoGallery/MediaThumbnail.tsx
+++ b/ui/src/components/photoGallery/MediaThumbnail.tsx
@@ -70,6 +70,11 @@ const HoverIcon = styled.button`
   }
 
   transition: opacity 100ms, background-color 100ms;
+
+  @media (max-width: 1000px) {
+    /* Always show icons on mobile/touch devices */
+    opacity: 0.5;
+  }
 `
 
 type FavoriteIconProps = {


ui/src/components/photoGallery/MediaThumbnail.tsx
```diff
@@ -70,6 +70,11 @@ const HoverIcon = styled.button`
   }
 
   transition: opacity 100ms, background-color 100ms;
+
+  @media (max-width: 1000px) {
+    /* Always show icons on mobile/touch devices */
+    opacity: 0.5;
+  }
```
kkovaletp, line 77: @jdevora, I guess that this code impacts all the devices/browsers, as there is no condition here. How difficult would it be to conditionally limit the impact of this code to the devices with touch screens?
BTW, what exactly does the `@media (max-width: 1000px)` line do in the scope of our code?

kkovaletp:
@jdevora, thanks for the PR.

> It also includes an additional change that allows the information sidebar to be closed by clicking the “i” icon again.
> 
> If this second change makes the PR too broad, I’m happy to remove it.

In general, both changes are small, so there is no problem with keeping such ones together; however, at the time of writing this comment, I'm not sure that we even need to merge the change for the EXIF parser (as I wrote in the linked issue).
So, could you please create a separate PR for the UI changes and address my comment, while keeping this PR focused on the EXIF parser change?

ui/src/components/photoGallery/MediaThumbnail.tsx
jdevora: @kkovaletp You're correct. I followed the existing `@media (max-width: 1000px)` convention from `MediaGallery.tsx:26` and `Messages.tsx:15`.

  However, after further investigation, **the application primarily uses 1024px as its mobile breakpoint** - Tailwind's `lg:` breakpoint is used extensively across the main layout components to distinguish between mobile and desktop layouts. The `1000px` breakpoint only exists in those two styled-components files.

  For the gallery icons, which approach would you prefer?

  1. `@media (max-width: 1024px)` - Match the predominant mobile breakpoint used throughout the app
  2. `@media (hover: none) and (pointer: coarse)` - Targets the main issue (lack of hover capability) and avoids affecting desktop browsers with small windows
  3. `@media (max-width: 1024px), (hover: none) and (pointer: coarse)` - Small screens OR touch devices

  I'd lean toward option 1 or 3 for consistency with the main layout system.

jdevora:
> @jdevora, thanks for the PR.
> In general, both changes are small, so there is no problem with keeping such ones together; however, at the time of writing this comment, I'm not sure that we even need to merge the change for the EXIF parser (as I wrote in the linked issue). So, could you please create a separate PR for the UI changes and address my comment, while keeping this PR focused on the EXIF parser change?

 @kkovaletp I do believe the EXIF parser bug is real. I commented addressing your concerns in #1364.

  Once that discussion is resolved, I'll be happy to follow your suggestion to split this PR or proceed as needed based on your feedback.

ui/src/components/photoGallery/MediaThumbnail.tsx
kkovaletp: @jdevora, I'd choose option 3, as it merges 1 and 2 together, providing more consistent and better UX. Am I right? Do you see a negative impact from it?

kkovaletp:
As I wrote in the issue, I'd like to wait for @googollee  to review this backend change, as he wrote that code and better understands the nuances.

Issue #755, opened 2022-10-10
Server error: internal system error at (initialSetupWizard)


Smiggel:
**Describe the bug**
Added stack (below) and visited the page. I tried to complete the setup. Set the username and password. As filepath I set /photos.

Changing the path to something else not used in the stack works. But when I try to update the path in the settings, I can not set it to /photos either.

**To Reproduce**
1. User stack below to setup docker with own variables.
2. Go to the interface
3. Fill username and password
4. Set path to /photos

**Expected behavior**
After submitting, I expected the setup to complete.

**Screenshots**
<img width="950" alt="Schermafbeelding 2022-10-10 om 10 19 55" src="https://user-images.githubusercontent.com/1825953/194824497-b49af202-86df-4f2d-aeff-f782ea907be1.png">

**Your environment:**
Raspberry PI 4 + OpenMediaVault + Docker

 - Are you using Docker? [yes]
 - Database [mariadb 10.x]
 - Browser [Firefox 105.0.2]
 
**Additional context**
```
version: "3"

services:
  photoview:
    image: viktorstrate/photoview:2
    restart: always
    ports:
      - "5000:80"
    environment:
      - PUID=1000
      - PGID=1000
      - PHOTOVIEW_DATABASE_DRIVER=mysql
      - PHOTOVIEW_MYSQL_URL=photoview:photoview@tcp(192.168.1.94:3306)/photoview
      - PHOTOVIEW_LISTEN_IP=photoview
      - PHOTOVIEW_LISTEN_PORT=80

      # Change This: to the directory where your photos are located on your server.
      # If the photos are located at `/home/user/photos`, then change this value
      # to the following: `/home/user/photos:/photos:ro`.
      # You can mount multiple paths, if your photos are spread across multiple directories.
      - /srv/dev-disk-by-uuid-6da15416-013d-4e63-930e-2bd24701f958/Pictures:/photos:ro
```

Smiggel:
Logs:

> 2022/10/10 08:32:39 Warn: invalid root path: '/photos'
> 
> stat /photos: no such file or directory
> 
> runtime error: invalid memory address or nil pointer dereference
> 
> goroutine 105 [running]:
> 
> runtime/debug.Stack()
> 
> 	/usr/lib/go-1.18/src/runtime/debug/stack.go:24 +0x68
> 
> runtime/debug.PrintStack()
> 
> 	/usr/lib/go-1.18/src/runtime/debug/stack.go:16 +0x20
> 
[...]

PR #1367, opened 2026-01-05
Add breadcrumb navigation to shared albums

### LINKED ISSUE #1366: Add breadcrumb when you share an album
When sharing an album that contains sub-albums, once a user navigates into one of them, there is no way in the UI to return to the parent album or even to the root album.

Image

Display the same breadcrumb navigation to the parent album that is available when the user is logged in.

Using the browser's back button 

**Additional context**
It should visually match (or mimic) the breadcrumbs shown when the user is logged in.

Image

jdevora:
When sharing an album with subalbums, users can now navigate back
  to parent albums using breadcrumb navigation that matches the
  logged-in user experience.

  Backend changes:
  - Preload ParentAlbum association when querying albums with token credentials
  - Added preloading for both root shared album and subalbum queries

  Frontend changes:
  - Modified SHARE_ALBUM_QUERY to include parentAlbum data
  - Reused BreadcrumbList component from AlbumTitle for consistent styling
  - Added sharedRootAlbumId prop to prevent navigation above shared root
  - Breadcrumb only shows when in a subalbum (not at shared root level)
  - Regenerated TypeScript types from GraphQL schema

The current UI when you are in a shared sub-album:
image

The UI after the feature has been implemented:
image

The related feature request is #1366 

## Summary by CodeRabbit

* **New Features**
  * Shared album pages now include a breadcrumb path limited to the shared root, letting users navigate parent albums within a share using the share token and selected root.
  * Breadcrumb data is threaded through gallery and title components and supports custom link generation for breadcrumb items; shared breadcrumbs override local fetching when provided.

* **Tests**
  * Test coverage updated to include shared-root breadcrumb behavior and query variables for shared albums.

### DIFF
--- a/api/graphql/generated.go
+++ b/api/graphql/generated.go
@@ -59,16 +59,17 @@ type DirectiveRoot struct {
 
 type ComplexityRoot struct {
 	Album struct {
-		FilePath    func(childComplexity int) int
-		ID          func(childComplexity int) int
-		Media       func(childComplexity int, order *models.Ordering, paginate *models.Pagination, onlyFavorites *bool) int
-		Owner       func(childComplexity int) int
-		ParentAlbum func(childComplexity int) int
-		Path        func(childComplexity int) int
-		Shares      func(childComplexity int) int
-		SubAlbums   func(childComplexity int, order *models.Ordering, paginate *models.Pagination) int
-		Thumbnail   func(childComplexity int) int
-		Title       func(childComplexity int) int
+		FilePath     func(childComplexity int) int
+		ID           func(childComplexity int) int
+		Media        func(childComplexity int, order *models.Ordering, paginate *models.Pagination, onlyFavorites *bool) int
+		Owner        func(childComplexity int) int
+		ParentAlbum  func(childComplexity int) int
+		Path         func(childComplexity int) int
+		PathForShare func(childComplexity int, token string, rootAlbumID int) int
+		Shares       func(childComplexity int) int
+		SubAlbums    func(childComplexity int, order *models.Ordering, paginate *models.Pagination) int
+		Thumbnail    func(childComplexity int) int
+		Title        func(childComplexity int) int
 	}
 
 	AuthorizeResult struct {
@@ -285,6 +286,7 @@ type AlbumResolver interface {
 
 	Thumbnail(ctx context.Context, obj *models.Album) (*models.Media, error)
 	Path(ctx context.Context, obj *models.Album) ([]*models.Album, error)
+	PathForShare(ctx context.Context, obj *models.Album, token string, rootAlbumID int) ([]*models.Album, error)
 	Shares(ctx context.Context, obj *models.Album) ([]*models.ShareToken, error)
 }
 type FaceGroupResolver interface {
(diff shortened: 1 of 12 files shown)


kkovaletp:
@jdevora, thanks for the PR. Actually, I had this idea in my backlog in my forked repo: https://github.com/kkovaletp/photoview/issues/755
And I asked the bot to prepare an implementation plan for it. So, as you've implemented it faster, please take a look at that suggestion and compare it with your implementation: do you see some useful things? Do we really need all those API changes? Did this plan inspire you to improve the implementation?

jdevora:
> @jdevora, thanks for the PR. Actually, I had this idea in my backlog in my forked repo: [kkovaletp#755](https://github.com/kkovaletp/photoview/issues/755) And I asked the bot to prepare an implementation plan for it. So, as you've implemented it faster, please take a look at that suggestion and compare it with your implementation: do you see some useful things? Do we really need all those API changes? Did this plan inspire you to improve the implementation?

  @kkovaletp, thanks for reviewing! I checked issue #755 and compared the implementations.

  The key difference: Issue #755 suggests using the `path `field (full hierarchy), but that field doesn't work for shared albums because the Path resolver returns an empty array when there's no authenticated user in context 
 Looking at the backend code `(api/graphql/resolvers/album.go:78-86)`, the Path resolver has this logic:

  ```
func (r *albumResolver) Path(ctx context.Context, obj *models.Album) ([]*models.Album, error) {
      user := auth.UserFromContext(ctx)
      if user == nil {                    // ← Share tokens don't set a user!
          empty := make([]*models.Album, 0)
          return empty, nil               // ← Returns EMPTY array
      }
      return actions.AlbumPath(r.DB(ctx), user, obj)
  }
```

  My implementation uses `parentAlbum `with `Preload`, which works immediately with share tokens and provides the core functionality: navigating back through parent albums.

  Why the API changes are needed: The bot's implementation plan may have overlooked that share token authentication doesn't populate a user context, so the existing path resolver wouldn't return any data. My approach required backend changes to preload the parent album relationship specifically for token-based queries.

  Trade-off: The current implementation only shows the immediate parent (one level up) instead of the full path hierarchy, making it inconsistent with the authenticated user experience. I could extend this to use path for full consistency and show the complete breadcrumb hierarchy, but that would require:
  - Modifying the Path resolver to support share tokens
  - Adding logic to limit the path at the shared root album
  - More complexity in the backend

  The current approach prioritizes simplicity and provides functional navigation. However, I'm happy to implement the full path approach if you think the consistency and complete hierarchy are worth the additional backend complexity!

kkovaletp:
@jdevora, thanks for your analysis. 
I see your point about minimal valuable change, but having just 1 parent album in the breadcrumbs doesn't add much value, as there is the Back button, doing the same functionality right now. The main power of breadcrumbs is the ability to get back on several layers with 1 click.
If the existing Path resolver is so limited and focused on the user context, please investigate an alternative implementation of a Path resolver for shares (which might or might not have a token, but for sure won't have a user). Or, maybe it is better to refactor the existing Path code to be more generic and work with shares and the main app, depending on the request.

jdevora:
> @jdevora, thanks for your analysis. I see your point about minimal valuable change, but having just 1 parent album in the breadcrumbs doesn't add much value....
Thanks for the feedback! You're right that showing only one parent level doesn't add much value.

  I considered two approaches:

  1. New pathForShare field - dedicated resolver for share tokens, keeps the existing path untouched
  2. Refactor existing Path resolver - modify to handle both auth contexts, more invasive

  I went with Option 1 for lower regression risk and cleaner separation of concerns. The implementation:

  - Adds pathForShare(token: String!, rootAlbumId: ID!) that reuses the existing GetParents() recursive query
  - Validates the token and limits the path to the shared root
  - Extended AlbumTitle and AlbumGallery to accept the breadcrumb data as props, so share pages use the exact same layout as authenticated pages

kkovaletp:
@jdevora, thanks!
I like the current implementation on the architectural level. I did a quick review and don't see major issues so far.
Please cover the changes with unit tests, where it makes sense.

@googollee, please review the backend part. There are not so many changes, so it shouldn't take much time (I guess)

googollee (commented):
The backend part looks good. It would be better with some tests.

PR #1376, opened 2026-01-28
Added some error handling

KarlGoodMan:
@kkovaletp I've added some error handling to the logic. Regarding the frontend tests, I have limited experience with this area, so any feedback would be appreciated.

Also, I noticed the oldExpireDate field in Sharing.tsx, which is fetched from the database. I believe it needs to be truncated or reformatted before being displayed on the frontend; otherwise, its display might be affected by timezone offsets

## Summary by CodeRabbit

* **New Features**
  * Share tokens with passwords now enforce strict credential validation

* **Bug Fixes**
  * Improved error handling and state recovery when toggling share expiration dates in the sharing interface

* **Documentation**
  * Minor grammar correction in share token mutation descriptions

### DIFF
--- a/api/graphql/resolvers/share_token.go
+++ b/api/graphql/resolvers/share_token.go
@@ -98,6 +98,10 @@ func (r *queryResolver) ShareToken(ctx context.Context, credentials models.Share
 	}
 
 	if token.Password != nil {
+		if credentials.Password == nil {
+			return nil, errors.New("unauthorized")
+		}
+
 		if err := bcrypt.CompareHashAndPassword([]byte(*token.Password), []byte(*credentials.Password)); err != nil {
 			if err == bcrypt.ErrMismatchedHashAndPassword {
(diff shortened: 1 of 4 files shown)


kkovaletp:
> I noticed the oldExpireDate field in Sharing.tsx, which is fetched from the database. I believe it needs to be truncated or reformatted before being displayed on the frontend; otherwise, its display might be affected by timezone offsets

@KarlGoodMan, nice catch. I noticed that once I set the expiration date, it shows me another day (my server and client are in the UTC+2 timezone), but I had no time to investigate it. Probably this is the cause.
Please check it and consistently fix in addition to these fixes)

Regarding the tests, I usually ask the bot to analyze the code and suggest consistent tests to cover it, then iteratively fix the issues until those tests and the code work fine. You can do the same, but as better prompts you'll write, as better tests it will generate for you - check my PR for inspiration/examples if you want.

PR #1384, opened 2026-03-01
scanner: support configurable extension skips for read-only mounts

jonaslewin:
## Summary
- add `PHOTOVIEW_SCANNER_SKIP_EXTENSIONS` to skip known non-media extensions before MIME detection
- keep the feature documented in `README.md` and `api/example.env`
- only suppress `DirEntry.Info()` failures when they are permission-related, while still surfacing unexpected scan errors
- add scanner tests for extension parsing and permission-vs-unexpected file info errors

## Why
Some mounted photo roots also contain large non-media datasets such as GPS logs. When the mount is read-only, `.photoviewignore` is not always an option. This adds a cheap global filter so the scanner can avoid unnecessary MIME detection work on those files while keeping the PR focused on the scan failure problem.

## Testing
- `go test ./scanner/scanner_cache -run TestShouldSkipMediaPath`

## Summary by CodeRabbit

* **New Features**
  * Added scanner filtering capability to skip non-media file extensions during media detection via the PHOTOVIEW_SCANNER_SKIP_EXTENSIONS environment variable.

* **Bug Fixes**
  * Improved error handling for permission issues when scanning media files; now logs and continues gracefully.

* **Documentation**
  * Added configuration guidance for scanner filtering in README and example environment file.

* **Tests**
  * Added comprehensive test coverage for media path filtering and permission error scenarios.

### DIFF
--- a/api/scanner/scanner_album.go
+++ b/api/scanner/scanner_album.go
@@ -17,22 +17,30 @@ import (
 	"gorm.io/gorm"
 )
 
-func albumMediaInfo(ctx scanner_task.TaskContext, entry os.DirEntry, mediaPath string) (fs.FileInfo, bool) {
+func shouldSkipMediaInfoError(err error) bool {
+	return os.IsPermission(err)
+}
+
+func albumMediaInfo(ctx scanner_task.TaskContext, entry os.DirEntry, mediaPath string) (fs.FileInfo, bool, error) {
 	if ctx.GetCache().ShouldSkipMediaPath(mediaPath) {
-		return nil, false
+		return nil, false, nil
 	}
 
 	itemInfo, err := entry.Info()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


kkovaletp:
@jonaslewin, thanks for the contribution, but I'm not sure that we need to add this workaround.

> When the mount is read-only, .photoviewignore is not always an option.

Why do you think so? Can you please describe the exact case, when it is not an option, and why?
As far as I understand, the `.photoviewignore` is the main way of configuring exceptions. If it doesn't work for some cases, I'd expect a bug there and vote for fixing it instead of implementing alternatives and adding unneeded complexity to the product.

jonaslewin:
Thanks for your comment. I understand now after some more research that I can mount a `.photoviewignore` that has this functionality.

PR #1398, opened 2026-03-29
Add Darktable build scripts in `dependencies` image.

googollee:
Part of #1400 , and also for using XMP to generate JPEG previews.

To build a `photoview` image in test branches, it requires `darktable` in the `dependencies` image first. This PR builds `darktable` into the `dependencies` image. Since the main `Dockerfile` isn't changed, it won't copy `darktable` to the `photoview`.

Darktable requires libheif, so this PR changes the build steps.

Before:

- libraw
- libheif
- imagemagick

Now:

- libheif
- libraw
- imagemagick
- darktable (with libheif)

Darktable repo pins a special version of libraw as a submodule. See [their doc](https://github.com/darktable-org/darktable?tab=readme-ov-file#get-submodules). It doesn't need to build libraw outside `darktable`.

## Summary by CodeRabbit

* **New Features**
  * Images now include Darktable binaries with optional version pinning and automatic release detection.

* **Chores**
  * Added a native Darktable build path and integrated its artifacts into final images.
  * Reordered and consolidated build stages to improve cross-build consistency and caching.
  * CI now detects and forwards Darktable version into the build.

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -24,7 +24,7 @@ jobs:
   build:
     name: Build Dependencies Image
     runs-on: ubuntu-latest
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'dependencies_image') }}
+    if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'dependencies_image') }}
 
     steps:
       - name: Checkout


dependencies/build_darktable.sh
```diff
@@ -0,0 +1,97 @@
...
+
+# Fallback to the latest version if DARKTABLE_VERSION is not set
+if [[ -z "${DARKTABLE_VERSION}" ]]; then
+  echo "WARN: Darktable version is empty, most likely the script runs not on CI."
+  echo "Fetching the latest version from Darktable repo..."
+  DARKTABLE_VERSION=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
+    "https://api.github.com/repos/darktable-org/darktable/releases/latest" | jq -r '.tag_name')
+fi
+
+set -euo pipefail
```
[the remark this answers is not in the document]
googollee: Just follow the same pattern in other build scripts. Tend not to change it.

dependencies/build_darktable.sh
```diff
@@ -0,0 +1,97 @@
...
+  libpugixml-dev:"${DEB_HOST_ARCH}" \
+  librsvg2-dev:"${DEB_HOST_ARCH}" \
+  libsqlite3-dev:"${DEB_HOST_ARCH}" \
+  libtiff-dev:"${DEB_HOST_ARCH}" \
+  libwebp-dev:"${DEB_HOST_ARCH}"
+
+URL="https://github.com/darktable-org/darktable.git"
+echo download Darktable repo from "$URL"
+git clone --depth 1 --branch ${DARKTABLE_VERSION} --recurse-submodules --shallow-submodules ${URL} darktable || true
+cd darktable
```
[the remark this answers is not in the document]
googollee: It's useful when developing the script, since `git clone` fails if the repo already exists.

In a clean environment, `git clone` failure will cause `cd darktable` failure. It's fine to keep developing easy here.

dependencies/Dockerfile
```diff
@@ -47,6 +47,20 @@ RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=imagemagick-${TARGETARCH}-${
...
+ENV BUILD_CACHE_DIR=/build-cache
+WORKDIR /tmp/build-darktable
+COPY --chmod=0755 build_darktable.sh .
+RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=imagemagick-${TARGETARCH}-${DARKTABLE_VERSION} \
+    --mount=type=secret,id=github_token \
+    if [ -f /run/secrets/github_token ]; then \
+    export GITHUB_TOKEN=$(cat /run/secrets/github_token); \
+    fi \
+    && set -a && source /env && set +a \
+    && ./build_darktable.sh
```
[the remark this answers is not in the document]
googollee: Fixed.

dependencies/build_darktable.sh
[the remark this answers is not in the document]
googollee: `rm` first slows the build. If a build succeeds, then it should be fine.

kkovaletp:
LGTM so far, but the build fails, so I'll wait till the final version of the script

googollee:
It's much harder than I thought to fix this cross compiling problem. I'm working on it now.

googollee:
It's very hard to do cross-compile for darktable. It's building script doesn't accept any Debian compile environments.

I'm trying with native building with qemu arm64 environment now.

kkovaletp:
How much would we get from the self-built Darktable CLI compared to the one we had in 2.4.0?
Maybe it's not worth the effort...

googollee:
We removed `darktable` because suse repo depends on a lot of Debian repo for libraries, like `libheif` and `libraw`. Since Debian stable won't catch up new versions in the library, we were  missing new features for `heif` and `raw`, which are pretty critical for `photoview`. And that's why we are moving to self-build `imagemagick`.

Debian trixie was cut on  August 9th, 2025, and current libraries are somehow newish. But if we go back to what `2.4.0` did, we may face that problem again in the future.

dependencies/build_darktable.sh
```diff
@@ -0,0 +1,103 @@
...
+  libwebp-dev
+
+URL="https://github.com/darktable-org/darktable.git"
+echo download Darktable repo from "$URL"
+git clone --depth 1 --single-branch --branch ${DARKTABLE_VERSION} ${URL} darktable || true
+cd darktable
+git config submodule.src/tests/integration.update none
+git submodule update --init --recursive --depth 1 --recommend-shallow --single-branch
+
+cp /app/debian-cross.toolchain.cmake .
```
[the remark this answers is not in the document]
googollee: Done.

dependencies/build_darktable.sh
```diff
@@ -0,0 +1,101 @@
...
+  iso-codes \
+  xsltproc \
+  zlib1g \
+  libavif-dev \
+  libcurl4-openssl-dev \
+  libexiv2-dev \
+  libgmic-dev \
+  libgraphicsmagick1-dev \
+  libgtk-3-dev \
+  libheif-dev \
```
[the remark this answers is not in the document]
googollee: Yes, that's the goal. The priority is to ensure that `darktable` can be compiled successfully. When switching to `darktable`, I'll consider how to modify `libheif` part.

dependencies/build_darktable.sh
```diff
@@ -0,0 +1,101 @@
...
+: "${DEB_HOST_GNU_TYPE:=$(dpkg-architecture -a "$DEB_HOST_ARCH" -qDEB_HOST_GNU_TYPE)}"
+CACHE_DIR="${BUILD_CACHE_DIR:-/build-cache}/Darktable-${DARKTABLE_VERSION}"
+CACHE_MARKER="${CACHE_DIR}/Darktable-${DARKTABLE_VERSION}-complete"
+
+# Check if this specific version is already built and cached
+if [[ -f "$CACHE_MARKER" ]] && [[ -d "${CACHE_DIR}/output" ]]; then
+  echo "Darktable ${DARKTABLE_VERSION} found in cache, reusing..."
+  mkdir -p /output
+  cp -ra "${CACHE_DIR}/output/"* /output/
+  exit 0
```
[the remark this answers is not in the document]
googollee: It's the same pattern as other build scripts. In CI, it should be always a clean environment without any cache. In dev environment, developer should take the response.

dependencies/build_darktable.sh
```diff
@@ -0,0 +1,101 @@
...
+  --enable-opencl \
+  --enable-openexr \
+  --enable-openmp \
+  --enable-webp"
+
+./build.sh \
+  --prefix "/opt/darktable" \
+  --build-type "Release" \
+  --install \
+  ${FEATURES} \
```
kkovaletp: @googollee, the last `\` is redundant. Not a big issue, just a forgotten thing)

dependencies/build_darktable.sh
```diff
@@ -0,0 +1,101 @@
...
+
+mkdir -p /output/opt
+cp -a /opt /output/
```
kkovaletp: @googollee, I know that we build in a clean Docker image, and the /opt folder will contain only the new `darktable` sub-folder, but it is better to specify it explicitly, because the /opt folder is a system folder, so there is a small chance that it might contain some other content in future versions of the base image.

dependencies/Dockerfile
```diff
@@ -57,10 +57,32 @@ RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=ffmpeg-${TARGETARCH}-${JELLY
...
+ENV BUILD_CACHE_DIR=/build-cache
+WORKDIR /tmp/build-darktable
+COPY --chmod=0755 build_darktable.sh .
+RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=darktable-${TARGETARCH}-${DARKTABLE_VERSION} \
+    --mount=type=secret,id=github_token \
+    if [ -f /run/secrets/github_token ]; then \
+    export GITHUB_TOKEN=$(cat /run/secrets/github_token); \
+    fi \
+    && set -a && source /env && set +a \
+    && ./build_darktable.sh
```
kkovaletp, line 84: @googollee, so you don't use the libheif and libraw, built by us earlier. Why? I thought that it was 1 of the main intentions for self-building it...

kkovaletp:
@googollee, I'm reading the Docker image build log for x64 and ARM image, and I see:
- some pre-compilation tests failed
- a list of optional not-found modules
- a list of files, referenced in the build config, but not found
- LibRAW is compiled with many things set to NO
- Darktable is built with many locales, while we need only the EN one
- ARM compilation log contains some errors, like:
```
In file included from /usr/include/c++/14/algorithm:61,
from /usr/include/exiv2/types.hpp:13,
from /tmp/build-darktable/darktable/src/common/exif.cc:23:
/usr/include/c++/14/bits/stl_algo.h: In instantiation of 'constexpr std::pair<_FIter, _FIter> std::minmax(initializer_list<_Tp>) [with _Tp = float]':
/tmp/build-darktable/darktable/src/common/exif.cc:5658:38: required from here
5658 | const auto x_minmax = std::minmax({ points[idx],
| ~~~~~~~~~~~^~~~~~~~~~~~~~~
5659 | points[idx + 2],
| ~~~~~~~~~~~~~~~~
5660 | points[idx + 4],
| ~~~~~~~~~~~~~~~~
5661 | points[idx + 6] });
| ~~~~~~~~~~~~~~~~~~
/usr/include/c++/14/bits/stl_algo.h:3435:5: note: parameter passing for argument of type 'std::pair<float, float>' when C++17 is enabled changed to match C++14 in GCC 10.1
3435 | minmax(initializer_list<_Tp> __l)
| ^~~~~~
```
- the build runs for 1 hour 43 mins. This is a huge increase. In case of a cache miss, we'll need to wait for 2 hours per job run (for instance, during PR commits)

How big is the final image?
Please review these findings and see if you can fix/satisfy some of them, and make the build size/time smaller

dependencies/build_darktable.sh
googollee: Fixed.

dependencies/build_darktable.sh
googollee: Sure, fixed.

dependencies/Dockerfile
googollee: I just updated the script to use self-built libheif.

Darktable pins a libraw submodule in its repo. They update libraw when necessary. Since darktable repo is very active, I'll depend on them to do the compatible thing and update libraw.

googollee:
I didn't enable the darktable integration tests in the build. We grab the source with the stable version, and it's not necessary to run tests again.

Most of the not-found (if not all of them) are related to devices and display, which we won't use. For the locales, there is no way to disable them.

The message you pasted here is "note", not an error. I think it comes from the difference between gcc and clang. Let darktable take care of them.

`LibRAW is compiled with many things set to NO`. `darktable` maintains the feature. For example, darktable links to rawspeed directly, and it doesn't enable it in libraw.

Yes, it takes much longer to compile. Part of it comes from the huge dependencies, and the main part comes from emulating an `arm64` environment when building `darktable`. They don't provide a way for cross-compiling, and I can't find the way either. Debian uses native compiling for `arm64`, so do I here.

.github/workflows/dependencies.yml
kkovaletp: @googollee, with this long build time, I think that we need to modify the job run triggers:
- remove the `push` and `pull_request` triggers
- set the `schedule` to run nightly at the least busy time (check our other schedules and assume 2.5 hours build time, and don't schedule it at the beginning of an hour)
- add the manual trigger, which allows running the job manually at any time. Check the GitHub CI docs - I don't remember its name and syntax.

dependencies/Dockerfile
```diff
@@ -1,39 +1,39 @@
...
+FROM build-libheif AS build-libraw
+ARG LIBRAW_VERSION
 ARG TARGETARCH
 ENV BUILD_CACHE_DIR=/build-cache
-WORKDIR /tmp/build-libheif
-COPY --chmod=0755 build_libheif.sh .
-RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=libheif-${TARGETARCH}-${LIBHEIF_VERSION} \
+WORKDIR /tmp/build-libraw
+COPY --chmod=0755 build_libraw.sh .
+RUN --mount=type=cache,target=${BUILD_CACHE_DIR},id=libraw-${TARGETARCH}-${LIBRAW_VERSION} \
```
kkovaletp, line 28: @googollee, do we still need this order switch? Does it give us any benefits?
If not, let's revert these changes to limit the number of changes in the PR.

kkovaletp: Ah, I see. It makes sense because `libheif` is then used in the `darktable` build, so it is better to compile it earlier.
Resolving this comment)

dependencies/build_darktable.sh
```diff
@@ -0,0 +1,100 @@
...
+
+mkdir -p /output/opt
+cp -a /opt/darktable /output/opt/darktable
+file /output/opt/darktable/bin/darktable-cli
+
+# After successful build, cache the results
+echo "Caching Darktable ${DARKTABLE_VERSION} build results..."
+mkdir -p "${CACHE_DIR}/output"
+cp -ra /output/* "${CACHE_DIR}/output/"
+touch "$CACHE_MARKER"
```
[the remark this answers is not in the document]
googollee: no need to add the complexity here. Just cache with version.

googollee:
The image size increases from 152MiB to 195MiB. Considering it includes many configurations like lens profiles, I think it's OK. The final size may be reduced a little when removing `imagemagick`.

I can't say the final photoview image. I stil wait on this PR to build the image.

.github/workflows/dependencies.yml
googollee: Updated. Please check.

kkovaletp:
> The image size increases from 152MiB to 195MiB.

That is totally fine. It builds for such a long time, so I thought, there will be GBs...

.github/workflows/dependencies.yml
```diff
@@ -1,13 +1,15 @@
...
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: 'Why trigger building the dependencies?'
+        required: false
+        default: 'Dependencies are changed'
```
kkovaletp, line 11: @googollee, thanks for implementing this trigger.
As far as I understand from the docs and conversation with ChatGPT, it will start working only after merging to `master` and will always build from the `master` branch.
Please read my conversation here: https://chatgpt.com/share/69d0019f-6cd0-8332-98da-008f63963f98

I think that this is a good thing, but for easy checks inside PRs (like this one), in addition to this trigger, it would be useful to add an `issue_comment` trigger with the `/build_deps` (or similar) command, and/or a labeled trigger with creating a special label "build deps image". Then, in case of a comment, we can add such a comment to trigger the build in the PR; and for a label - just assign this label from the list.
As for me, the label looks better, as we don't need to remember the command and just click the label to trigger the build. I guess that we can unlabel and label it many times to trigger it again.

googollee: Updated. Now the dependencies build will be triggered when a PR with the `dependencies` label. It's better than some text in the comment.

kkovaletp: I agree that assigning a label is more intuitive and easier than remembering a special command and writing it in the comment.
1 thing is that the `dependencies` label is used automatically by Dependabot. Also, we use it to indicate that the PR works with project dependencies, so reusing it as a trigger would cause unintended builds of the image. That is why I suggested creating a new dedicated label with a clear name, so we don't reuse it for something else.

googollee: How about a new`dependencies_image` label?

.github/workflows/dependencies.yml
```diff
@@ -20,6 +24,7 @@ jobs:
   build:
     name: Build Dependencies Image
     runs-on: ubuntu-latest
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'dependencies_image') }}
```
kkovaletp: @googollee, I just realized that this condition would work only for the PR runs. Scheduled runs would skip this job, as there is no `event.pull_request` context.
We need to set the following condition to satisfy all possible cases:
```
if: ${{ github.event_name != 'pull_request' || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'dependencies_image')) }}
```
BTW, I feel that this condition can be simplified to the following one, but I'm not sure that this is correct for GitHub, and we cannot test it till merge to `master`:
```
if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'dependencies_image') }}
```

googollee: Thanks for pointing it out. Updated. I think the simplified way should work.

PR #1403, opened 2026-03-31
Migrate MapBox -> MapLibre to remove the API key requirement

### LINKED ISSUE #744: Update Places to use geolocation and / or map centering to orient on markers
When loading the Places tab, I'm dropped into a geo that none of my images are located in.  I need to manually drag the map to an area of photos.

To mirror other Places functionality from other apps, I'd like to see the map for places centered to a bounding area that covers all of my photo locations (markers).  Alternatively, centering using geolocation on first load.

**Additional context**
From the mapbox documentation, it looks like it'll be a combination of using the `center` prop: https://docs.mapbox.com/help/tutorials/custom-markers-gl-js/

Possibly the `fitBoundsOptions` prop as well: https://docs.mapbox.com/mapbox-gl-js/api/markers/

### LINKED ISSUE #789: Please supports Non-English map in Place

### LINKED ISSUE #879: MapLibre instead of MapBox
I'm starting using photoview because I want a self-hosted solution that is free and open-source. Using MapBox doesn't seem to comply with my constraints.

Seems like MapLibre is a viable alternative, from what I understand.

Not sure, I took the first result from StackExchange to propose a solution.

**Additional context**
https://github.com/maplibre/maplibre-gl-js

### LINKED ISSUE #1308: Replace Mapbox with OpenFreeMap
Mapbox is no longer free to use without a credit card.
In the mean time the [OpenFreeMap](https://openfreemap.org/) project has come around, which provides free unlimited vector tiles without any tokens.

Replacing Mapbox with OpenFreeMap would add map support without having to configure any tokens.

OpenFreeMap uses the [MapLibre](https://maplibre.org/maplibre-gl-js/docs/) library for web rendering, of which there also exists a [react binding library](https://visgl.github.io/react-maplibre/), which looks very similar to the Mapbox react library that we already are using.

CommanderStorm:
This has the advantage of removing the licence key requirement, and generally cleans up the code imo.

Here is how the new "design" looks.
I chose a bit of a different style of map, I think it looks good.
The image being a 404 is an unrelated issue, the frontend does not serve the assets, the backend does.

Recording+2026-04-01+at+01 46 27 (1)

(demo looks blurry and choppy because of upload restrictions in github)

CC @louwers

Resolves https://github.com/photoview/photoview/issues/879
Resolves https://github.com/photoview/photoview/issues/789
Resolves https://github.com/photoview/photoview/issues/744
Resolves https://github.com/photoview/photoview/issues/1308

## Summary by CodeRabbit

* **New Features**
  * Replaced Mapbox with MapLibre for maps; improved label localization, RTL support, globe-centering, marker behavior, and cleanup.
  * Admins can configure separate light/dark map style URLs via Settings; map style changes apply in-app.

* **Chores**
  * Added a server toggle to disable all map features (removes Places page, sidebar map, and map style settings); updated example envs and frontend example API endpoint.

* **Tests**
  * Added tests for globe-centering, cluster/marker UI, and places reducer.

* **Documentation**
  * Added Map Feature docs and self-hosting/off‑grid instructions.


ui/src/components/maplibre/MaplibreMap.tsx
```diff
@@ -0,0 +1,186 @@
...
+    ['get', 'name'],
+  ]
+
+  for (const layer of style.layers) {
+    if (layer.type !== 'symbol') continue
+    const tf = (layer.layout as any)?.['text-field']
+    if (!tf) continue
+    map.setLayoutProperty(layer.id, 'text-field', nameExpr)
+  }
+}
```
CommanderStorm, line 79: related:
- https://github.com/osm-americana/diplomat/issues/40

ui/src/components/maplibre/MaplibreMap.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+function getStyleUrl() {
+  return isDarkMode()
+    ? 'https://tiles.openfreemap.org/styles/dark'
+    : 'https://tiles.openfreemap.org/styles/positron'
+}
```
CommanderStorm: this is a stylistic shift, but I think these look more in-line with the rest of the app than mapbox' streets and dark style.

ui/vite.config.js
```diff
@@ -11,6 +11,9 @@ export default defineConfig({
   server: {
     port: 1234,
   },
+  build: {
+    target: 'es2022',
+  },
```
CommanderStorm, line 16: nessresary due to the older node version which implicitly sets the browser target.
-> we need to tell vite that we need to target newer browsers that support `BigInteger`

ui/example.env
```diff
@@ -1 +1 @@
-REACT_APP_API_ENDPOINT=http://localhost:4001/
\ No newline at end of file
+REACT_APP_API_ENDPOINT=http://localhost:4001/api
```
CommanderStorm, line 1: Otherwise the docker file cannot connect to the backend

ui/src/components/maplibre/MaplibreMap.tsx
```diff
@@ -0,0 +1,186 @@
...
+ */
+function localizeLabels(map: maplibregl.Map, locale: string) {
+  const style = map.getStyle()
+  if (!style?.layers) return
+
+  const nameExpr: maplibregl.ExpressionSpecification = [
+    'coalesce',
+    ['get', `name:${locale}`],
+    ['get', 'name'],
+  ]
```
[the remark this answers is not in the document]
CommanderStorm: I know, basically this requires https://github.com/osm-americana/diplomat/issues/40

ui/src/Pages/PlacesPage/MapPresentMarker.tsx
```diff
@@ -36,35 +36,28 @@ const QUERY_MEDIA = gql`
...
-        ?.properties as MediaMarker | undefined
-
-      if (media === undefined) {
-        reject('ERROR: media is undefined')
-        return
-      }
+  if (cluster) {
+    const mediaSource = map.getSource('media') as maplibregl.GeoJSONSource
+    const features = await mediaSource.getClusterLeaves(id as number, 1000, 0)
+    return features.map(feat => feat.properties) as MediaMarker[]
```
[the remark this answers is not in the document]
CommanderStorm: no change in behaviour from the old mapbox version

ui/src/components/maplibre/MaplibreMap.tsx
kkovaletp: I guess Maplibre has a list of themes. Can we group light and dark themes in 2 lists, and depending on the global theme, let the user choose 1 of the Map themes from the corresponding list in the settings?
We can add 1 more dropdown under the Theme preferences section on the Settings page

louwers: MapLibre itself does not distribute styles or tile data, but OpenFreeMap may be nice for this use case because users don't need to get an API key.

[Here](https://github.com/hyperknot/openfreemap-styles) are the styles they offer. Yes there's also a dark version.

https://tiles.openfreemap.org/styles/dark

See here: https://maplibre.org/maputnik/?style=https://tiles.openfreemap.org/styles/dark

kkovaletp: @CommanderStorm, thanks. I used the last URL to check all the styles from the repo. There is only 1 dark theme (which is "dark"), and I don't like it, as it is just a black map with gray labels. It is not practically usable.
Are there alternative styles somewhere else? If not, I'd better not have any dark theme for the map than ship their dark theme.
From light themes, I personally like the liberty one, as it draws the map with more topographic details, which would let users more easily understand the location (especially tourists, making pictures outside of cities and villages)

kkovaletp:
@CommanderStorm, please review the bot's findings in the https://github.com/photoview/photoview/pull/1403#pullrequestreview-4040946361 (and future similar comments) under collapsed blocks. Once done, put a Like/Dislike reaction to the comment, so I know that you've reviewed those comments.

ui/src/components/maplibre/MaplibreMap.tsx
kkovaletp: For the 1st implementation, we can go with 1 light and 1 dark (if you find a good one, because the "dark" theme from the https://tiles.openfreemap.org/styles/dark is not usable) theme and automatically pick one depending on the global app theme.
Adding a dropdown with a list of themes would be a future enhancement.

kkovaletp: Another question:
Can the new Map implementation work in offline mode (with no internet on the server)?
If yes, we need to pull the styles and all other remote resources to the `dist` folder at the project build time, so that the deployed app doesn't need to query remote servers through the internet.
If not, we need to have a feature flag to let users disable the Places page, as it works now. We have users preferring offline deployments and/or strictly hardened environments for privacy.

ui/src/components/maplibre/MaplibreMap.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+      map.current = m
+
+      m.on('style.load', () => {
+        m.setProjection({ type: 'globe' })
+        if (localeRef.current) {
+          localizeLabels(m, localeRef.current)
+        }
+        onStyleLoadRef.current?.(m, maplibre)
+      })
```
kkovaletp, line 158: @CommanderStorm, when the map loads for the 1st time, it renders in the Globe mode and puts the media markers on the map - the same way as it works now with MapBox. Right?
What side of the globe will be shown to the user by default?
With the current MapBox implementation, it will always be the GPS 0:0 point (Africa Atlantic). For users from other continents, this is an annoying UX bug, requiring them to drag and rotate the globe each time to the location of their media markers.
If you can implement the starting point calculation based on the coordinates with the majority of users' media markers, this would be a big improvement)

ui/src/components/maplibre/MaplibreMap.tsx
```diff
@@ -0,0 +1,186 @@
...
+      if (map.current) {
+        const observer = (map.current as any)._themeObserver as
+          | MutationObserver
+          | undefined
+        observer?.disconnect()
+        map.current.remove()
+        map.current = null
+      }
+    }
+  }, [])
```
kkovaletp, line 197: @CommanderStorm, don't we have any dependencies in such a complex hook?
Fine if no, just checking)

ui/vite.config.js
kkovaletp: In my forked repo, I have this setting:
```
esbuild: {
      target: 'es2020', // Ensure compatibility with browsers, not older than from 2021
      logOverride: { 'this-is-undefined-in-esm': 'silent' },
    },
```
It is fine if we need to apply the `es2022` because of MapLibre or your code requirements.
Isn't it better to use the `esbuild` section name?
Is the log override useful here as well?

ui/package.json
kkovaletp, line 1: @coderabbitai, I put this comment here, but it is relevant to all PR changes.
This PR's goal is to consistently migrate the map functionality from MapBox to the MapLibre library.
Please carefully analyze the PR end-to-end in the context of the project, find all inconsistencies, regressions in the UX, app behavior, visual design, functionality, and other aspects, if any. Also, please compare the previous implementation of the map with the capabilities and features of MapLibre, and let us know if there are some MapLibre features/capabilities that are useful for Photoview but not implemented in this PR.
When performing your analysis, please use all your knowledge (but verify your learnings for consistency with the project code before applying them), the code and configuration from this repo, all relevant docs and code from MapLibre and MapBox projects. Verify your analysis results and resolve all assumptions before submitting them as a reply.

kkovaletp: @CommanderStorm, please review these findings as well.
And, please cover new code with unit tests, where it makes sense: https://github.com/photoview/photoview/pull/1403#issuecomment-4186838828

ui/src/components/maplibre/MaplibreMap.tsx
CommanderStorm: > Can the new Map implementation work in offline mode (with no internet on the server)? If yes, we need to pull the styles and all other remote resources to the `dist` folder at the project build time, so that the deployed app doesn't need to query remote servers through the internet

Depends on which kind of infra is acceptable and what level of qualty you want.
Serving data for maps is notiroiusly large, even if cut down you won't get smaller than 2-5GB for planet scale, even if you cut down on size (=features), compress agressively and drop some non-core functionality.

And at that point you won't have many of the niceties of modern maps like properly translated labels/... You will get english labels and highly simplified geometrys.

The "current, full featured map" (= using openfreemap tiles) would take a docker file and 80GB of storage on the offline-mode users devices to download and preprocess OSMs data.
The preprocessing step takes about an 30m on resonable hardware.

If offline-capability is a requirmenet and restricting the zoom an option (think: you can zoom in to the city level, but there won't be more data after this), there is another option, but it also would require downloading data.

How about this? https://docs.protomaps.com/basemaps/flavors#light
Maybe with a lower-zoom version of https://mapterhorn.com/viewer/

Here is how terrain looks like if you drop the offline requirmeent..
https://mapterhorn.com/examples/terrain/

What is the flag for this offline mode or is this just not implemented currently?

louwers: How about we use a PMTiles as a vector tile source... Maybe use one of the Protomaps' styles.

@kkovaletp It's 125GB in this case... Not something you want to bake into a Docker container, but quite doable as a one-time deployment step.

https://pmtiles.io/archive/#url=https%3A%2F%2Fdemo-bucket.protomaps.com%2Fv4.pmtiles

We could allow overriding the default (online) style(s) with an environment variable?

CommanderStorm: As I said above, can be cut down significantly, but that looses features like translation.. not sure if this is desirable.

In terms of data size, a good cut is not having zoom levels or features => So if we only go down to a certain Zoom.. -> https://pmtiles.io/#url=https%3A%2F%2Fbuild.protomaps.com%2F20260404.pmtiles&map=10.92/48.2591/11.6622&showTileBoundaries=true

To give a better intuition:
- maps use a quad, so every zoom level you zoom in is 4x the data tiles
- every tile is roghly the same size, except the last (because you want to allow overzooming)
- so Z0..15 -> 125GB, Z0..14 -> 80GB, ...
- If I for example simplify down quite a lot and do deep cuts, I can get things down to ~10-20GB

ui/src/components/maplibre/MaplibreMap.tsx
CommanderStorm: No, the empty dep array is correct by design. All props that could change are stored in refs and read inside the effect/callbacks, so the effect intentionally runs only on mount/unmount.
This avoids expensive map re-initialization when props change.

Map re-inaitilastion would look like a flash, so not desirable

ui/package.json
CommanderStorm: - 1. was real
- 4./9. kind of by design/needs discussion above, 
- 5. is not something that I would usually add until someone complains, but sure
- 6. pre existing, don't know what that is talking about
- 8. does not really add value here.. why would you want a ScaleControl for your picture map?

rest is bogus or halucinated

ui/src/components/maplibre/MaplibreMap.tsx
CommanderStorm: not as simple because "majority of users' media markers" is not as simple of a problem.

I implemented a solution which should solve this, I think.
@louwers can you look over and see if you know of a simpler way than the two way iteration using the vector math?
I somewhat came up with that on the whiteboard and it works well according to the values I tested, but maybe there is a simpler way that I cannot see.

ui/src/components/maplibre/globeCenter.ts
```diff
@@ -0,0 +1,178 @@
...
+  // Compute centroid of only the points visible from the chosen center,
+  // which excludes outliers on the opposite hemisphere
+  const visible = vecs.filter(p => dot(candidate, p) > Number.EPSILON)
+  const centroid =
+    visible.length > 0 ? getSphericalCentroid(visible) : candidate
+
+  // Compute zoom from the center to ALL points (outliers widen the view)
+  const center = toCoord(centroid)
+  const centerVec = toVec(center)
+  const zoom = computeZoom(centerVec, vecs)
```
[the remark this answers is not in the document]
CommanderStorm: no, just plain out wrong. I added a testcase to prove you wrong

ui/src/components/maplibre/MaplibreMap.tsx
kkovaletp: @CommanderStorm, @louwers, thanks for the explanation and ideas on how to make it work offline, but adding GBs to the app size is not an option at all. I'd think about a ew hundreds of MBs, but not more.

So, let's implement a feature flag, enabled by default with the "REACT_APP_MAP_ENABLED" env var (default `true`, set to `false` to disable), to disable it completely. Please see how other UI env vars are declared and used across the code: https://github.com/photoview/photoview/blob/master/ui/src/%40types/index.d.ts

Please add it to the `example.env` instead of the removed MapBox token and route through both docker-compose examples.

ui/package.json
```diff
@@ -24,13 +24,13 @@
   },
   "dependencies": {
     "@apollo/client": "^3.6.9",
+    "@mapbox/mapbox-gl-rtl-text": "^0.3.0",
```
kkovaletp: @coderabbitai, is there a better alternative to this RTL text plugin from the MapLibre ecosystem?
I'm asking because in this PR, we migrate from MapBox, but still need to use its RTL plugin, connecting to the map from the MapLibre ecosystem. This looks strange and might cause compatibility issues now or in the future.

@CommanderStorm FYI

ui/vite.config.js
kkovaletp: @CommanderStorm, any response?)

ui/src/components/maplibre/MaplibreMap.tsx
kkovaletp: I know, I've mixed several topics in 1 conversation, and now it is messy to keep them all together. I'll start another conversation thread to continue the discussion about the themes.

kkovaletp: @CommanderStorm, this is the fork of the https://github.com/photoview/photoview/pull/1403#discussion_r3019043030 conversation to continue the initial topic discussion about the Map themes.

My last comment in that parent thread about this topic was https://github.com/photoview/photoview/pull/1403/changes/BASE..ef55eba589fd4ae4afc26601e55d03359837a86f#r3035318585. Please answer it with your thoughts/vision.

CommanderStorm: I got from your coomment here https://github.com/photoview/photoview/pull/1403#discussion_r3035313894 that you would prefer an "outdoor-ish" style and "free" as in no api key.

Unfortunately, well tuned styles in this area have only been written in the company realm:
- for the terrain -> https://www.maptiler.com/maps/#style=topo-v4&lang=auto&mode=3d&position=13.16/47.03808/11.07388/0.00/60.0
- https://docs.protomaps.com/basemaps/flavors#light

The latter one is self hostable, but not really compatible with what openfreemap offers, at least not without a lot of tinkering.

Here is how 3d terrain using another "free" (funded by nlnet/hetzner/cloudflare) service looks like
https://mapterhorn.com/examples/terrain/

kkovaletp: @CommanderStorm, thanks for the more examples.
Actually, I'd prefer a balanced style with some basic details of nature, as well as detailed cities. The https://maplibre.org/maputnik/?style=https://tiles.openfreemap.org/styles/liberty is a good one. Maybe it would be better if it had some basic representation of heights (lakes vs. mountains). It would be great if this style had a dark/night mode, but if not, and you cannot find anything similar in the dark style, it would be better to go with 1 theme unconditionally. As I mentioned earlier, the earlier shared dark style is not practically usable.

The examples from your last reply look too detailed and work slowly on my Mac. I guess, because of the heavy load of the terrain details.

Sure, we can use only styles that are free for non-commercial usage and compatible with our open-source license. Integrating non-free for open-source styles (as well as any other dependencies) is not an option for us.

CommanderStorm: Okay, if terrain is too heavy, how about just hillshading and transitioning to the natural earth if we move out?



```diff
{
  "version": 8,
  "sources": {
    "ne2_shaded": {
      "maxzoom": 6,
      "tileSize": 256,
      "tiles": [
        "https://tiles.openfreemap.org/natural_earth/ne2sr/{z}/{x}/{y}.png"
      ],
      "type": "raster"
    },
    "openmaptiles": {
      "type": "vector",
      "url": "https://tiles.openfreemap.org/planet"
-     }
+     },
+     "hillshadeSource": {
+       "type": "raster-dem",
+       "url": "https://tiles.mapterhorn.com/tilejson.json"
+     }
  },
  "sprite": "https://tiles.openfreemap.org/sprites/ofm_f384/ofm",
  "glyphs": "https://tiles.openfreemap.org/fonts/{fontstack}/{range}.pbf",
  "layers": [
    {
      "id": "background",
      "type": "background",
      "paint": {
        "background-color": "#f8f4f0"
      }
    },
+     {
+       "id": "hills",
+       "type": "hillshade",
+       "source": "hillshadeSource",
+       "paint": {
+       "hillshade-method": "igor",
+       "hillshade-illumination-direction": 315,
+       "hillshade-shadow-color": "#000000",
+       "hillshade-highlight-color": [
+         "rgba(255, 227, 227, 1)"
+       ],
+       "hillshade-accent-color": "rgba(255, 159, 0, 1)",
+       "hillshade-exaggeration": 0.15,
+       "hillshade-illumination-altitude": [
+         10
+       ]
+     }
+   },
    {
      "id": "natural_earth",
      "type": "raster",
      "source": "ne2_shaded",
      "maxzoom": 7,
      "paint": {
        "raster-opacity": [
          "interpolate",
          ["exponential", 1.5],
          ["zoom"],
          0,
          0.4,
          6,
          0
        ]
      }
    },
````

Screenshot_20260405_191341

File: ui/vite.config.js (hunk 3 above)
Comment (CommanderStorm) [reply ui/vite.config.js:16]: If we want the RTL plugin to work as it does, 2022 it is.
dynamic imports via `?url` are only avaliable in es2022..

File: ui/src/components/maplibre/MaplibreMap.tsx (hunk 2 above)
Comment (kkovaletp) [reply ui/src/components/maplibre/MaplibreMap.tsx]: @CommanderStorm, this looks good, thanks!

**Comment by CommanderStorm** [comment]
@kkovaletp could you approve CI to run again? I think this should pass CI, but this is a bit harder since my environment is a bit newer than CI, so..

**Comment by CommanderStorm** [comment]
I dropped the RTL feature, that seems to not entirely be compatible with the fairly old vite version used to build this project, at least it would require the CDN fallback.. likely not acceptable.

File: package.json
Comment (kkovaletp) [inline package.json:1]: @CommanderStorm, this file is redundant. The `ui/package.json` is used.

**Comment by kkovaletp** [comment]
> I dropped the RTL feature, that seems to not entirely be compatible with the fairly old vite version used to build this project, at least it would require the CDN fallback.. likely not acceptable.

Unfortunately, we have had no UI maintainer in the project for years. Nobody can review PRs (including mine), so I cannot contribute to the UI part of the main repo.

That's why I've done many and keep implementing fixes and version bumps for the UI in my forked repo, shipping it as an external Docker image, which anyone can run together with the main Photoview image, serving backend. Details are in the https://discord.gg/MgQhXd32 and in my repo: https://github.com/kkovaletp/photoview/blob/master/ui/Separate-UI-README.md

**Comment by kkovaletp** [comment]
> my environment is a bit newer than CI, so..

You can use the https://github.com/photoview/photoview/blob/master/dev-compose.yaml as an example to set up your dev environment based on Docker. It will then be consistent with the project's dependencies and environment

**Comment by kkovaletp** [comment]
@CommanderStorm, any progress so far? Did you stuck with something, I can help you with?

**Comment by CommanderStorm** [comment]
Sorry, I have not had the time to fix the remaining issues.
Will try to get this ready for review over the weekend

File: ui/package.json [hunk 11]
```diff
@@ -47,7 +46,7 @@
     "i18next": "^21.8.13",
     "js-cookie": "^3.0.5",
     "luxon": "^3.7.2",
-    "mapbox-gl": "^2.9.1",
+    "maplibre-gl": "^5.21.1",
```
birkskyum, line 49: @CommanderStorm Would be great to move to maplibre-gl v6

PR #1404, opened 2026-04-06
Cross build darktable.

googollee:
Export PKG_CONFIG and let `darktable` build with cross-build files. And other optimizations.

## Summary by CodeRabbit

* **Chores**
  * Introduced a streamlined darktable build stage and removed a redundant native build step to simplify image assembly.
  * Centralized build cache environment to reduce repetition across stages.
  * Tighter, architecture-qualified dependency installation for more predictable cross-builds.
  * Explicit toolchain/configuration selection for consistent multi-arch artifacts.
  * Expanded cache restore to repopulate runtime directories and refresh library caches for faster rebuilds.
  * Minor configure/ordering tweaks to stabilize build feature handling.

### DIFF
--- a/dependencies/build_imagemagick.sh
+++ b/dependencies/build_imagemagick.sh
@@ -54,11 +54,11 @@ cd ImageMagick-*
 
 FEATURES="--with-heic --with-jpeg --with-png --with-raw --with-tiff --with-webp"
 
+pkg-config --list-all
 ./configure \
   --enable-64bit-channel-masks \
   --enable-static --enable-shared --enable-delegate-build \
-  --without-x --without-magick-plus-plus \
-  --without-perl --disable-doc \
+  --without-x --without-magick-plus-plus --without-perl \
   --host="${DEB_HOST_GNU_TYPE}" \
   ${FEATURES}
 
--- a/dependencies/build_imagemagick.sh
+++ b/dependencies/build_imagemagick.sh
@@ -54,7 +54,6 @@ cd ImageMagick-*
 
 FEATURES="--with-heic --with-jpeg --with-png --with-raw --with-tiff --with-webp"
 
-pkg-config --list-all
 ./configure \
   --enable-64bit-channel-masks \
   --enable-static --enable-shared --enable-delegate-build \
--- a/dependencies/build_libheif.sh
+++ b/dependencies/build_libheif.sh
@@ -15,18 +15,6 @@ set -euo pipefail
 CACHE_DIR="${BUILD_CACHE_DIR:-/build-cache}/libheif-${LIBHEIF_VERSION}"
 CACHE_MARKER="${CACHE_DIR}/libheif-${LIBHEIF_VERSION}-complete"
 
-# Check if this specific version is already built and cached
-if [[ -f "$CACHE_MARKER" ]] && [[ -d "${CACHE_DIR}/output" ]]; then
-  echo "libheif ${LIBHEIF_VERSION} found in cache, reusing..."
-  mkdir -p /output
-  cp -ra "${CACHE_DIR}/output/"* /output/
-  exit 0
-fi
-
-echo "Building libheif ${LIBHEIF_VERSION} (cache miss)..."
-
-echo Compiler: "${DEB_HOST_GNU_TYPE}" Arch: "${DEB_HOST_ARCH}"
-
 apt-get install -y \
   libdav1d-dev:"${DEB_HOST_ARCH}" \
   libde265-dev:"${DEB_HOST_ARCH}" \
@@ -37,6 +25,23 @@ apt-get install -y \
   libnuma-dev:"${DEB_HOST_ARCH}" \
   zlib1g-dev:"${DEB_HOST_ARCH}"
 
+# Check if this specific version is already built and cached
+if [[ -f "$CACHE_MARKER" ]] && [[ -d "${CACHE_DIR}/output" ]]; then
+  echo "libheif ${LIBHEIF_VERSION} found in cache, reusing..."
+  mkdir -p /output /usr/local/{bin,lib,lib/pkgconfig,include}
+  cp -ra "${CACHE_DIR}/output/"* /output/
+  cp -ra /output/bin/* /usr/local/bin/
+  cp -ra /output/lib/* /usr/local/lib/
+  cp -ra /output/pkgconfig/* /usr/local/lib/pkgconfig/
+  cp -ra /output/include/* /usr/local/include/
+  ldconfig
+  exit 0
+fi
+
+echo "Building libheif ${LIBHEIF_VERSION} (cache miss)..."
+
+echo Compiler: "${DEB_HOST_GNU_TYPE}" Arch: "${DEB_HOST_ARCH}"
+
 URL="https://api.github.com/repos/strukturag/libheif/tarball/${LIBHEIF_VERSION}"
 echo download libheif from "$URL"
 curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 -o ./libheif.tar.gz \
@@ -45,8 +50,7 @@ curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 -o ./libheif.tar.gz \
 tar xfv ./libheif.tar.gz
 cd ./*-libheif-*
 cmake \
-  --preset=release-noplugins \
-  -DENABLE_PLUGIN_LOADING=OFF \
+  --preset=release \
   -DWITH_GDK_PIXBUF=OFF \
   -DCMAKE_SYSTEM_PROCESSOR="${DEB_HOST_ARCH}" \
   -DCMAKE_C_COMPILER="${DEB_HOST_GNU_TYPE}"-gcc \
--- a/dependencies/build_libheif.sh
+++ b/dependencies/build_libheif.sh
@@ -45,7 +45,7 @@ curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 -o ./libheif.tar.gz \
 tar xfv ./libheif.tar.gz
 cd ./*-libheif-*
 cmake \
-  --preset=release \
+  --preset=release-noplugins \
   -DENABLE_PLUGIN_LOADING=OFF \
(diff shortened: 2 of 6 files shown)


dependencies/build_libheif.sh
```diff
@@ -45,13 +46,13 @@ tar xfv ./libheif.tar.gz
 cd ./*-libheif-*
 cmake \
   --preset=release \
+  -DENABLE_PLUGIN_LOADING=OFF \
+  -DWITH_GDK_PIXBUF=OFF \
```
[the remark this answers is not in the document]
googollee: Fixed.

googollee:
The new build is less than 1 hour. @kkovaletp should we revoke the dependence image build conditions?

googollee:
@kkovaletp I fixed the problem that imagemagick can't be built when libheif/libraw are cached. The `build-libheif` needs to copy so/include/pc files into system directory `/usr/local/lib` and install dependencies. Otherwise `build-imagemagick` can't find those dependencies and build fails. Same as `build-libraw`.

dependencies/build_imagemagick.sh
```diff
@@ -54,11 +54,11 @@ cd ImageMagick-*
 
 FEATURES="--with-heic --with-jpeg --with-png --with-raw --with-tiff --with-webp"
 
+pkg-config --list-all
 ./configure \
   --enable-64bit-channel-masks \
   --enable-static --enable-shared --enable-delegate-build \
-  --without-x --without-magick-plus-plus \
-  --without-perl --disable-doc \
+  --without-x --without-magick-plus-plus --without-perl \
```
kkovaletp, line 60: Why not `--disable-doc` anymore?

dependencies/build_libheif.sh
```diff
@@ -15,27 +15,33 @@ set -euo pipefail
...
 
+apt-get install -y \
+  libdav1d-dev:"${DEB_HOST_ARCH}" \
+  libde265-dev:"${DEB_HOST_ARCH}" \
+  libjpeg62-turbo-dev:"${DEB_HOST_ARCH}" \
+  libopenh264-dev:"${DEB_HOST_ARCH}" \
+  libopenjp2-7-dev:"${DEB_HOST_ARCH}" \
+  libpng-dev:"${DEB_HOST_ARCH}" \
+  libnuma-dev:"${DEB_HOST_ARCH}" \
+  zlib1g-dev:"${DEB_HOST_ARCH}"
```
kkovaletp, line 26: Why do we need to spend time installing these packages for the cached dependency case? If they are needed for later scripts, those scripts can install them directly.

dependencies/build_libraw.sh
```diff
@@ -15,23 +15,28 @@ set -euo pipefail
 CACHE_DIR="${BUILD_CACHE_DIR:-/build-cache}/LibRaw-${LIBRAW_VERSION}"
 CACHE_MARKER="${CACHE_DIR}/LibRaw-${LIBRAW_VERSION}-complete"
 
+apt-get install -y \
+  libjpeg62-turbo-dev:"${DEB_HOST_ARCH}" \
+  liblcms2-dev:"${DEB_HOST_ARCH}" \
+  zlib1g-dev:"${DEB_HOST_ARCH}"
```
kkovaletp, line 21: same question. Writing just to highlight this block for future updates

kkovaletp:
> The new build is less than 1 hour. @kkovaletp should we revoke the dependence image build conditions?

Still too long.

dependencies/build_imagemagick.sh
googollee: There is no `--disable-doc` in the latest version.

dependencies/build_libheif.sh
googollee: They are required by next stages. `imagemagick` depends on `libheif`, and it also depends on those libraries to build. It's not a good idea to install them again in the `imagemagick` stage, since `imagemagick` are from `libheif`.

dependencies/build_libraw.sh
googollee: Same as `libheif`.

dependencies/build_libheif.sh
kkovaletp: it is not "again" as we have the caching condition. 
When there is an option to skip some steps in a chain, each step (after ensuring that it has to run) should install all its dependencies.
Also, `apt install` checks if the requested package is already installed and skips it if it is.

So, in the chain of LibHeif > LibRAW > ImageMagick, I see the right way of installing deps, like the following:
1. LibHeif:
- check if cached
- if not, install all needed deps except ones installed for sure in the prep. stage
- build
- cache + output
2. LibRAW:
- check if cached
- if not, install all needed deps (including those, potentially installed by LibHeif) except ones installed for sure in the prep. stage
- build
- cache + output
3. ImageMagick:
- check if cached
- if not, install all needed deps (including those, potentially installed by both prev. stages) except ones installed for sure in the prep. stage
- build
- cache + output

Then, if some/both libs are cached, ImageMagick will install all needed packs by itself. If each step builds from scratch (cache miss), then the repeated packs will be installed by the 1st caller and skipped by subsequent ones. If all 3 steps are cached, then no packs are installed, and the scripts finish fast.

googollee: Not really. In general, a cached stage should produce the same stage as non-cached.

When non-cached, `libraw`/`libheif` will add below things in the image:

- `apt install` packages in the system directories
- self-building packages in the system directories
- `/output` files.

The mounted cache keeping all things from `/output`. Without this change, when the cache hits, `libraw`/`libheif` will add below things in the image:

- `/output` files

`imagemagick` image will miss all packages from `apt install` and self-buliding packages when `libraw`/`libheif` hits mounting-cache, so it fails building. 

Moving `apt install` before cache-checking ensures all needed packages installed with and without caching mount.

`imagemagick`/`darktable` don't need to change since they are nodes of the building chains.

googollee: Do you mean to repeat all packages of `apt install` from `libraw`/`libheif` in `imagemagick`? I don't think it's necessary because it should always based on `libraw` and `libheif` for self building packages.

I tend to keep the stage the same with and without mounting-cache. It's not good to have a surprise when cache happening suddenly.

kkovaletp: not all. only those required by each stage, without assuming them already installed earlier (except those installed on the prep. stage)

With the current implementation, there will be noticeable time spent on `apt install` for libs even if they are already in the cache. I want to avoid this by running the `apt install` only after ensuring there is a cache miss case. I understood the idea behind your change: if 1/both libs are cached, but ImageMagick is not, the corresponding APT packs were not installed in the earlier image stages, and ImageMagick fails. That is why I propose to run `apt install` for all needed packages (except those installed on the prep. stage): if there was an active build of lib(s), they are installed, and APT will skip them; if not, it will install them before building ImageMagick.

googollee: If so, when we add new dependent stages for `imagemagick`, we need to update `apt install` in the `imagemagick` stage. Since `imagemagick` is built from `libraw`/`libheif`, it's not necessary because previous stages should be self-complete.

googollee: It's not good to have a surprise of changing stages when a cache happens suddenly.

googollee:
@kkovaletp I can't find where you mentioned that installing `darktable` from obs. Based on my investigation, I believe this is feasible. Here are some potential changes:

1. Drop self-build `libheif`

We introduced self-build `libheif` with https://github.com/photoview/photoview/issues/1062. It was in Debian 12 bookworm. Now in Debian 13 Trixie, `libheif` was upgraded to 1.19.8, and it provides a group of `libheif-plugins-*`. We can install those plugins from Debian now.

2. Use OBS Darktable

OBS `darktable-cli` doesn't link to `libraw.so`. That means `darktable` [turns off](https://github.com/darktable-org/darktable/blob/cae87d8810505692105302f7484bcc1da1de8ef6/src/external/LibRaw-cmake/CMakeLists.txt#L83) the share library of `libraw` and [static links](https://github.com/darktable-org/darktable/blob/cae87d8810505692105302f7484bcc1da1de8ef6/src/external/LibRaw-cmake/CMakeLists.txt#L469) `libraw` in its binary. You can verify it by `ldd /usr/bin/darktable-cli`. I also ensure it doesn't depend on system `libraw` library after installing OBS `darktable`, on both x64 and arm64.

3. Drop `dependencies` image

We can download `darktable` from OBS (like 2.4.0) and `ffmpeg` from `jellyfin` in the main Dockerfile, after `darktable-cli` replacing `imagemagick`. Without `imagemagick`, there is no necessary to build `libraw` since `darktable` self-contains `libraw`. It reduces the complexity of bulid. If so, I'll drop this PR and do it after `darktable-cli` PR finishing.

What do you think?

kkovaletp:
@googollee, we had a conversation about the previous way of installing Darktable, which we have in 2.4.0. This conversation was in the initial PR for adding self-built Darktable to our deps image. I'm not sure what "Darktable from OBS " is, but if it eliminates/covers all risks/issues we had in 2.4.0, I'm fine with downloading it from there instead of building it ourselves.

I don't like the idea of completely replacing ImageMagick with Darktable. Our last agreement was that you would develop a PoC and measure the overall scan job time, comparing it with the ImageMagick reference for the same media collection/server/DB. I still expect a significant scan time increase for Darktable, and if that will be the case, I'd suggest having both image processors integrated, call ImageMagick by default, and switch to Darktable only for the specific cases, like RAW + XMP. We can define the list of suitable cases later.

So, for now, please keep the deps image and the self-built dependencies pipeline.

googollee:
In 2.4.0, it downloads from OBS (Open Build Service).

I prefer correctness and simplicity over performance, and I don't want to maintain two implementations of one feature. I insist on removing `ImageMagick` when `darktable` is ready. If our users aren't reporting performance issues widely with 2.4.0,  no need to optimize it.

kkovaletp:
I'll take a break in this discussion to react not too emotionally.
Meanwhile, you can search user messages in our Discord and issues related to scanning large media libraries, if you want to.

googollee:
Closed in favor of #1405.

Issue #1400, opened 2026-03-30
Generate preview jpg and web jpg applying edits in xmp files.


googollee:
Lightroom can store raw edits in xmp (or in its private storage, which we don’t discuss here). Those edits won’t store in the raw files, to avoid changing them.  If we want to show images the same as what Lightroom shows after editing, we need to generate preview jpg and web jpg applying the edits in xmp.

Imagemagick can’t parse xmp files. `darktable-cli` may be the only way to do that, since it’s a widely-used, open-source tool to replace Lightroom.

GlassedSilver:
Are Lightroom edits fully compatible with Darktable?

googollee:
Since darktable mentions Lightroom in [their sidecar XMP doc](https://docs.darktable.org/usermanual/3.8/en/overview/sidecar-files/sidecar-import/), I tend to believe they try their best to make XMP compatible, even they never promise it.

googollee:
With some test, importing images in Darktable GUI applies the Lightroom XMP file, and it generates a new Darktable XMP file.

Meanwhile, running `darktable-cli <Input RAW file> <Input XMP file> <Output JPEG file>  --core --configdir /tmp/darktable --conf "plugins/lighttable/xmp_lookup=1" --conf "metadata/sidecar_priority=1" --conf "plugins/lighttable/xmp_text_fulltext=1"` can apply Darktable XMP file, but not Lightroom XMP file.

I'm still working on the way of applying the Lightroom XMP file with `darktable-cli`.

googollee:
I put this idea on hold now.

After reading a lot of darktable code, I'm pretty sure that `darktable-cli` can't apply edits (or develops in darktable code context) from Lightroom XMP files. `darktable-cli` calls [`dt_lightroom_import` with `dev=NULL`](https://github.com/darktable-org/darktable/blob/06f2ded73e3fe55d891c1f67941109b5281ea75f/src/common/image.c#L2042), which means it only reads tags from Lightroom XMP files, but [no develops](https://github.com/darktable-org/darktable/blob/06f2ded73e3fe55d891c1f67941109b5281ea75f/src/develop/lightroom.h#L23). darktable only reads develops from Lightroom XMP files when it runs as GUI.

I believe they have good reasons for doing this, but I don't know. I may ask about it in darktable repo if I finally decide to do that.

Before finding a way to apply develops in Lightroom XMP files, I don't think it's worth continuing.

PR #1407, opened 2026-04-10
Rollback commit 946ddd8: Add Darktable build scripts in `dependencies` image. (#1398)

googollee:
See https://github.com/photoview/photoview/issues/1400#issuecomment-4226087175

## Summary by CodeRabbit

* **Chores**
  * Updated CI workflow schedules and trigger conditions.
  * Restricted automated pull-request builds to the main branch.
  * Reordered dependency build stages and streamlined build pipelines.
  * Removed a native library/component and its associated build script from the automated build process.


kkovaletp (commented):
@googollee, basically, this PR is fine. I just noticed that you don't remove the build script for Darktable. Is it on purpose or just escaped diff?
I guess it is safe to drop it, as we have the git branch history in GitHub, which stores even deleted branches, so all the changes for this commit will be accessible in the history when needed.

googollee:
The script should be removed. Updated.

kkovaletp (approved):
LGTM

Issue #1205, opened 2025-05-16
iphone 16 .heic files lead to decoding error when added to photoview


skiwlkr:
**Describe the bug**
Decoding error when uploading a HEIF image taken with the latest iphone 16. The image shows up in photoview, but only a white rectangle is displayed. Photos made with a iphone 13 in .heic format work without any issues. It seems like iphone 16 .heics have different metadata

Errormessages:
info box in photoview frontent: ould not get media by media_id and user_id from database: record not found"
photoview log: " image decoding: failed to decode HEIF image (/photos/vier/adfkas.heic): Invalid input: Unspecified: Metadata not correctly assigned 5/1 to image1"

**To Reproduce**
1. upload .heic image into the photo library
2. see if it's decoded correclty and showing up in frontend

**Expected behavior**
uploading a heic image and photoview is decoding it correclty and it's showing up in the photoview frontend.

**Screenshots**

<img width="495" alt="Image" src="https://github.com/user-attachments/assets/b15f260d-c1b0-4775-b56a-57d5d018d60b" />

**Your environment:**
photoview runs on a synology docker enviroment

 - Are you using Docker? [yes]
 - Database [e.g. postgres 1.6]
 - Browser [e.g. chrome]
 - Version [e.g. 2.4.0.  master ([a1aadf67]]

**Additional context**

I tried decoding the sample image from this [thread](https://github.com/photoview/photoview/issues/1062) but it didn't work although im running the master a1aadf67 and libheif seems to be on version 1.18

kkovaletp:
@skiwlkr, thanks for reporting the issue.
@googollee is working on switching to ImageMagick completely and dropping our own thumbnails generation code in the PR #1204.
Can you please share your iPhone 16 HEIF image example, which triggers that metadata error? We'll test our new implementation on this photo to make sure that it works fine.

googollee:
I'll add the file in this https://github.com/photoview/photoview/issues/1062 and check.

googollee:
Added the file in the test, see https://github.com/googollee/photoview/commit/32ff4e8174e0b8855aa9af62dca1554250989c9f. It works well.

@skiwlkr Could you provide the problem file, if it's possible?

skiwlkr:
Thanks for the quick reply. 
I've attachted the heic file. It is from a iphone 16 SE

[IMG_5960.heic.zip](https://github.com/user-attachments/files/20263882/IMG_5960.heic.zip)

googollee:
[...]

PR #1408, opened 2026-04-11
Remove self-build `libheif` in dependencies and cleanup build scripts.

googollee:
- Remove self-built `libheif` in dependencies
- Remove build-cache in dependencies
  - The cache avoids rebuilding packages when changing building options in scripts. I provide `build.sh` to enable caching layers by package versions.
  - Another case that cache fails:
    - Clean all caches
    - Update `build_imagemagick.sh` but has some errors.
      - For example, add `exit -1` in the top of the script.
    - Build the image and it fails.
    - Fix `build_imagemagick.sh`
      - Remove `exit -1` from the script.
    - Mean while, update `lib_raw.sh` to avoid loading docker layer cache.
      - Add an empty line on the top.
    - Build the image and it keeps failing, which it should not.
      - Can't enable `libraw` and `libheif`.
  - Let me know what the initial idea was when introducing this build-cache.
- Move compressing ui files in the UI building stage. Reduce the size of the final image.
- Force enabling more formats when building ImageMagick
- To keep backward compatibility, the dependencies of `libheif` are still installed in `install_{build,runtime}_dependencies.sh`. They could be removed after generating a new dependencies image.

Test image from offical build workflow: `ghcr.io/googollee/photoview:cleanup_build`

Test image from self-build without any self-built `libheif` and dependencies: `googollee/photoview:clean_script`

## Summary by CodeRabbit

* **New Features**
  * Added a centralized dependency-version resolver that emits resolved dependency tags.

* **Chores**
  * Reworked container build pipeline with an intermediate dependency stage and reordered asset compression/copying.
  * Removed legacy build stages and cache-reuse paths; simplified artifact propagation.
  * Hardened build/test/CI scripts with stricter failure handling and consolidated package installs.
  * Slimmed runtime images by avoiding recommended packages and trimming install steps.

* **Style**
  * Expanded .dockerignore to exclude additional test/build artifacts.


scripts/install_runtime_dependencies.sh
```diff
@@ -1,17 +1,65 @@
...
+  exit 0
+fi
+
 apt-get update
-apt-get install -y --no-install-recommends curl libimage-exiftool-perl
 
-# libheif dependencies
-apt-get install -y --no-install-recommends libdav1d7 librav1e0.7 libde265-0 libx265-215 libjpeg62-turbo libopenh264-8 libpng16-16t64 libnuma1 zlib1g
+# exiftool
+apt-get install -y --no-install-recommends libimage-exiftool-perl
```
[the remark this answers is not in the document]
googollee: All `install_xx.sh` should be run after `set_compiler_env.sh`.

dependencies/build.sh
```diff
@@ -0,0 +1,30 @@
...
+if [[ $# = "1" ]] && [[ -n "$1" ]]; then
+  BUILD_ARGS="-t $1"
+fi
+
+docker build \
+  --build-arg "LIBRAW_VERSION=${LIBRAW_VERSION}" \
+  --build-arg "IMAGEMAGICK_VERSION=${IMAGEMAGICK_VERSION}" \
+  --build-arg "JELLYFIN_FFMPEG_VERSION=${JELLYFIN_FFMPEG_VERSION}" \
+  ${BUILD_ARGS} \
+  .
```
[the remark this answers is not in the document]
googollee: Updated.

scripts/install_runtime_dependencies.sh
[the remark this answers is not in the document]
googollee: The release stage doesn't compile anything. It copies files from build stage and won't call curl. The release stage needs to keep the size, and it's better not installing curl in release stage.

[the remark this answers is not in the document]
googollee: Fixed.

dependencies/build_imagemagick.sh
```diff
@@ -52,20 +48,43 @@ curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 -o ./magick.tar.gz \
...
 
 ./configure \
   --enable-64bit-channel-masks \
-  --enable-static --enable-shared --enable-delegate-build \
-  --without-x --without-magick-plus-plus \
-  --without-perl --disable-doc \
+  --enable-static \
+  --enable-shared \
+  --enable-delegate-build \
+  ${FEATURES[@]} \
```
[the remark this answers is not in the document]
googollee: Fixed.

kkovaletp:
Hi @googollee, you used the PR description to ask some questions and clarify some aspects, so I'll reply to those lines with my comments, answers, and questions:

> * Remove self-built `libheif` in dependencies

If I remember correctly, when we added a self-built `libheif`, it was too old in the Debian registry. Now we use the newer image, so the package is newer as well, but eventually it might get outdated in the same way. I don't remember any major cases caused by this self-built lib so far. What benefits do we get from removing it from the build flow and replacing it with the Debian package?

> * Remove build-cache in dependencies

If I understand the changes, you want to remove the file-level cache for built packages. Not the workflow's Docker layers cache.

> * The cache avoids rebuilding packages when changing building options in scripts. I provide `build.sh` to enable caching layers by package versions.

This is strange that it fails this way... BTW, I didn't find where the `build.sh` is called in the pipeline. Is it just for local manual testing?

> * Let me know what the initial idea was when introducing this build-cache.

The main idea behind this file-level cache was to cover the case when only 1 of our dependencies is updated, while all other ones are still on their latest versions. With only Docker layer caching, the updated dependency and all subsequent layers are always rebuilt, while with this file-level cache, the latest versions, built earlier, are copied from the cache and don't spend the build time. So, the total job run time is not so long because of unnecessary rebuilds for previously built dependencies.

BTW, at some point in time, I've simplified the Docker layer cache config in my forked repo (https://github.com/kkovaletp/photoview/blob/master/.github/workflows/dependencies.yml), after which the job runs even faster: https://github.com/kkovaletp/photoview/actions/workflows/dependencies.yml.
I run A LOT of iterations for workflow execution in my repo during UI dependencies updates and my PR development, so the longest job execution time is crucial for me. I didn't push this change to the main repo, because we have much fewer job runs there and it is not so critical to wait ~20 - 30 min in the worst case. However, as you touch this area, please take a look and check if it makes sense to implement the same.

> * Move compressing ui files in the UI building stage. Reduce the size of the final image.

This, unfortunately, breaks the setting version commit workaround:
- The commit hash is unique per push
- The commit hash is provided to the Docker build context as a build arg
- A Docker build arg invalidates the cache for the current layer and all subsequent layers in the image build sequence
- That is why the commit hash should be imported and used as late in the Dockerfile as possible, allowing all earlier layers to be cached and reused if no other changes exist
- So, when we build the UI, we insert the constant "-=<GitHub-CI-commit-sha-placeholder>=-" instead of the real commit hash
- The line `RUN find /app/ui/assets -type f -name "SettingsPage.*.js" -exec sed -i 's/"-=<GitHub-CI-commit-sha-placeholder>=-"/"'"${COMMIT_SHA}"'"/g' {} \;` finds this constant in the compiled Settings page asset and replaces it with the real commit hash
- Only after this manipulation, we can compress our assets for the final web server usage, because with your changes, the compressed version of the Settings page asset still contains the "-=<GitHub-CI-commit-sha-placeholder>=-".

> * To keep backward compatibility, the dependencies of `libheif` are still installed

What exactly do you expect to be broken now if the currently unused dependencies are removed? The main image takes only FFmpeg and ImageMagick from this image. Everything else should be internal, and any changes should be fine while these 2 outputs are available in a healthy condition. Do I miss something?

googollee:
> Remove self-built libheif in dependencies.

The original issue is #1205. The old Debian didn't contain any libheif plugins, but the new one has. As the test, `libheif` already contains necessary plugins and it works well with the image in the issue. Since HEIF is a common and basis format and it's already widely used for years, like JPEG/TIFF/PNG, I tend to rely on the official library.

> you want to remove the file-level cache for built packages. Not the workflow's Docker layers cache.

Yes.

> I didn't find where the build.sh is called in the pipeline.

Because I don't know the use case. Putting versions in workflow will reduce the build time, like what you did in your repo. I can update it in the workflow.

> The main idea behind this file-level cache was to cover the case when only 1 of our dependencies is updated, while all other ones are still on their latest versions.

In theory, if `libraw` is changed, `imagemagick` should be rebuilt. For example, some const varibles may change in the libraw HEAD files and they should be updated in `imagemagick`. Even when the library maintains the compatible, it's better to re-build because no ABI declares in Linux area.

> What exactly do you expect to be broken now if the currently unused dependencies are removed?

The build/test workflow in PR will fail, because `imagemagick` requires `libheif`  in current `photoview/dependencies` image and it can't be run without `libheif` dependencies.

> Only after this manipulation, we can compress our assets for the final web server usage

OK. I'll roll this part back. It's not a big size comparing to the total image size.

googollee:
Now there is `version.sh` to output library versions, instead of `build.sh`.

Rolled back the compression of assets.

scripts/install_runtime_dependencies.sh
kkovaletp, line 1: @googollee, here and in the "scripts/install_build_dependencies.sh", you run the `apt-get install` command many times for different sets of packages. From the functional point of view, it works fine, but each execution of `apt-get` does the same (pretty long) APT registry refresh and other preparations. This is unnecessary when you run them one after another, because the OS state is exactly the same.
To save some execution time, I'd propose the following changes for both scripts:
```bash
...
apt-get update

# The 1st group of packages
apt-get install -y --no-install-recommends \
pkg1 \
pkg2 \
# The 2nd group
pkg3 \
pkg4 \
# etc.
last-pkg
```
With this implementation, the APT runs only once, prepares for the installation only once, and installs all requested packages in the optimal order, while the script is still separated by logical groups and can be maintained easily.

Dockerfile
```diff
@@ -52,14 +52,14 @@ ENV PATH="${GOPATH}/bin:${PATH}"
...
-    && set -a && source /env && set +a \
-    && /app/scripts/install_build_dependencies.sh \
+COPY --chmod=0755 scripts/set_compiler_env.sh /app/scripts/
+RUN source /app/scripts/set_compiler_env.sh
+
+COPY --chmod=0755 scripts/install_build_dependencies.sh /app/scripts/
+RUN set -a && source /env && set +a \
+    && /app/scripts/install_build_dependencies.sh
+COPY --chmod=0755 scripts/install_runtime_dependencies.sh /app/scripts/
+RUN set -a && source /env && set +a \
```
kkovaletp, line 69: @googollee, this is not a big issue, as this is an intermediate stage, but each separate COPY and RUN creates a new layer, which is a bit longer and uses a bit more space.
There is a best practice to combine similar COPY and RUN statements in one corresponding COPY and RUN.

Dockerfile
```diff
@@ -102,19 +102,21 @@ ARG TARGETPLATFORM
 # See for details: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
-COPY scripts/install_runtime_dependencies.sh /app/scripts/
+# Create a user to run Photoview server
+RUN groupadd -g 999 photoview \
+    && useradd -r -u 999 -g photoview -m photoview
```
kkovaletp, line 112: @googollee, it is better to keep user creation in the same big RUN to avoid unnecessary layer creation

kkovaletp:
> what you did in your repo. I can update it in the workflow.

I found the PR in my repo, where I implemented the CI build cache improvements: https://github.com/kkovaletp/photoview/pull/836/changes

Basically, I simplified the Docker layer cache names by removing keys (as they are unneeded and make it even worse) and clearly defined what build job writes caches, and what reads them only, to make sure that there is no race condition in writing the same cache during parallel job execution for the same commit.

Maybe I did other changes earlier and don't remember them now...

scripts/install_runtime_dependencies.sh
googollee: That comment doesn't work in Bash. I don't have a good way to comment with different groups in one command. If there is the way, I'm happy to update.

Both below don't work properly:

```bash
apt-get \
  # comment
  pkg1 \
  pkg2
```


```bash
apt-get \
  # comment \
  pkg1 \
  pkg2
```

Dockerfile
googollee: It's better for cache when developing with scripts. Since it's an intermediate stage, I tend to provide better cache here.

Dockerfile
googollee: It's better for cache. Since the final image is ~1GiB, one more layer is fine.

Well, groupadd/useradd is fast. If you insist, I will put them back.

googollee:
> I found the PR in my repo.

This pr differs a lot with master workflow. Since I didn't work on this direction now, could you try to PR your changes?

scripts/install_runtime_dependencies.sh
googollee: With some search, I found a way to do it. Please take a look.

Dockerfile
kkovaletp: Nah, if you're aware of it and still see benefits in the current implementation, I'm ok with it as well.

scripts/install_runtime_dependencies.sh
kkovaletp: Nice and elegant)

kkovaletp:
> could you try to PR your changes?

Sure, after this PR is merged.

> Test image from offical build workflow: `ghcr.io/googollee/photoview:cleanup_build`
> Test image from self-build without any self-built `libheif` and dependencies: `googollee/photoview:clean_script`

What is the deps image from the latest commit in this PR?
Did you test building the main Photoview image using the deps image from this PR? Are all dependencies installed and working well? To test it in the CI, we need to merge this PR 1st and rerun the main image build job on top of the new deps image, which would be too late if something is wrong and deps are broken in some way...

Can you please fix the Docker build job in this PR? I want to see how the main image builds with the latest changes in the scripts and Dockerfile in this PR.

googollee:
`ghcr.io/googollee/photoview:cleanup_build` is the image built with this PR main code and the official `photoview/dependencies` image.

`googollee/photoview:clean_script` is the image built with this PR main code and the dependencies image built from this PR too.

> Can you please fix the Docker build job in this PR?

What is the build problem of this PR? Current CI fails because I enable pushing the image to my repo, for generating `ghcr.io/googollee/photoview:cleanup_build`.

kkovaletp:
> `googollee/photoview:clean_script` is the image built with this PR main code and the dependencies image built from this PR too

Can I see the build job log? Better in the .dockerbuild format, so I can import it to my Docker Desktop for easy analysis)

> What is the build problem of this PR? Current CI fails because I enable pushing the image to my repo, for generating `ghcr.io/googollee/photoview:cleanup_build`

https://github.com/photoview/photoview/actions/runs/24359967292/job/71136781071?pr=1408
Docker login fails, and Docker Build is skipped, so no image is built at all. I want to see the build log for the deps image in this PR.

googollee:
> Docker login fails, and Docker Build is skipped, so no image is built at all. I want to see the build log for the deps image in this PR.

Do you still remember [this flag](https://github.com/photoview/photoview/pull/1408/changes/BASE..ff87a27bb291e5b43ea750b4f6d86462d97bdc9f#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721)? It enables building an image in the personal repo. CI build workflow with PR won't create any image by default. It will be rolled back when the review finishes.

You can check the actions in my repo: https://github.com/googollee/photoview/actions?query=branch%3Acleanup_build.

> Can I see the build job log? 

I don't use Docker Desktop, so it's hard to generate a `.dockerbuild` file. The command I built `googollee/photoview:clean_script` is below:

```
$ cd dependencies/                                                                                                                                           
$ docker build -t photoview/dependencies:trixie .                                                                 
[+] Building 1.1s (23/23) FINISHED                                                                  docker:default
 => [internal] load build definition from Dockerfile                                                          0.1s
 => => transferring dockerfile: 1.62kB                                                                        0.0s
 => [internal] load metadata for docker.io/library/debian:trixie-slim                                         0.0s
 => [internal] load .dockerignore                                                                             0.1s
 => => transferring context: 2B                                                                               0.0s
 => [base-prep 1/3] FROM docker.io/library/debian:trixie-slim                                                 0.0s
 => [internal] load build context                                                                             0.1s
 => => transferring context: 182B                                                                             0.0s
 => CACHED [base-prep 2/3] COPY --chmod=0755 prepare.sh /tmp/                                                 0.0s
 => CACHED [base-prep 3/3] RUN /tmp/prepare.sh     && mkdir -p /output/bin /output/lib /output/include /outp  0.0s
 => CACHED [build-libraw 1/3] WORKDIR /tmp/build-libraw                                                       0.0s
 => CACHED [build-libraw 2/3] COPY --chmod=0755 build_libraw.sh .                                             0.0s
 => CACHED [build-libraw 3/3] RUN --mount=type=secret,id=github_token     if [ -f /run/secrets/github_token   0.0s
 => CACHED [final-assembly 1/5] COPY --from=build-libraw /output/ /output/                                    0.0s
 => CACHED [build-imagemagick 1/3] WORKDIR /tmp/build-imagemagick                                             0.0s
 => CACHED [build-imagemagick 2/3] COPY --chmod=0755 build_imagemagick.sh .                                   0.0s
 => CACHED [build-imagemagick 3/3] RUN --mount=type=secret,id=github_token     if [ -f /run/secrets/github_t  0.0s
 => CACHED [final-assembly 2/5] COPY --from=build-imagemagick /output/ /output/                               0.0s
 => CACHED [download-jellyfin-ffmpeg 1/3] WORKDIR /tmp/download-ffmpeg                                        0.0s
 => CACHED [download-jellyfin-ffmpeg 2/3] COPY --chmod=0755 download_jellyfin-ffmpeg.sh .                     0.0s
 => CACHED [download-jellyfin-ffmpeg 3/3] RUN set -a && source /env && set +a     && ./download_jellyfin-ffm  0.0s
 => CACHED [final-assembly 3/5] COPY --from=download-jellyfin-ffmpeg /output/ /output/                        0.0s
 => CACHED [final-assembly 4/5] COPY --chmod=0755 output.sh /                                                 0.0s
 => CACHED [final-assembly 5/5] RUN /output.sh                                                                0.0s
 => CACHED [release 2/2] COPY --from=final-assembly /artifacts.tar.gz /                                       0.0s
 => exporting to image                                                                                        0.1s
 => => exporting layers                                                                                       0.0s
 => => writing image sha256:c3ded45dd7ccd3a8f0c3e4f8bdda0e215c2a6c0e4cad195f0a53676d7a921f5f                  0.0s
 => => naming to docker.io/photoview/dependencies:trixie                                                      0.0s
$ cd ..                                                  
$ docker build -t googollee/photoview:clean_script .
[+] Building 1.6s (40/40) FINISHED                                                                  docker:default
 => [internal] load build definition from Dockerfile                                                          0.1s
 => => transferring dockerfile: 6.26kB                                                                        0.0s
 => [internal] load metadata for docker.io/library/node:18                                                    0.5s
 => [internal] load metadata for docker.io/photoview/dependencies:trixie                                      0.0s
 => [internal] load metadata for docker.io/library/golang:1.26-trixie                                         0.5s
 => [internal] load metadata for docker.io/library/debian:trixie-slim                                         0.0s
 => [internal] load .dockerignore                                                                             0.0s
 => => transferring context: 339B                                                                             0.0s
 => [api  1/15] FROM docker.io/library/golang:1.26-trixie@sha256:c0074c718b473f3827043f86532c4c0ff537e3fe7a8  0.1s
 => => resolve docker.io/library/golang:1.26-trixie@sha256:c0074c718b473f3827043f86532c4c0ff537e3fe7a81b8219  0.1s
 => FROM docker.io/photoview/dependencies:trixie                                                              0.0s
 => [internal] load build context                                                                             0.1s
 => => transferring context: 48.64kB                                                                          0.0s
 => [release  1/10] FROM docker.io/library/debian:trixie-slim                                                 0.0s
 => [ui 1/6] FROM docker.io/library/node:18@sha256:c6ae79e38498325db67193d391e6ec1d224d96c693a8a4d9434985567  0.0s
 => CACHED [release  2/10] RUN groupadd -g 999 photoview     && useradd -r -u 999 -g photoview -m photoview   0.0s
 => CACHED [release  3/10] COPY --chmod=0755 scripts/install_runtime_dependencies.sh /app/scripts/            0.0s
 => CACHED [release  4/10] WORKDIR /dependencies                                                              0.0s
 => CACHED [api  2/15] COPY --chmod=0755 scripts/set_compiler_env.sh /app/scripts/                            0.0s
 => CACHED [api  3/15] RUN source /app/scripts/set_compiler_env.sh                                            0.0s
 => CACHED [api  4/15] COPY --chmod=0755 scripts/install_build_dependencies.sh /app/scripts/                  0.0s
 => CACHED [api  5/15] RUN set -a && source /env && set +a     && /app/scripts/install_build_dependencies.sh  0.0s 
 => CACHED [api  6/15] COPY --chmod=0755 scripts/install_runtime_dependencies.sh /app/scripts/                0.0s 
 => CACHED [api  7/15] RUN set -a && source /env && set +a     && /app/scripts/install_runtime_dependencies.  0.0s 
 => CACHED [api  8/15] COPY --from=photoview/dependencies:trixie /artifacts.tar.gz /dependencies/             0.0s
 => CACHED [api  9/15] WORKDIR /dependencies                                                                  0.0s
 => CACHED [api 10/15] RUN set -a && source /env && set +a     && git config --global --add safe.directory /  0.0s 
 => CACHED [api 11/15] COPY api/go.mod api/go.sum /app/api/                                                   0.0s 
 => CACHED [api 12/15] WORKDIR /app/api                                                                       0.0s 
 => CACHED [api 13/15] RUN set -a && source /env && set +a     && go env     && go mod download     && sed -  0.0s 
 => CACHED [api 14/15] COPY api /app/api                                                                      0.0s 
 => CACHED [api 15/15] RUN set -a && source /env && set +a     && go env     && go build -v -o photoview .    0.0s
 => CACHED [release  5/10] RUN --mount=type=bind,from=api,source=/dependencies/,target=/dependencies/     /a  0.0s 
 => CACHED [release  6/10] COPY api/data /app/data                                                            0.0s 
 => CACHED [ui 2/6] WORKDIR /app/ui                                                                           0.0s
 => CACHED [ui 3/6] COPY ui/package.json ui/package-lock.json /app/ui/                                        0.0s
 => CACHED [ui 4/6] RUN npm install --global npm@10     && if [ "production" = "production" ]; then           0.0s
 => CACHED [ui 5/6] COPY ui/ /app/ui                                                                          0.0s
 => CACHED [ui 6/6] RUN export REACT_APP_BUILD_DATE="$(date -u +'%Y-%m-%dT%H:%M:%S+00:00(UTC)')";     export  0.0s
 => CACHED [release  7/10] COPY --from=ui /app/ui/dist /app/ui                                                0.0s
 => CACHED [release  8/10] COPY --from=api /app/api/photoview /app/photoview                                  0.0s
 => CACHED [release  9/10] RUN find /app/ui/assets -type f -name "SettingsPage.*.js"         -exec sed -i 's  0.0s
 => CACHED [release 10/10] WORKDIR /home/photoview                                                            0.0s
 => exporting to image                                                                                        0.1s
 => => exporting layers                                                                                       0.0s
 => => writing image sha256:8a08962a7e4b179eb22f2fba8810c3d0ea636978257debae0965c472c7331ed7                  0.0s
 => => naming to docker.io/googollee/photoview:clean_script                                                   0.0s
$ docker push googollee/photoview:clean_script
The push refers to repository [docker.io/googollee/photoview]
5f70bf18a086: Layer already exists                                                                                 
b256b62bb48b: Layer already exists                                                                                 
86c3d44ad181: Layer already exists                                                                                 
3e220d2d3b66: Layer already exists                                                                                 
b42ba73e2fad: Layer already exists                                                                                 
80b6bcb0da2d: Layer already exists                                                                                 
8c71efdfcfb1: Layer already exists                                                                                 
c1bd5b3d8e54: Layer already exists                                                                                 
f3c52d0fe253: Layer already exists                                                                                 
60e70dddd9ea: Layer already exists                                                                                 
clean_script: digest: sha256:2b8a6cdabbe424f84c2debe1e177885ba9b7273ae7c6c672ffce693ff4e79e54 size: 2415
```

For `linux/arm64`, use `docker build --platform linux/arm64 ...` to build all images.

Dockerfile
```diff
@@ -102,29 +110,31 @@ ARG TARGETPLATFORM
...
+COPY --chmod=0755 scripts/install_runtime_dependencies.sh /app/scripts/
 WORKDIR /dependencies
-RUN --mount=type=bind,from=api,source=/dependencies/,target=/dependencies/ \
-    chmod +x /app/scripts/install_runtime_dependencies.sh \
-    # Create a user to run Photoview server
-    && groupadd -g 999 photoview \
-    && useradd -r -u 999 -g photoview -m photoview \
-    # Install required dependencies
-    && /app/scripts/install_runtime_dependencies.sh \
+RUN --mount=type=bind,from=dependencies,source=/dependencies/,target=/dependencies/ \
```
googollee, line 117: I optimize the build parallel in the release stage. The build time (without cache) reduces 7m30s -> 6m30s on my server.

- `install_runtime_dependencies.sh` doesn't depend on ui and api stages, and run it first while waiting for building other stages.
- Building ui is faster than building api. Copy assets from ui stage and compress assets first while waiting for ui stage.
- Finally, copy binaries from ui stage and install.

It makes more layers, but it increases the whole building speed with better parallelism between stages. It also works when building the image after changing the code.

The parallel optimization only works with multiple CPUs.

Dockerfile
```diff
@@ -52,27 +61,26 @@ ENV PATH="${GOPATH}/bin:${PATH}"
...
+RUN source /app/scripts/set_compiler_env.sh
+
+COPY --chmod=0755 scripts/install_build_dependencies.sh /app/scripts/
+RUN set -a && source /env && set +a \
+    && /app/scripts/install_build_dependencies.sh
+COPY --chmod=0755 scripts/install_runtime_dependencies.sh /app/scripts/
+RUN set -a && source /env && set +a \
     && /app/scripts/install_runtime_dependencies.sh
 
 # hadolint ignore=DL3022
```
kkovaletp: @googollee, as you removed the COPY command, I guess that its ignore comment is not needed anymore. Please also review other similar comments for consistency and check the Hadolint job results - it might report some warnings without failing the build.

Dockerfile
```diff
@@ -142,6 +152,8 @@ RUN find /app/ui/assets -type f -name "SettingsPage.*.js" \
         zstd -k -f -19 -T0 --no-progress "$file"; \
     done' sh {} +
 
+COPY --from=api /app/api/photoview /app/photoview
```
kkovaletp, line 153: This step will never be cached because the COMMIT build arg is injected earlier and will always invalidate the cache of that and all later layers. That is why the whole workaround should be the last expensive step.
I know that COPY is not so expensive, but still, if it can be cached, it is better. Right?

kkovaletp:
This is not a big issue, but there is an inconsistency in the Debian OS settings, causing the following unneeded switch:
```
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (Can't locate Term/ReadLine.pm in @INC (you may need to install the Term::ReadLine module) (@INC entries checked: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.40.1 /usr/local/share/perl/5.40.1 /usr/lib/x86_64-linux-gnu/perl5/5.40 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.40 /usr/share/perl/5.40 /usr/local/lib/site_perl) at /usr/share/perl5/Debconf/FrontEnd/Readline.pm line 8, <STDIN> line 75.)
debconf: falling back to frontend: Teletype
debconf: unable to initialize frontend: Teletype
debconf: (This frontend requires a controlling tty.)
debconf: falling back to frontend: Noninteractive
```
As we know for sure that there will be automated scripts execution and a non-interactive console, we can set the Debian Frontend to Noninteractive in all stages of the Dockerfile except the `release` by setting `ENV DEBIAN_FRONTEND=noninteractive`.
For the release stage, it is possible to attach an interactive shell to the Photoview container for debugging purposes, so I'd keep it the default on the OS/image level and:
- set it in each script separately at the top
- prepend this var setting to each apt-get command in the Dockerfile's RUN step.

Here is the documentation for this frontend:

>    noninteractive
>           This is the anti-frontend. It never interacts with you  at  all,
>           and  makes  the  default  answers  be used for all questions. It
>           might mail error messages to root, but that's it;  otherwise  it
>           is  completely  silent  and  unobtrusive, a perfect frontend for
>           automatic installs. If you are using this front-end, and require
>           non-default  answers  to questions, you will need to preseed the
>           debconf database; see the section below  on  Unattended  Package
>           Installation for more details.

Dockerfile
googollee: Compression takes more time than copying a binary. When both ui and api are changed, compression first could be done while waiting for api.

I tend to keep it. When developing, COMMIT doesn't change, and this copy could be cached. In CI workflow, it's better to reduce the whole time.

Dockerfile
googollee: I removed all hadolint comments. It still looks happy.

googollee:
> DEBIAN_FRONTEND=noninteractive

Updated with ARG/ENV. Set it to noninteractive in CI workflow.

Dockerfile
kkovaletp: No, here is the warning reported:
Dockerfile:88 DL3062 warning: Pin versions in go. Instead of `go install <package>` use `go install <package>@<version>`

https://github.com/photoview/photoview/actions/runs/24444448825/job/71417075917?pr=1408

googollee: This one is a incorrect warning. `go install <package>` in a project with `go.mod` compiles the package with the version pinned in `go.mod`. It should not use `go install <package>@<version>`. See [the doc](https://pkg.go.dev/cmd/go#hdr-Compile_and_install_packages_and_dependencies):

> If the arguments have version suffixes (like @latest or @v1.0.0), "go install" builds packages in module-aware mode, ignoring the go.mod file in the current directory or any parent directory, if there is one.
> If the arguments don't have version suffixes, "go install" may run in module-aware mode or GOPATH mode, depending on the GO111MODULE environment variable and the presence of a go.mod file.

And about `GO111MODULE`, see [the doc](https://go.dev/ref/mod#mod-commands):

> GO111MODULE=on or is unset, the go command runs in module-aware mode, even when no go.mod file is present.

googollee: Added ignoring comment.

.github/workflows/build.yml
```diff
@@ -146,6 +146,7 @@ jobs:
             NODE_ENV=production
             VERSION=${{ github.ref_name }}
             COMMIT_SHA=${{ steps.git.outputs.COMMIT_SHORT_SHA }}
+            DEBIAN_FRONTEND=noninteractive
```
kkovaletp: There is no need for this build argument: there are no cases when another value needs to be provided:
- all non-final stages always need this value, as app build and test execution are non-interactive activities
- the final stage becomes the image, so the default value needs to be on the OS level with all scripts being called with the `DEBIAN_FRONTEND=noninteractive` prefix:
`DEBIAN_FRONTEND=noninteractive /app/scripts/install_runtime_dependencies.sh` for example. Then the env var is set only for this particular call (including &&).
- in case a need for manual debugging on a non-final stage (like, in the UI/API tests container), we can easily redefine the OS level env var in bash

Having this ARG just in case doesn't introduce an issue right now, but in the future, it might cause layer caching problems if this value is different across builds or dynamic.

googollee: > all non-final stages always need this value, as app build and test execution are non-interactive activities

This may be not true. We use api/ui stages in `dev-compose.yaml` for developing. It may need interactive front end in the developing environments. So we can't set a fixed value of the frontend in those stages.

> it might cause layer caching problems if this value is different across builds or dynamic.

I don't think that value should be changed frequently in a specific environment. I'd like to keep it until we really meet a caching problem.

Dockerfile
```diff
@@ -33,51 +33,59 @@ ARG VERSION=unknown-branch
 ENV VERSION=${VERSION}
 ARG TARGETARCH
 ENV TARGETARCH=${TARGETARCH}
-# hadolint ignore=SC2155
-RUN export REACT_APP_BUILD_DATE="$(date -u +'%Y-%m-%dT%H:%M:%S+00:00(UTC)')"; \
-    export REACT_APP_BUILD_COMMIT_SHA="-=<GitHub-CI-commit-sha-placeholder>=-"; \
-    export REACT_APP_BUILD_VERSION="${VERSION}-${TARGETARCH}"; \
-    export REACT_APP_API_ENDPOINT="${REACT_APP_API_ENDPOINT}"; \
```
kkovaletp, line 40: These `export` calls are needed for safety, because the `npm run build` is not the builder. It is closer to `make install`, where the main process reads the config and calls the relevant executors. It behaves like an orchestrator.
I don't remember the exact issues I saw when I tested different env var declaration approaches here, but the one with exports works the best.

.github/workflows/build.yml
kkovaletp: UPD: as this ARG is now different for the build and tests CI workflows, the cache is going to be always unused by the test building jobs.
This will be a problem when I create a PR with the cache optimization, based on my repo's changes.

kkovaletp: > This may be not true. We use api/ui stages in `dev-compose.yaml` for developing. It may need interactive front end in the developing environments. So we can't set a fixed value of the frontend in those stages.

ok, I see your point. Then, how about adding this env var to all our `.sh` scripts and prefix all APT and other sensitive calls with it, like in my example above?

googollee: Updated in the `Dockerfile`. It should be not added in `.sh` scripts because they may be run in the develop evnironment. Building an image is always no interactive process.

> UPD: as this ARG is now different for the build and tests CI workflows, the cache is going to be always unused by the test building jobs.

Our build workflow and test workflow have different ARGS now. For example, NODE_ENV is set as `production` in build and `testing` in test. It's the first ARG in ui stage, and it causes all cache invalid. I have no idea about it now and I decide not to touch them in this PR.

Dockerfile
googollee: I checked generated assets and those variables are injected. In general, forking processes should inherit environment from the parent.

Otherwise we may ignore the hadolint warning.

kkovaletp: If you want to keep the current implementation with no exports, I don't have specific details, as I tested this a long time ago and don't recall the exact issue. I agree that child/sub-processes have to inherit the environment from the parent, but there were some corner cases...

@coderabbitai, can you please carefully analyze this code change in the context of the project and this conversation? Do you see any problems with replacing exported variables with the prefixed declaration for the `npm run build` call? Are there some not-obvious corner cases and differences in the behavior between these 2 approaches that might result in a different final result?

Dockerfile
```diff
@@ -33,51 +33,59 @@ ARG VERSION=unknown-branch
...
+
 ### Build API ###
 FROM --platform=${BUILDPLATFORM:-linux/amd64} golang:1.26-trixie AS api
 ARG TARGETPLATFORM
+ARG DEBIAN_FRONTEND
 
-# See for details: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
+ENV DEBIAN_FRONTEND="${DEBIAN_FRONTEND}"
```
kkovaletp: We don't need this and lines 54, 109, 113 with the prefixed calls

googollee: Yes. I forgot to remove them. Updated.

.github/workflows/build.yml
kkovaletp: > Our build workflow and test workflow have different ARGS now

This is true, but only for the UI test image. The API test image should not be affected by any previously existing ARGs.

kkovaletp: Resolving the thread, as now everything is consistent.

kkovaletp:
@googollee, I guess, we're done)
You can revert the flag in the build workflow, and I'll approve the PR.

Dockerfile
```diff
@@ -99,32 +105,33 @@ RUN set -a && source /env && set +a \
...
-    # Install required dependencies
-    && /app/scripts/install_runtime_dependencies.sh \
+RUN --mount=type=bind,from=dependencies,source=/dependencies/,target=/dependencies/ \
+    # One step to install dependencies and clean up to avoid storing cache in the layer.
+    DEBIAN_FRONTEND=noninteractive /app/scripts/install_runtime_dependencies.sh \
     # Install self-building libs
     && cp -a lib/*.so* /usr/local/lib/ \
     && ldconfig \
-    && apt-get install -y ./deb/jellyfin-ffmpeg.deb gzip brotli zstd \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y ./deb/jellyfin-ffmpeg.deb \
```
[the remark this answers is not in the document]
googollee: Done.

.github/workflows/build.yml
```diff
@@ -12,8 +12,8 @@ on:
     - cron: '18 1 * * 4'
 
 env:
-  IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
-  # IS_PUSHING_IMAGES: true # Switch on when pushing images with forked repos.
+  # IS_PUSHING_IMAGES: ${{ github.event_name != 'pull_request' && github.repository == 'photoview/photoview' }}
+  IS_PUSHING_IMAGES: true # Switch on when pushing images with forked repos.
```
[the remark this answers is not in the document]
googollee: Rolled back.

kkovaletp:
@googollee, I've resumed the bot's review, and we now have a list of findings)
Please review them as usual (nitpicks as well). I didn't check them all, but at least some of them look valid...

Dockerfile
```diff
@@ -99,32 +105,34 @@ RUN set -a && source /env && set +a \
...
-    && /app/scripts/install_runtime_dependencies.sh \
+RUN --mount=type=bind,from=dependencies,source=/dependencies/,target=/dependencies/ \
+    # One step to install dependencies and clean up to avoid storing cache in the layer.
+    DEBIAN_FRONTEND=noninteractive /app/scripts/install_runtime_dependencies.sh \
     # Install self-building libs
     && cp -a lib/*.so* /usr/local/lib/ \
     && ldconfig \
-    && apt-get install -y ./deb/jellyfin-ffmpeg.deb gzip brotli zstd \
+    && DEBIAN_FRONTEND=noninteractive apt-get install \
+      -y --no-install-recommends ./deb/jellyfin-ffmpeg.deb \
```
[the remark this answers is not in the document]
googollee: Not really. `install_runtime_dependencies.sh` doesn't depend on `set_compiler_env.sh` and doesn't require `DEB_HOST_ARCH` to be set because it's not for compile. The script installs `curl` in it.

scripts/set_compiler_env.sh
```diff
@@ -1,38 +1,56 @@
-#!/bin/sh
-set -eu
+#!/bin/bash
+set -euo pipefail
 
 # Configure environment for cross-compiling.
 
-: ${TARGETPLATFORM=linux/`dpkg --print-architecture`}
+: "${TARGETPLATFORM=linux/$(dpkg --print-architecture)}"
```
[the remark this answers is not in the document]
googollee: `TARGETARCH` is `arm` if `DEBIAN_ARCH` is `armhf` or `armel`. So it's not necessary to check it again. See line 12 - line 20.

We don't support `i386` either.

googollee:
Well, after a new `photoview/dependencies` image was built, cross-building keeps failing, even I rolled back to a earily succeeded commit. I have no clue yet. I'll see if I can fix it.

dependencies/download_jellyfin-ffmpeg.sh
```diff
@@ -1,43 +1,34 @@
...
 
-echo "Downloading jellyfin-ffmpeg ${JELLYFIN_FFMPEG_VERSION} (cache miss)..."
+echo "Downloading jellyfin-ffmpeg ${JELLYFIN_FFMPEG_VERSION}..."
 
-echo Compiler: "${DEB_HOST_GNU_TYPE}" Arch: "${DEB_HOST_ARCH}"
+echo "Compiler: ${DEB_HOST_GNU_TYPE} Arch: ${DEB_HOST_ARCH}"
 
 VER="${JELLYFIN_FFMPEG_VERSION#v}"
-MAJOR_VER=$(echo "${VER}" | cut -d. -f1)
+MAJOR_VER="$(echo "${VER}" | cut -d. -f1)"
```
kkovaletp, line 27: @googollee, I just noticed this pattern in a few places across scripts: double-quotes inside a bigger quoted expression using the same double-quotes.
Are you sure that Bash processes this as needed instead of just finding the next double-quote to close the quoted sequence? I mean, is the whole expression still quoted instead of 2 separate quoted parts and the `${VER}` not quoted?

scripts/install_build_dependencies.sh
```diff
@@ -1,21 +1,56 @@
...
+  "liblcms2-dev:${DEB_HOST_ARCH}"
+  "zlib1g-dev:${DEB_HOST_ARCH}"
 
-# Install libheif dependencies
-apt-get install -y --no-install-recommends libdav1d-dev:${DEB_HOST_ARCH} libde265-dev:${DEB_HOST_ARCH} libjpeg62-turbo-dev:${DEB_HOST_ARCH} libopenh264-dev:${DEB_HOST_ARCH} libpng-dev:${DEB_HOST_ARCH} libnuma-dev:${DEB_HOST_ARCH} zlib1g-dev:${DEB_HOST_ARCH}
+  # ImageMagick dependencies
+  "libbz2-dev:${DEB_HOST_ARCH}"
+  "libdjvulibre-dev:${DEB_HOST_ARCH}"
+  "libfftw3-dev:${DEB_HOST_ARCH}"
+  # "libheif-dev:${DEB_HOST_ARCH}"
```
kkovaletp: Do we need to keep these commented-out libs here?
I see that this is part of your latest fix, but what is the idea behind it? Why does the presence of these libs break the build?

scripts/set_compiler_env.sh
```diff
@@ -1,47 +1,70 @@
...
+if [[ "${TARGETARCH}" != "amd64" && "${TARGETARCH}" != "arm64" ]]; then
+  echo "Warning: ${TARGETPLATFORM} is NOT supported. Just compile with the best efforts."
+
+  # best efforts
+  DEBIAN_ARCH="armel"
+  if [ "${TARGETVARIANT}" = "v7" ]
   then
-    DEBIAN_ARCH=armhf
+    DEBIAN_ARCH="armhf"
   fi
```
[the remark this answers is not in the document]
googollee: We don't support it, nor active failure. For other platforms like riscv, just try compiling with luck.

scripts/set_compiler_env.sh
```diff
@@ -1,47 +1,72 @@
...
-dpkg-architecture -a $DEBIAN_ARCH >/env
-export $(cat /env)
+  # best efforts
+  if [[ "${TARGETARCH}" = "arm" ]]
+    DEBIAN_ARCH="armel"
+    if [ "${TARGETVARIANT}" = "v7" ]
+    then
+      DEBIAN_ARCH="armhf"
+    fi
+  fi
```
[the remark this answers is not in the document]
googollee: Fixed.

dependencies/download_jellyfin-ffmpeg.sh
googollee: It works when not directly nesting quotes. I do that because Rabbitai suggests such quotes when the change touches it, so I decide to do it once and for all.

What do you mean `${VER}` not quoted?

scripts/install_build_dependencies.sh
googollee: I'm not sure how it fixes the build. I just submitted it to ensure it passes CI. I'm still working.

dependencies/download_jellyfin-ffmpeg.sh
kkovaletp: when we write the `"$(echo "${VER}" | cut -d. -f1)"` expression, we expect that Bash parses it in the following way:
1. resolve `"${VER}"` as the quoted string value and replace it in the original expression
2. resolve the whole expression as the quoted string

However, when processing quotes, Bash can simply go from left to right:

1. process the `"$(echo "` as the quoted string
2. process the `" | cut -d. -f1)"` as the quoted string
3. resolve the `${VER}` as unquated string
4. put everything together

I was thinking if the internal quotes should be escaped, like `"$(echo \"${VER}\" | cut -d. -f1)"`... I know that we cannot use single quotes, as they work differently in Bash and disable var substitution, which is not what we want here.

googollee: It's not hard to verify:

```bash
$ VER=7.1.3-5
$ MAJOR_VER="$(echo "${VER}" | cut -d. -f1)"
$ echo $MAJOR_VER
7
```

It parses with AST, not just matching characters.

scripts/install_build_dependencies.sh
googollee: libheif upgrades their APIs recently. With the latest dependencies image, it requires some new C API to build. When we build `photoview` image, to compatible with the dependencies image in this pr, we install libheif library from apt. The link finds the wrong `so` (to the one installed by apt, but not the one copied from dependencies).

Since we put all self-building files under `/usr/local/lib`, I added [`/usr/local/lib`](https://github.com/photoview/photoview/pull/1408/changes#diff-764fc0216eb38ea17a28d3c464f3903192b42dc0ecef86cdf5b764e22a2a8120R7) as the first place manually to `ldconfig`. It fixes the problem with the right order to load so files.

googollee: I updated `googollee/photoview:clean_script` with the latest code in PR. Let me know if you want to build a test image generated by the workflow.

googollee: The new libheif 1.12 supports sequences images and added a group of new APIs. Current Debian pins on 1.19.

kkovaletp: having 2 different versions of the same lib in the system might lead to more problems. Can we use our version everywhere consistently?

kkovaletp: > The new libheif 1.12 supports sequences images and added a group of new APIs. Current Debian pins on 1.19.

How can Debian pin a higher version than the released one?

googollee: Our current `dependencies` image has a higher version.

> having 2 different versions of the same lib in the system might lead to more problems. Can we use our version everywhere consistently?

It's because we are removing self-build `libheif` in this PR. To avoid workflow failure:

- When building with current `dependencies` image, it installs `libheif` from `apt-get` and copies `libheif` from `dependencies` image, and set `/usr/local/lib` for self-built `libheif` as high priority.
- When we have a new `dependencies` image which is built from this PR, it installs `libheif` from `apt-get`. It's not harm to set `/usr/local/lib` because there is no `libheif` in the new `dependencies` image, and we still put our self-built `so` files in that directory.

kkovaletp: Did I understand it correctly that this is just a temporary coexistence? Once this PR is merged, there will be a cleanup PR to make all things consistent, including the libheif installation.

googollee: Yes. All dependencies of libheif (but not libheif itself and plugins) in the script will be removed after we have a new dependencies image.

kkovaletp:
Is the PR ready for the final review, or are you still working on something?

googollee:
It's ready for review. Please take a look.

dependencies/build_imagemagick.sh
```diff
@@ -1,48 +1,48 @@
...
   echo "WARN: ImageMagick version is empty, most likely the script runs not on CI."
   echo "Fetching the latest version from ImageMagick repo..."
-  IMAGEMAGICK_VERSION=$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
-    "https://api.github.com/repos/ImageMagick/ImageMagick/releases/latest" | jq -r '.tag_name')
+  IMAGEMAGICK_VERSION="$(curl -fsSL --retry 2 --retry-delay 5 --retry-max-time 60 \
+    "https://api.github.com/repos/ImageMagick/ImageMagick/releases/latest" | jq -r '.tag_name')"
+  if [[ -z "${IMAGEMAGICK_VERSION}" ]]; then
+    echo "ERROR: Failed to resolve latest ImageMagick tag_name from GitHub API" >&2
+    exit 1
+  fi
```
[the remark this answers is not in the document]
kkovaletp: @googollee, this is not a big issue, as the script will fail later with an easily traceable error, but using `jq -er` in all such places looks like a simple fix.
So, this is up to you whether to apply it or skip.

kkovaletp (approved):
Other than the JQ minor issue, I don't see any other impediments, so LGTM

dependencies/build_imagemagick.sh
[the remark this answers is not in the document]
googollee: Fixed. Now even missing tags, `jq` outputs as empty strings.

dependencies/versions.sh
```diff
@@ -0,0 +1,41 @@
...
+  "https://api.github.com/repos/LibRaw/LibRaw/releases/latest" | jq -r '.tag_name // ""')
+IMAGEMAGICK_VERSION=$(curl "${CURL_FLAGS[@]}" \
+  "https://api.github.com/repos/ImageMagick/ImageMagick/releases/latest" | jq -r '.tag_name // ""')
+JELLYFIN_FFMPEG_VERSION=$(curl "${CURL_FLAGS[@]}" \
+  "https://api.github.com/repos/jellyfin/jellyfin-ffmpeg/releases/latest" | jq -r '.tag_name // ""')
+
+# Output as environment variables
+echo "LIBRAW_VERSION=${LIBRAW_VERSION}"
+echo "IMAGEMAGICK_VERSION=${IMAGEMAGICK_VERSION}"
+echo "JELLYFIN_FFMPEG_VERSION=${JELLYFIN_FFMPEG_VERSION}"
```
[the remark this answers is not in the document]
googollee: Not necessary. If these variables are empty, following scripts can handle it.

PR #1430, opened 2026-06-06
Prevent move/merge face if the target group is going to contain a duplicated pictures

kkovaletp:
## Summary by CodeRabbit

* **Bug Fixes**
  * Merge now validates up front and aborts with a clear error if combining groups would introduce duplicate media into the destination.
  * Moving image faces now validates during the move and prevents operations that would create duplicate media in the destination.
  * Ownership checks now correctly handle empty selections and non-admin users.

* **Tests**
  * Expanded coverage for merges, moves, duplicate-media rejection scenarios, and non-admin/ownership cases.

### DIFF
--- a/api/graphql/resolvers/faces.go
+++ b/api/graphql/resolvers/faces.go
@@ -178,6 +178,14 @@ func (r *mutationResolver) CombineFaceGroups(ctx context.Context, destinationFac
 		sourceFaceGroups = append(sourceFaceGroups, sourceFaceGroup)
 	}
 
+	hasDuplicateMedia, err := faceGroupsWouldContainDuplicateMedia(db, destinationFaceGroupID, sourceFaceGroupIDs)
+	if err != nil {
+		return nil, err
+	}
+	if hasDuplicateMedia {
+		return nil, errors.New("cannot merge face groups because the destination would contain duplicate images")
+	}
+
 	// Perform the merge
 	updateError := db.Transaction(func(tx *gorm.DB) error {
 
@@ -246,6 +254,14 @@ func (r *mutationResolver) MoveImageFaces(ctx context.Context, imageFaceIDs []in
 
 		for _, imageFace := range userOwnedImageFaces {
 			userOwnedImageFaceIDs = append(userOwnedImageFaceIDs, imageFace.ID)
+		}
+
+		hasDuplicateMedia, err := movingImageFacesWouldCreateDuplicateMedia(tx, destFaceGroup.ID, userOwnedImageFaceIDs)
+		if err != nil {
+			return err
+		}
+		if hasDuplicateMedia {
+			return errors.New("cannot move faces because the destination would contain duplicate images")
 		}
 
 		var sourceFaceGroups []*models.FaceGroup
--- a/api/graphql/resolvers/faces.go
+++ b/api/graphql/resolvers/faces.go
@@ -178,16 +178,16 @@ func (r *mutationResolver) CombineFaceGroups(ctx context.Context, destinationFac
 		sourceFaceGroups = append(sourceFaceGroups, sourceFaceGroup)
 	}
 
-	hasDuplicateMedia, err := faceGroupsWouldContainDuplicateMedia(db, destinationFaceGroupID, sourceFaceGroupIDs)
-	if err != nil {
-		return nil, err
-	}
-	if hasDuplicateMedia {
-		return nil, errors.New("cannot merge face groups because the destination would contain duplicate images")
-	}
-
 	// Perform the merge
 	updateError := db.Transaction(func(tx *gorm.DB) error {
+
+		hasDuplicateMedia, err := faceGroupsWouldContainDuplicateMedia(tx, destinationFaceGroupID, sourceFaceGroupIDs)
+		if err != nil {
+			return err
+		}
+		if hasDuplicateMedia {
+			return errors.New("cannot merge face groups because the destination would contain duplicate images")
+		}
 
 		if err := tx.
 			Model(&models.ImageFace{}).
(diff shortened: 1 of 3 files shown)


api/graphql/resolvers/faces.go
```diff
@@ -178,6 +178,14 @@ func (r *mutationResolver) CombineFaceGroups(ctx context.Context, destinationFac
...
+	hasDuplicateMedia, err := faceGroupsWouldContainDuplicateMedia(db, destinationFaceGroupID, sourceFaceGroupIDs)
+	if err != nil {
+		return nil, err
+	}
+	if hasDuplicateMedia {
+		return nil, errors.New("cannot merge face groups because the destination would contain duplicate images")
+	}
+
 	// Perform the merge
 	updateError := db.Transaction(func(tx *gorm.DB) error {
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, good catch. Please suggest a consistent and reliable fix for this issue by providing the exact code replacements for all required changes, including tests, if any.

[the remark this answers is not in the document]
kkovaletp: @coderabbitai, from your code block, it is unclear what exactly I need to remove, change, or replace to apply it based on the current code. Please give me the exact code replacements, from which I can easily understand what and where I should do.

api/graphql/resolvers/faces.util.go
```diff
@@ -59,19 +59,34 @@ func userOwnedFaceGroup(db *gorm.DB, user *models.User, faceGroupID int) (*model
 }
 
 func getUserOwnedImageFaces(tx *gorm.DB, user *models.User, imageFaceIDs []int) ([]*models.ImageFace, error) {
-	if err := user.FillAlbums(tx); err != nil {
-		return nil, err
+	var userOwnedImageFaces []*models.ImageFace
```
googollee: Defining variables close to where they are used is a good practice. Suggest:

```go
if len(imageFaceIDs) == 0 {
  return nil, nil
}

// ...

var userOwnedImageFaces []*models.ImageFace
if err := query. // Line 89
	Where(imageFacesIDInQuestion, imageFaceIDs).
	Find(&userOwnedImageFaces).Error; err != nil {
	return nil, err
}
```

api/graphql/resolvers/faces.util.go
```diff
@@ -80,6 +95,49 @@ func getUserOwnedImageFaces(tx *gorm.DB, user *models.User, imageFaceIDs []int)
 	return userOwnedImageFaces, nil
 }
 
+func faceGroupsWouldContainDuplicateMedia(db *gorm.DB, destinationFaceGroupID int, sourceFaceGroupIDs []int) (bool, error) {
```
googollee: How about `HasDuplicateMedia(db *from.DB, faceGroupIDs ...int)`? Since the parameters have `faceGroupIDs`, it's easy to know that the duplicate media is in those face groups.

Where is this function called?

api/graphql/resolvers/faces.util.go
```diff
@@ -80,6 +95,49 @@ func getUserOwnedImageFaces(tx *gorm.DB, user *models.User, imageFaceIDs []int)
...
+	if err := db.
+		Table("(?) AS duplicate_media", duplicateMediaQuery).
+		Count(&count).Error; err != nil {
+		return false, err
+	}
+
+	return count > 0, nil
+}
+
+func movingImageFacesWouldCreateDuplicateMedia(db *gorm.DB, destinationFaceGroupID int, imageFaceIDs []int) (bool, error) {
```
googollee: Where is this function called?

api/graphql/resolvers/faces.util.go
kkovaletp: @googollee,
- the `HasDuplicateMedia` name doesn't look correct, as at the function execution time, each group doesn't contain duplicates yet. The destination group might contain them in the future if the merge transaction runs. Maybe, `wouldHaveDuplicateMediaAfterMerge`?
- basically, having shorter params for the function requires moving the `faceGroupIDs := append([]int{destinationFaceGroup.ID}, sourceFaceGroupIDs...)` to the caller side. I've done it, take a look.
- this function is called from the `faces.go` line 184.

api/graphql/resolvers/faces.util.go
kkovaletp: @googollee, to make the function names similar, maybe rename this one to `wouldHaveDuplicateMediaAfterMoveFace`?
This function is called from the `faces.go`, line 259.

api/graphql/resolvers/faces.util.go
googollee: GitHub doesn't show `faces.go` content by default. :/

How about `HasDuplicateMediaInFaceGroups`? "Would" sounds incorrect because, in the current state—even without any movement—the groups as a whole already contain duplicate media.

api/graphql/resolvers/faces.util.go
googollee: Yeah, I don't like the name `movingImageFacesWouldCreateDuplicateMedia()`. The function name implicts that the following must be a move. If it's in a workflow, it's fine, but it's in `util`, and it can be used in other behaviors.

Since it checks a status, how about `haveDuplicateFacesInGroup(DB *grom.DB, faceGroupID int, imageFaceIDs []int) (bool, error)`?

api/graphql/resolvers/faces.util.go
kkovaletp: I've added the `Union` at the end of your suggested name. Please check.

api/graphql/resolvers/faces.util.go
kkovaletp: I realized that this is all overcomplicated... I've named it based on what it actually does: `doesFaceGroupContainImages()`. The function shouldn't know from which context and for what business purpose it is called, so this name looks better. What do you think?

googollee (approved):
LGTM

PR #1446, opened 2026-07-10
Return HTTP 404 for missing file requests

kkovaletp:
The current implementation always falls back to the `index.html` for not existing paths in requests.
The issue is that if a request asks for a file and expects the specific MIME type in the response, getting the `index.html` instead causes the "MIME type mismatch" error in the browser's console instead of the expected "File not found" error.

This PR fixes this case.

I've declared the static list of file extensions for missing files instead of a regexp, guessing that a file was requested based on a dot in the URL or a more complex condition - such guessing will always be unreliable and can cause more bugs than solve.

## Summary by CodeRabbit

* **Bug Fixes**
  * Missing JavaScript, CSS, image, and font files now return a proper 404 error instead of loading the application shell.
  * Improved handling of requests for nonexistent static assets.

* **Tests**
  * Added coverage to verify missing asset requests return 404 responses.

### DIFF
--- a/api/routes/spa.go
+++ b/api/routes/spa.go
@@ -11,6 +11,33 @@ import (
 	"github.com/photoview/photoview/api/log"
 )
 
+// staticAssetExtensions lists file extensions considered "static asset" requests.
+// Missing files with these extensions return 404 instead of falling back to
+// index.html. This mirrors the `path_regexp staticAsset` matcher used in
+// ui/Caddyfile's `@missingStaticAsset` block, so both the Go SPA handler and
+// Caddy behave consistently.
+var staticAssetExtensions = []string{
+	".js", ".mjs", ".css",
+	".html", ".htm",
+	".md", ".map", ".json", ".webmanifest",
+	".png", ".svg", ".ico",
+	".jpg", ".jpeg", ".gif", ".webp", ".avif",
+	".woff", ".woff2", ".ttf", ".eot",
+	".txt", ".xml", ".wasm",
+}
+
+// isStaticAssetExtension reports whether ext (as returned by filepath.Ext)
+// belongs to the known set of static asset extensions.
+func isStaticAssetExtension(ext string) bool {
+	ext = strings.ToLower(ext)
+	for _, knownExt := range staticAssetExtensions {
+		if ext == knownExt {
+			return true
+		}
+	}
+	return false
+}
+
 // SpaHandler implements the http.Handler interface, so we can use it
 // to respond to HTTP requests. The path to the static directory and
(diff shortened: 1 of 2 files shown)


googollee:
The code looks good. But I don't like the way of checking file extensions. I think we should create a 404 page for all not-found requests, and provide a redirect link to the index. Even a simple text 404 page is enough in the first step.

kkovaletp:
@googollee, in general, I agree with you: replying with a 404 response is a bad UX, and most of the products redirect users to a custom error page instead. 
However, this PR fixes a slightly different issue. Here is the scenario:
- We run a previous Photoview image (container), and users have open sessions in their browsers
- Now, we decide to update Photoview on the server by pulling a newer image and recreating the Compose services
- So, Photoview restarts in a new container, based on the new image
- This new image has some UI changes, so Vite (the UI builder) has rebuilt the UI bundle, and all assets now have different hashes in their names (and even some assets were renamed, because Vite decides which pages to pack in which asset files with no guarantees of keeping the same final asset names).
- After the server restart, users continue working with Photoview in their sessions, as they don't yet know that the server was updated and assets have changed. Once they try to switch pages (Timeline, People, Map, etc.), the browser checks if the corresponding asset is in the local cache, and if not, it asks for it from the server based on the old list of assets from the previous version (lazy loading). It asks for all relevant files for that asset (JS, CSS, and MAP for debug mode)
- And the issue is that there is no such asset on the server anymore, so the router falls back to serving the `index.html` with a different MIME type. And this leads to type mismatch errors in the dev tools console and normal communication in the dev tools network tab. This is completely confusing behavior, as clients ask for non-existent files and have to get the correct 404 responses for those requests.

I hope that I explained the issue clearly. If not, feel free to ask questions.

Your proposal to serve a custom 404 page is good, but it would look like the current behavior, with the only difference that now we return a response with a redirect to another page. Still not what is usually expected.
Instead of serving the error page from the server with a redirect in response, I'd propose to implement a listener for 4XX and 5XX response codes, which should catch the error message and show it in the error notification pop-up (like we do for scanner notifications). This would be UI code, and we have no maintainer to review it. Are you ok to do the review?

googollee:
Yes. You can send me UI changes for review.

I know the case here. Ideally, the router should redirect not-existed resources to 404 page. I know redirecting to index is easy for developing, but it's hard to identify missing resources by updating server and invalid URL.

kkovaletp:
Ok, so I'll make a new branch from this PR and develop the UI listener with notifications. Once done, I'll tag you for the review.
It might take some time...

PR #1448, opened 2026-07-10
Optional `rootPath` for the `user` resolver

kkovaletp:
This PR adds the optional `rootPath` field to the `user` resolver.
There are no changes in UI, and these changes are mostly for compatibility with my external UI image, which has the corresponding functionality.
The main changes are in the `api/graphql/resolvers/user.go`, which is not shown on GitHub by default, so please click on the "Load diff". 
I hope that these changes are safe for this repo and won't introduce any issues, as if the UI doesn't send the new `rootPath` field, this code should be skipped and the backend should work as before.

## Summary by CodeRabbit

* **New Features**
  * Added an optional root path when creating a user.
  * When provided, the root path is cleaned and used to create the user’s initial root album.
* **Bug Fixes**
  * Ensured user registration and root album creation are completed together, preventing partially completed setup.

### DIFF
--- a/api/graphql/resolvers/user.go
+++ b/api/graphql/resolvers/user.go
@@ -145,7 +145,7 @@ func (r *mutationResolver) UpdateUser(ctx context.Context, id int, username *str
 }
 
 // CreateUser is the resolver for the createUser field.
-func (r *mutationResolver) CreateUser(ctx context.Context, username string, password *string, admin bool) (*models.User, error) {
+func (r *mutationResolver) CreateUser(ctx context.Context, username string, password *string, admin bool, rootPath *string) (*models.User, error) {
 	var user *models.User
 
 	transactionError := r.DB(ctx).Transaction(func(tx *gorm.DB) error {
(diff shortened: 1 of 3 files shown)


api/graphql/resolvers/user.go
```diff
@@ -155,6 +155,14 @@ func (r *mutationResolver) CreateUser(ctx context.Context, username string, pass
 			return err
 		}
 
+		if rootPath != nil && *rootPath != "" {
+			cleanedPath := path.Clean(*rootPath)
```
googollee, line 159: How about moving `Clean()` into `NewRootAlbum()`. And do you have a plan to use this new API?

googollee: See https://github.com/photoview/photoview/pull/1447/changes#diff-46ce03b017c3c7c7fd35cc530af92c87f97c818314b1182d0a02e1e3c8fcc1d2R21

kkovaletp: @googollee, sure, I can remove the `cleanedPath := path.Clean(*rootPath)` line here, as it is also cleaned inside the `NewRootAlbum()`, like you pointed.
I added it here just to satisfy our code scanner (but it is still noisy and unhappy, thinking that we keep using unverified user data).

I use this API in my forked UI that is running in a separate container and can be connected to this backend or to the backend from my forked repo. I proposed this PR here just for compatibility reasons, so my UI can still work with this repo's backend. As we don't have a UI maintainer and nobody can review my PRs, I develop all UI fixes and improvements in my forked repo for at least 1 year (or even longer), so my version of UI is far away from the main repo's version. This means that I cannot backport my changes to the main repo. If there is a maintainer in the future, there will be only one option to perform the code review of my UI and move it at once to the main repo, replacing the current one entirely.

googollee: Well. OK. It's no harm to do that twice.

googollee (approved):

PR #1452, opened 2026-07-19
Add labels for share links

floze-the-genius:
## Summary

- persist an optional label on album and media share tokens
- allow owners and administrators to relabel shares from the existing sidebar options
- keep organizational labels private from anonymous and unrelated share-token viewers
- trim labels, clear blank values, and preserve user-entered casing
- surface rename failures inline and migrate existing databases through the standard GORM AutoMigrate path
- update extracted translation catalogs, including English defaults and Russian translations

## Validation

- `npm test -- --run src/components/sidebar/Sharing.test.tsx` (4 passed)
- `npx eslint src/components/sidebar/Sharing.tsx src/components/sidebar/Sharing.test.tsx --max-warnings 0 --config .eslintrc.js --no-cache`
- `npx prettier --check src/components/sidebar/Sharing.tsx src/components/sidebar/Sharing.test.tsx src/components/sidebar/__generated__/sidebarGetAlbumShares.ts src/components/sidebar/__generated__/sidebarGetPhotoShares.ts src/components/sidebar/__generated__/sidebarSetShareTokenLabel.ts src/extractedTranslations/{da,de,en,es,fr,it,ja,nl,pl,pt,ru,sv,uk,zh-CN,zh-HK,zh-TW}/translation.json`
- `npm run extractTranslations`
- `npm run genSchemaTypes`
- `npm run build`
- `go test ./database -run TestMigrateDatabaseAddsShareTokenLabel -count=1`
- `git diff --check`

The focused database migration test passes. The actions/resolver Go test packages cannot compile locally because the ImageMagick, JPEG, and dlib development headers required by existing transitive packages are not installed. GraphQL bindings were regenerated with gqlgen's supported `skip_validation` mode, and the temporary setting was removed afterward.

Closes #1449

## Summary by CodeRabbit

* **New Features**
  * Added optional labels to album and media share links.
  * Added a new share-label update action and UI support to edit/clear labels from the sharing panel.
  * Share labels are displayed in the shares list and persisted after trimming/sanitization.
* **Bug Fixes**
  * Share token labels are only visible to the share owner or administrators.
* **Tests**
  * Added/updated backend, UI, and migration tests to cover label creation, editing, visibility rules, sanitization behavior, and error handling.

PR #1454, opened 2026-07-24
Prevent ExifTool argument injection from multiline paths

fedek-xbow:
## Summary

- Reject carriage returns and line feeds in arguments before writing to ExifTool's persistent argument stream.
- Validate the complete command before changing reader state or buffering bytes, preventing rejected input from desynchronizing the next request.
- Add regression tests for CR/LF injection, zero-byte writes on rejection, and successful process reuse afterward.

## Security impact

Photoview runs ExifTool with `-stay_open True -@ -` and writes each argument as a line on standard input. Unix filenames may contain line breaks, so writing a scanner-controlled path verbatim can turn one path into multiple ExifTool arguments. An attacker who can place a scanner-visible file with such a name can inject ExifTool options and reach command execution under the Photoview process account.

The validation is applied in `rawSendCommand`, covering metadata queries, MIME detection, embedded-preview extraction, output paths, and future call sites using this transport.

## Tests

- `go test ./scanner/externaltools/exiftool -count=1`
- `go test -tags no_face_detection ./scanner/... -count=1`
- `go test -tags no_face_detection ./... -p 1 -count=1`
- `go vet -tags no_face_detection ./scanner/...`

All commands pass in the project's dependency image with ExifTool 13.25.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security when processing ExifTool commands by rejecting arguments containing line breaks.
  * Prevented potentially injected command content from being sent.
  * Ensured ExifTool remains usable after rejecting invalid input.

### DIFF
--- a/api/scanner/externaltools/exiftool/exiftool.go
+++ b/api/scanner/externaltools/exiftool/exiftool.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"os/exec"
+	"strings"
 	"sync"
 )
 
@@ -29,6 +30,8 @@ type Exiftool struct {
 
 const marker = "{ready}\n"
 const bufferSize = 10240
+
+var errArgumentContainsLineBreak = errors.New("exiftool arguments cannot contain line breaks")
 
 // New returns a new instance of Exiftool.
 func New() (*Exiftool, error) {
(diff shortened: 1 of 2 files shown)


kkovaletp:
@fedek-xbow, thanks for the contribution!
While this PR technically fixes the bug, the incorrect path already exists in the DB at this point, so potentially another code or external tool can still be vulnerable.
So, to properly handle this case, I'd ask you to analyze the https://github.com/photoview/photoview/tree/master/api/scanner pipeline (you can start reading the code from the https://github.com/photoview/photoview/blob/master/api/scanner/scanner_queue/queue.go#L223) and find the places where we discover the album path and media path for the 1st time. These would be the best points to sanitize them before usage and store them in the DB after making sure that the sanitized path still points to the FileSystem's object.

The sanitizing itself can be done in a similar way, as suggested in the https://github.com/photoview/photoview/pull/1452#discussion_r3650112801 comment (the exact cases need to be double-checked against FS-related limitations).

PR #1459, opened 2026-08-02
fix: reload sharing options when sidebar media changes

### LINKED ISSUE #925: Access link is not reloaded on the sidebar under "Sharing options" if user clicks on (i) of different media without closing the sidebar
**Describe the bug**
I have a list of media on my timeline. When I hover the mouse over them, I see the (i) button for each, which opens the sidebar with image details.
So, if I click on the (i) of different media without closing the sidebar, I'll always see the same content under the "Sharing options", as the 1st clicked media has.

**To Reproduce**
1. Make sure that you have some media with shared access links configured, and some without them
2. Open Photoview on the Timeline page with the sidebar closed
3. Click on (i) of some media without the link
4. Click on (i) of some media with the link
5. See that the link is not shown
6. Close the sidebar and reopen it again for the last media
7. The link is shown
8. Repeat the steps 2-7, but reverse the order of 3 and 4. Expect the opposite behavior in 5 and 7

**Expected behavior**
All the content of the sidebar is reloaded for each media even if the sidebar is not closed

**Additional context**
I think that other types of content in the sidebar might have the same issue, so it makes sense to validate that as well

waterWang:
## Description

When the sidebar is pinned and the user clicks on a different media item without closing the sidebar, the "Sharing options" section shows stale data from the previously viewed media. The access link is not refreshed.

## Root Cause

The `SidebarPhotoShare` component uses `useLazyQuery` with a `useEffect` that has an empty dependency array `[]`. When the `id` prop changes (new media selected), the effect does not re-run, so the shares query is never re-executed.

## Fix

Added `id` to the `useEffect` dependency array so the shares query is re-fetched whenever the media selection changes.

## Test Plan

1. Open Photoview on the Timeline page
2. Click on the (i) button of a media item without a share link
3. Without closing the sidebar, click on (i) of a different media item that has a share link
4. **Before fix**: The share link from the first media is still shown
5. **After fix**: The share link is correctly loaded for the new media

Closes #925

## Summary by CodeRabbit

* **Bug Fixes**
  * Photo-sharing details now refresh correctly when switching between photos in the sidebar.

### DIFF
--- a/ui/src/components/sidebar/Sharing.tsx
+++ b/ui/src/components/sidebar/Sharing.tsx
@@ -460,7 +460,7 @@ export const SidebarPhotoShare = ({ id }: SidebarSharePhotoProps) => {
         },
       })
     }
-  }, [])
+  }, [id])
 
   const loading = queryLoading || mutationLoading
 


kkovaletp:
@waterWang , this is a duplicate PR to your other PR #1460

PR #1486, opened 2026-09-13
Stop logging users out on every network hiccup

try-and-succeed:
to all five.

### Any network hiccup logs the user out

The Apollo error link clears the auth cookie for **every** network error: a request timeout, a
laptop waking up on a different network, a server-side 500. None of those say anything about the
user's session, yet all of them end it.

Only an actual authentication failure does now — HTTP 401/403, or a GraphQL `unauthorized` error —
and the "you are being logged out" note follows the same condition instead of being printed
unconditionally.

The notification subscription is excluded from the auth handling altogether. It reconnects on its
own after a server restart and can transiently see a stale token while doing so, which is not the
user's session being rejected.

Two smaller defects in the same handler come along:

- a plain connection failure carries no `result`, and reading the error list through it threw from
  inside the error handler itself;
- a connection that never reached the server produced no message at all, leaving the user with
  nothing to go on. It now reports that the server could not be reached.

### Testing

- `npm run build`, `eslint` on the changed file

The handler has no unit test: it is wired into the Apollo link chain, and the behaviour worth
pinning down (a dropped connection must not clear the cookie) needs a client instance and a
failing transport. Happy to add one if you would rather have it covered.

was verified locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication is cleared only after unauthorized HTTP 401 or 403 responses; GraphQL authorization errors and subscription failures preserve the session.
  * Subscription failures no longer create duplicate notifications.
  * HTTP 500 responses and connection failures now preserve the session.
  * Added clearer fallback notifications when the server cannot be reached.
  * Improved handling and display of server-provided error messages.
* **Localization**
  * Added notification translation entries across supported languages.


try-and-succeed:

ui/src/apolloClient.ts
try-and-succeed, line 1: @coderabbitai, I'll write this comment here, but it is relevant for all changes in this PR, not only this file. As a final review step before asking the maintainers to merge, please carefully analyze this PR within the scope of the whole app, including business logic, UX, the testing framework and other relevant aspects.

Context: this PR stops the Apollo error link from clearing the auth cookie on every network error. Only an HTTP 401/403 or a GraphQL `unauthorized` error logs the user out now, subscription errors never do, and a connection that never reached the server produces a notification instead of throwing inside the handler. Pay particular attention to: whether any genuinely invalid session could now survive and leave the UI in a broken logged-in state; the interaction with the WebSocket subscription's reconnect; and every place in the app that relied on the old behaviour of being logged out after an error.

Please answer:

try-and-succeed: Thanks — the duplicate subscription messages are fixed in `d8a1857`, together with the missing tests.

- The error link no longer adds messages for subscription operations; `SubscriptionsHook` owns those. It still logs them, and still never logs the user out over one.
- Writing the tests turned up a second duplicate, older than this PR: for a response with an error body, Apollo passes the same errors both as `graphQLErrors` and inside `networkError.result`, and both branches reported them. They are reported once now, plus a short note when a 401/403 ends the session.
- Seven tests cover the link, following your list: 401, 403 and GraphQL `unauthorized` log out; a 500 and a failed connection keep the session and report once without throwing; a failing subscription neither logs out nor adds a message. Removing the subscription exception or the duplicate check each fails a test.

ui/src/apolloClient.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    expect(clearTokenCookie).toHaveBeenCalled()
+  })
+
+  test('an HTTP 401 with an error body reports the error and the logout, once each', async () => {
+    await run(QUERY, {
+      error: httpError(401, [{ message: 'invalid authorization token' }]),
+    })
+
+    expect(clearTokenCookie).toHaveBeenCalled()
+    expect(messagesAdded).toBe(2)
```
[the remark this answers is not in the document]
try-and-succeed: Fixed in `555c77f`. You are right that a count of two would also accept the same error twice — the duplicate the handler now suppresses. The tests collect the headers of the added messages and assert them exactly: `Something went wrong` and `Session ended` for a 401 with an error body, one `Something went wrong` for a 500, one `Network error` for a connection that never reached the server. Reverting the duplicate check fails two of them.

try-and-succeed:
{"body":"On the collapsed finding in https://github.com/photoview/photoview/pull/1486#pullrequestreview-5215432131 (cover GraphQL errors in the subscription tests): agreed and fixed in the latest commit. You were right that the guard was untested — the subscription cases only sent network errors, so removing the `!isSubscription` check on the GraphQL `unauthorized` path left every test green. There is now a case that sends that error through a subscription and asserts both that the session survives and that no message is added; removing the guard fails it."}

ui/src/apolloClient.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    // user's own session is perfectly valid. Only the network-error path was
+    // covered here, so removing the guard on this one would have gone
+    // unnoticed.
+    await run(SUBSCRIPTION, {
+      result: { errors: [new GraphQLError('unauthorized')] },
+    })
+
+    expect(clearTokenCookie).not.toHaveBeenCalled()
+    expect(headersAdded).toEqual([])
+  })
```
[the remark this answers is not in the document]
try-and-succeed: Added in the latest commit. You are right that only the 401 went through a subscription: the guard treats both alike, so weakening it for 403 alone would have kept every test green. The subscription case now sends 401, 403 and a dead socket, and asserts for all three that the session survives and no message is added — splitting the guard so that 403 logs out fails it.

kkovaletp:
@try-and-succeed, thanks for the contribution.
I checked your implementation, and I don't think that it is the optimal one.
Please take a look at this implementation in my forked repo: https://github.com/kkovaletp/photoview/blob/master/ui/src/apolloClient.ts
Feel free to backport it here and optionally enhance it if you want.

BTW, please pay attention to the notification messages implementation: In my repo, they are translated and contain message keys.

try-and-succeed:
Thanks for looking at it!

**Translated messages:** good catch, that one is fixed in `d0b5e3c`. The notification messages now go
through i18next under `notification.error.*`, with the previous English text as the default value,
and the keys are extracted into all locale bundles with `npm run extractTranslations`. There is also
a test that pins the wiring down: it registers a translation for one key and asserts the message
uses it, so a hardcoded string would fail it.

**Your implementation:** I would rather not copy code out of your fork, if you don't mind. Your repo
adds license terms on top of this project's AGPL-3.0, and I don't think I can bring code carrying
those terms into the main repo without creating a licensing question for everyone downstream. If you
would like your version here, it would be cleanest coming from you as its own PR — you are the
author, so that question does not arise.

What I would very much like is to fix what you found in this PR. Could you tell me what you consider
non-optimal about this implementation? If it is the shape of the handler (the `isSubscription`
branching, the duplicate-message guard, the 401/403 check), I am happy to restructure it however you
prefer. The behaviour this PR is after is narrow:

- only an HTTP 401/403 or a GraphQL `unauthorized` error clears the cookie, so a timeout, an offline
  client or a server-side 500 no longer logs the user out;
- subscription errors never clear it, because the notification subscription reconnects on its own and
  can transiently see a stale token;
- a connection that never reached the server reports something instead of nothing, and reading the
  error list through a missing `result` no longer throws inside the handler.

Happy to adjust any of that.

ui/src/apolloClient.ts
kkovaletp, line 1: @coderabbitai, this PR improves the Apollo Client interactions with the API server by better handling different types of errors. However, I have a different implementation of similar functionality in my forked repo: https://github.com/kkovaletp/photoview/blob/master/ui/src/apolloClient.ts
Could you please carefully analyze this file from the PR (including all unchanged lines for completeness) and then carefully analyze https://github.com/kkovaletp/photoview/blob/master/ui/src/apolloClient.ts from my fork (both analyses have to be in the scope of the corresponding repo to ensure consistency of the codebase).
Once done, please compare both implementations, list what is implemented differently or missing in one of them (always clearly point to the source of each implementation, so I can easily understand what you are referring to). If similar functionality is implemented differently in both files, please provide pros and cons of each implementation, followed by your recommendation of which one is more reliable, functional, efficient, and consistent with its own codebase. If you see that something important is missing in both implementations based on your analysis, please highlight it as well.

kkovaletp:
@try-and-succeed,

> There is also
> a test that pins the wiring down: it registers a translation for one key and asserts the message
> uses it, so a hardcoded string would fail it.

I don't think that this is the correct way to test the translations' correctness. The tool has an extractor and linter, which should be used for all UI code at once and highlight all inconsistencies instead of manually writing and maintaining unit tests for specific lines: https://github.com/kkovaletp/photoview/blob/master/.github/workflows/tests.yml#L273-L276 - the example.

> Your repo adds license terms on top of this project's AGPL-3.0, and I don't think I can bring code carrying
> those terms into the main repo without creating a licensing question for everyone downstream

That is true, and taking someone's code for copy-paste into another project is a bad idea regardless of the license (apart from forking, which creates a base for another project). What I suggested was to analyze that implementation, understand what was done there and why, and compare it with your implementation to detect weak points and decide how to reimplement them in a better way. In other words, use it as inspiration, not as a copy-paste source. Sorry if I wasn't clear enough. However, to use my code even as a source of inspiration, you still need to accept that license.

ui/src/apolloClient.ts
kkovaletp: @try-and-succeed, here is an example of the automated implementations comparison. It is up to you to decide if you want to use these results and how exactly.

try-and-succeed:
Fair enough on both counts.

**The test is gone** (`34c34af`). The keys and their English defaults stay; only the unit test that
asserted one message goes through `i18n.t` is removed. For the record on what it was after: not the
correctness of any translation, but that a message reaches i18next at all — a hardcoded string would
have failed it. An extractor-based check is the better tool for that, and a much wider net.

Which leads to the gap: **this repository has no i18n job at all.** `npm run extractTranslations`
exists, but nothing in `.github/workflows/` runs it, so a `t()` call whose key never made it into the
bundles, or a message that skips `t()` altogether, goes unnoticed here. If you would like it, I am
happy to open a separate PR adding that check, written from scratch for this repo rather than taken
from yours:

- run `npm run extractTranslations` in CI and fail if the working tree is dirty afterwards, which
  catches keys that were added to the code but never extracted;
- optionally also fail on a `t()` call without a default value.

Say if you want it, and whether you would rather have it as its own workflow or a step in the
existing `Tests` one.

**On your fork as inspiration:** thanks for the clarification, and understood — since even reading it
that way means accepting the additional license, I will stay out of it and work from this repo plus
whatever you tell me directly.

That said, the comparison you asked CodeRabbit for is public in this thread, and its conclusion is
useful to both of us: it recommends this PR's error policy for authentication invalidation, and it
lists your fork's infrastructure improvements (`graphql-ws` reconnect control, connection-ack
timeout, backoff with jitter, page-unload disposal, the queued message provider, the `keyArgs`
refinement) as things to adopt separately rather than fold in here. It also names two gaps that
exist on both sides, which I can pick up:

1. **`message === 'unauthorized'` is a string comparison.** A structured `extensions.code` from the
   server would be the stable contract. That is an API-side change, so it wants its own issue or PR —
   tell me if you want me to open one.
2. **`SubscriptionsHook` still adds a message per subscription error event**, so a reconnecting
   subscription can repeat a notification even though this PR stopped the error link from doubling
   it. Deliberately untouched here to keep this PR to one change, and a good follow-up.

Either of those as a follow-up PR, gladly — just say which you would like first.

kkovaletp:
@try-and-succeed 

> nothing in `.github/workflows/` runs it, so a `t()` call whose key never made it into the bundles, or a message that skips `t()` altogether, goes unnoticed here. If you would like it, I am happy to open a separate PR adding that check, written from scratch for this repo rather than taken from yours:
> 
> * run `npm run extractTranslations` in CI and fail if the working tree is dirty afterwards, which
>   catches keys that were added to the code but never extracted;
> * optionally also fail on a `t()` call without a default value.

Yes, having that CI check would be great. Running the extraction and checking the git state afterward is a good minimal approach. You can check how it is implemented in this repo for the API's `go generate` command. I'm not saying that this check should also be done this way. Just an example. Having the translations linter as part of this check would make it even more useful, but I'm not sure if this old version of the lib supports it.
I think that this check should be part of the testing workflow. Can be a separate job or part of the UI testing job, depending on how you'd like to implement it.

> since even reading it
> that way means accepting the additional license, I will stay out of it and work from this repo plus
> whatever you tell me directly.

Good. BTW, are you a citizen of one of the countries banned by that license explicitly?
Yes, the results from the bot's comparison can be used because that comparison was initiated by me and the results are public in this repo.

> it recommends ... as things to adopt separately rather than fold in here.

I'd not rely on the bot's recommendations and instead use the exact comparison results. The bot is good at code analysis and comparison, but it is bad at making decisions (same as all currently existing AI models). So, it is up to you what to implement and the scope of which PR(s)

> 1. **`message === 'unauthorized'` is a string comparison.** A structured `extensions.code` from the
>    server would be the stable contract. That is an API-side change, so it wants its own issue or PR —
>    tell me if you want me to open one.

Can't we use HTTP/WebSocket codes for that? Are there cases with those codes missing? If yes, let's have an issue/PR for that.

> 2. **`SubscriptionsHook` still adds a message per subscription error event**, so a reconnecting
>     subscription can repeat a notification even though this PR stopped the error link from doubling
>     it. Deliberately untouched here to keep this PR to one change, and a good follow-up.

Can this be covered by using the specific notification key for those events? Then, the new notification would replace the previous one with the same key

try-and-succeed:
@kkovaletp, thanks. Both of your questions turned into changes, in `b4c4d33`.

### HTTP codes instead of `message === 'unauthorized'`: yes, the string check is gone

I checked where each signal comes from on the API side:

- **An unknown or expired token** is turned away by `auth.Middleware` with **HTTP 401**, before any resolver runs. The loader filters on `expire > now`, so expiry lands there too. The status code covers this case on its own.
- **A GraphQL `unauthorized`** arrives with HTTP 200, in two situations:
  1. **No token was sent** (`@isAuthorized` and the resolvers' `user == nil` checks). Then there is no cookie to clear.
  2. **A logged-in user asks for something that is not theirs.** `AddMediaShare` and `AddAlbumShare` (`share_token_actions.go:34`, `:80`) return `auth.ErrUnauthorized` when the media or album is not the user's. There the string match was actively wrong: it cleared the cookie of a valid session.

So the session is now judged by the HTTP status alone (401/403). A GraphQL `unauthorized` is still shown as an error, but no longer logs anyone out. The test that pinned the old behaviour now asserts the opposite, and it fails against the previous code.

**One case where the status code itself is wrong, API side:** in `auth.Middleware`, a database error while loading the token's user also returns **401** (`auth.go`, the `err != nil` branch after `UserFromAccessToken.Load`). A database hiccup therefore logs every active user out, which is exactly what this PR is about, just from the other end. It should be a 500. That is a Go change, so I'd keep it out of this UI PR and open a small separate one, unless you'd rather have an issue first.

### `SubscriptionsHook`: yes, a fixed key does it

All errors of the notification subscription now use one key, and each new one replaces the previous. A stream that fails on every reconnect shows one message instead of a growing stack. `MessageState.add` already replaces by key, so the hook just uses the same pattern. New test, which fails with the random key restored.

### i18n CI check

I'll open it as a separate PR: a job in the `Tests` workflow. i18next-parser 8.x, which the repo pins, has `--fail-on-update` and `--fail-on-warnings`, so it can do a bit more than a `git diff` after extraction. I'll check what those flags catch in practice and describe it in the PR.

### Bot comparison

Understood: I used its findings, not its verdict.

### Your question

No, I am not a citizen of any of the countries named in that license. I'd just like to keep the focus on the PRs themselves.

try-and-succeed:
@coderabbitai review

ui/src/components/messages/SubscriptionsHook.ts
```diff
@@ -35,6 +35,28 @@ export interface Message {
...
+export const withSubscriptionError = (
+  messages: Message[],
+  error: Error
+): Message[] => [
+  ...messages.filter(m => m.key != SUBSCRIPTION_ERROR_KEY),
+  {
+    key: SUBSCRIPTION_ERROR_KEY,
+    type: NotificationType.Message,
+    props: {
+      header: 'Network error',
```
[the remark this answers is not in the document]
try-and-succeed: Fixed in `ad7d8c6`. The header now goes through `i18n.t('notification.error.network.header', 'Network error')`, the key the error link already uses for its network error, so every bundle has it and nothing new needed extracting.

I didn't add a test that asserts the translation lookup. On this PR the maintainer preferred an extractor check over unit tests per translated line, and that check is now #1503. The existing test still covers the one-message-per-key behaviour, with the English default as the header.

PR #1487, opened 2026-09-13
Add a show/hide toggle to password fields

try-and-succeed:
to all five.

### Password fields are write-only

Every password field in the app can only be typed into, never read back. A typo can only be found
by clearing the field and typing the whole thing again — on a phone keyboard, an expensive guess.

A show/hide toggle now sits **inside** the input, so it costs no layout and does not move anything
around it. It is a real `<button>` with an accessible name that flips between "Show password" and
"Hide password", and it takes `mousedown` away from the browser so revealing the text does not blur
the field mid-typing. Where a field already has an action button, the eye sits beside it rather
than replacing it.

The revealed state is reset when a field stops being a password field, so a component that swaps
its `type` cannot leave a password on screen.

### Testing

- `npm test` — four new tests: reveal and hide again, no button on other field types, the reset on
  type change, and the coexistence with an existing action button

was verified locally.

## Summary by CodeRabbit

- **New Features**
  - Added show/hide password controls to password fields.
  - Password visibility controls remain available alongside actions and while loading.
  - Password fields disable spellcheck, autocorrect, and autocapitalization.

- **Bug Fixes**
  - Passwords are remasked after submission, field-type changes, or when fields are disabled.
  - Preserved field focus when revealing or hiding passwords.
  - Corrected initial setup validation so photo-path errors display for the photo-path field.


ui/src/primitives/form/Input.tsx
```diff
@@ -79,33 +99,52 @@ export const TextField = forwardRef(
           />
         </div>
       )
-    } else if (action) {
+    } else if (action || isPassword) {
```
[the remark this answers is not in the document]
try-and-succeed: Fixed in `1abaeed`. You are right that `revealed` survives the loading state, so a revealed password stayed on screen with no way to put it away — and `Sharing.tsx` keeps the field enabled while saving, so the window is real rather than theoretical.

The reveal button and the action button now share one trailing cluster, and only the action button is replaced by the spinner. The field's right padding follows how many controls are actually there instead of being spelled out per combination, which as a side effect gives a plain loading field the padding it never had (the spinner used to sit on top of the text).

Regression test added for exactly the sequence you describe: reveal, then re-render as loading, then hide again. Reverting the render change fails it and leaves the other four green.

try-and-succeed:

ui/src/primitives/form/Input.tsx
```diff
@@ -27,20 +29,38 @@ export const TextField = forwardRef(
...
+    const isPassword = type === 'password'
+    const [revealed, setRevealed] = useState(false)
+    const effectiveType = isPassword && revealed ? 'text' : type
+
+    // A field that stops being a password one and later becomes one again
+    // would otherwise come back revealed, showing the secret unasked.
+    useEffect(() => {
+      if (!isPassword) {
+        setRevealed(false)
+      }
```
kkovaletp, line 52: @try-and-succeed, are you sure that the logic here should not be reversed? If the field is not a password anymore, set its revealed state to false (mask the value) - that looks wrong to me

kkovaletp:
@try-and-succeed, thanks for the UX improvement.
Please take a look at the bot's comment under the collapsible in https://github.com/photoview/photoview/pull/1487#pullrequestreview-5191110097 and put a reaction on that message once you process it.
Also, please answer my review question.
Finally, can you please share a screenshot of the input with both buttons? It looks consistent in the code, but let's make sure it is rendered as we expect.

A side thought: a password field is not only protected from exposing the value on screen, but also from copy-paste and reading it from the DOM. Switching the field type from secure password to plain text exposes the password to all potential listeners, which might cause a password leak. I didn't investigate this case, but there should be a solution for that... Maybe a lib or a better way of showing the value on screen while still protecting it from other types of exposure. Can you research that as well?

try-and-succeed:
On the collapsed finding in https://github.com/photoview/photoview/pull/1487#pullrequestreview-5191110097 (assert focus retention): **agreed and fixed** in `25aa60d`. The new test focuses the field and checks it keeps focus after both revealing and hiding; it fails if the toggle's `mousedown` handler is removed, which is exactly the regression it was invisible to before.

ui/src/primitives/form/Input.tsx
try-and-succeed: Yes, it is meant this way round — but the comment made it read as if it were inverted, so I have rewritten it in `35d47c9`.

`revealed` never masks or unmasks anything on its own; it only has an effect *while the field is a password field*, where it is what turns the type into `text` (`effectiveType = isPassword && revealed ? 'text' : type`). A field whose `type` is no longer `password` shows its value in the clear regardless of `revealed`.

So the effect is not masking a non-password field. It clears a flag that has just become meaningless, so that if the same field later becomes a password field again it arrives masked instead of already revealed. The `a field that stops being a password comes back hidden` test pins exactly that sequence: reveal, switch to `text`, switch back to `password`, expect `type="password"`.

try-and-succeed:
@kkovaletp — all four points done in `35d47c9`.

**The collapsed findings**: both processed and reacted to. The one you linked (focus retention) was fixed in `25aa60d`; a second one arrived on that commit — a revealed password could not be hidden once the field was *disabled*, which `PasswordProtectedShare` does while its request runs — and is fixed here too.

**Your review question** is answered on its thread: the logic is intended, the comment made it read as inverted, and I have rewritten the comment.

**Screenshots** — the protected-share page, taken from this branch, masked and revealed:

| | Masked | Revealed |
|---|---|---|
| Desktop | masked | revealed |
| Phone | masked | revealed |

Both controls sit inside the field without overlapping the text, and the focus ring stays on the input after toggling.

### What switching to `type="text"` actually exposes

Good question, and the answer turned out to be more specific than "the DOM":

- **Reading the value from the DOM or from listeners is not affected.** An input's `.value` is readable by any script on the page whatever its type, and key and input events fire the same way for both. `type="password"` protects against the screen, not against code running in the page — so there is nothing a type toggle gives away there that was not already reachable.
- **Copy and cut** are blocked by browsers for password fields and allowed once it is text. That is the user copying their own password, deliberately, from a field they chose to reveal; I left it alone.
- **Spellcheck is the real leak.** Chrome's Enhanced Spellcheck and Edge's Microsoft Editor send the contents of text fields to Google and Microsoft, and researchers showed in 2022 that this includes passwords revealed through "show password" buttons ([BleepingComputer](https://www.bleepingcomputer.com/news/security/google-microsoft-can-get-your-passwords-via-web-browsers-spellcheck/), [Dark Reading](https://www.darkreading.com/application-security/spellchecking-google-chrome-microsoft-edge-browsers-leaks-passwords)). The recommended mitigation is `spellcheck="false"`. **Done**: password fields now always carry `spellcheck="false"`, `autocorrect="off"` and `autocapitalize="off"`, applied after the caller's props so they cannot be switched back on.
- **Form history** is the other one: browsers may remember a submitted text field's value as an autocomplete suggestion. GOV.UK's password input handles this by switching the field back to `password` when its form is submitted ([GOV.UK Design System](https://design-system.service.gov.uk/patterns/passwords/), [their write-up](https://technology.blog.gov.uk/2021/04/19/simple-things-are-complicated-making-a-show-password-option/)). **Done** the same way — and the type is set on the element inside the submit event, because waiting for React to re-render from state would only happen after the browser has looked at the field; the test fails without that.

I did not find a library that does fundamentally better: the established implementations (GOV.UK, NHS, and the password inputs in the common React component libraries) all toggle the input type, and rendering the value somewhere else instead would still put it in the DOM. The two measures above close the gaps that are specific to the toggle.

Each of the three new tests fails with its change reverted. Also run: `tsc --noEmit` on the touched files, eslint, prettier, build.

ui/src/primitives/form/Input.tsx
kkovaletp, line 1: @coderabbitai, this comment relates to all changes in this PR and not only to this file.
As a final code review step before the merge, please carefully review this PR in the scope of the app, including UI, UX, the testing framework, and business logic.
Do you see any inconsistencies, regressions, performance or security issues, or any other types of issues? Do you want to highlight something else important?

kkovaletp: @try-and-succeed, please include fixes for these findings in this PR, even while those bugs exist in `master`.
The password input on the init setup page was intentionally plain text because we didn't have the reveal logic and wanted to avoid typos that could prevent admins from accessing the just-deployed instance. But now, we can safely switch it to password type.
The 2nd bug is an escaped one, but its fix should be small and straightforward, so let's fix it as well.

try-and-succeed: Done in `1e5151c`, both findings included here as asked.

**Initial setup password** is a password field now, so it gets the reveal button and the spellcheck and form-history safeguards with it — and thanks for the background on why it was plain text; the reveal button covers exactly that typo case.

**Photo path error**: the field read its required error from `formErrors.password`, so an empty path showed nothing and an empty password showed the path's message under the wrong field. It reads `formErrors.rootPath` now.

Two new tests in `InitialSetupPage.test.tsx` — the password starts masked and can be revealed, and with only the path left empty exactly the path's error appears — and both fail without their fix. Also run: eslint, prettier, build, and `tsc --noEmit` on the file (no new errors; the two existing `mockImplementation(() => null)` ones are upstream's).

kkovaletp (approved):
LGTM

PR #1488, opened 2026-09-13
Add a native share button to the media sidebar, and check the download response

try-and-succeed:
to all five.

### The sidebar can only download

On a phone, downloading is the wrong verb: what people want is to hand the picture to another app.
A share button now sits next to the download, using the Web Share API — the file itself where the
browser supports sharing files, the link otherwise. Where neither is available the button is not
rendered at all.

### The existing download never checked the response

`fetch` resolves for a 404 or a 500 just as happily as for a 200, so a failed request was written
to disk as a file containing an error page. Both the download and the share's file preparation
check `response.ok` now.

### Two failure modes worth naming

- **Preparing the file fails.** For anything other than the user aborting, the link is shared
  instead, so a bad network degrades rather than dead-ends.
- **The download outlasts the browser's transient user activation.** The share sheet then refuses
  to open, and by the time the file is ready there is no activation left to spend. The prepared
  file is kept from the moment it exists and the button says "Tap share again", so the second tap
  skips the download — on the file branch and the link branch alike. The hint sits on the button
  because the sidebar's message area is hidden below 1000px, and native share is a phone feature.

### Testing

- `npm test` — three tests covering the file share, the link fallback for a browser that cannot
  share files, and the link fallback when the download itself fails. Reverting the ordering of the
  kept file fails the link test and leaves the file one green.
- `npm run build`, `eslint` on the changed files

The translation files carry only the two new keys. The unrelated extractor drift in the Ukrainian
bundle that #1483 picked up is deliberately left out of this PR.

was verified locally.

## Summary by CodeRabbit

* **New Features**
  * Added per-variant share actions to the media download table.
  * Added link-sharing fallback and retry guidance after sharing failures.
  * Added dedicated, keyboard-accessible buttons for media and album download options.

* **Bug Fixes**
  * Prevented failed downloads and error responses from being saved as media.
  * Improved filename handling and removed query strings and fragments from downloaded filenames.
  * Avoided repeated sharing attempts after successful fallback sharing.

* **Tests**
  * Expanded coverage for sharing, retries, cancellations, downloads, and accessible table interactions.

* **Localization**
  * Organized download and sharing labels under grouped action translations.


ui/src/components/sidebar/SidebarDownloadMedia.tsx
```diff
@@ -246,6 +274,133 @@ const SidebarDownloadTable = ({ rows }: SidebarDownloadTableProps) => {
...
+
+      // The file couldn't be prepared or the sheet refused it for some other
+      // reason, but the link is still shareable, so try that rather than
+      // ending in nothing.
+      try {
+        await shareLink()
+      } catch (fallbackErr) {
+        if ((fallbackErr as Error)?.name !== 'AbortError') {
+          console.error('Link share failed too', fallbackErr)
+        }
```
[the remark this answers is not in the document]
try-and-succeed: Fixed in `2750c32`. Confirmed: the retry from the earlier round only covers the two branches that reach a prepared file, and this is the third one — the download fails outright, so there is nothing to keep, and if it failed slowly the link share standing in for it has no activation left either. The nested catch only logged, so every further tap paid for the same failing download before ending in the same nothing.

There is a `linkOnly` flag now: when the fallback link share is refused with `NotAllowedError`, the next tap goes straight to the link inside its own activation, no fetch.

One thing I added beyond the suggestion: both halves of the retry state are cleared through a single helper. An aborted link retry would otherwise have left the flag set, and later taps would have kept sharing the link even once a file share was possible again — the same stale-state shape as the bugs the earlier rounds found here.

Test added with the sequence you asked for (rejected fetch, rejected first link share, second tap with no fetch, and the button dropping the retry wording afterwards). Reverting the flag fails it and leaves the other three green.

ui/src/components/sidebar/SidebarDownloadMedia.tsx
```diff
@@ -246,6 +274,161 @@ const SidebarDownloadTable = ({ rows }: SidebarDownloadTableProps) => {
...
+    const shareLink = () =>
+      navigator.share({
+        title: media.title ?? undefined,
+        url: location.href,
+      })
+
+    try {
+      if (!navigator.canShare || linkOnly.current) {
+        await shareLink()
+        clearRetryState()
```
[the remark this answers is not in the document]
try-and-succeed: Fixed in `d9d3d64`. Confirmed — on the link path the outer catch treated the failure as a file-share failure and fell back to the link, which is the same call again on an activation that is already spent: one tap, two `share()` calls, and on a browser without `canShare` the first error logged as "native share failed" when no file was ever involved.

That branch handles its own failure now and arms the retry itself, which is what the accidental second call was doing by chance. Took the shape of your diff, with the error name hoisted into a local since the catch now checks it three times.

New test counts the calls: one per tap, no fetch on either, and the second tap goes through. It fails without the change and the other four stay green.

try-and-succeed:

ui/src/components/sidebar/SidebarDownloadMedia.tsx
try-and-succeed, line 1: @coderabbitai, I'll write this comment here, but it is relevant for all changes in this PR, not only this file. As a final review step before asking the maintainers to merge, please carefully analyze this PR within the scope of the whole app, including business logic, UX, the testing framework and other relevant aspects.

Context: this PR adds a native share button to the media sidebar using the Web Share API (file where supported, link otherwise), adds the missing `response.ok` check to downloads, and handles a download that outlasts the browser's transient user activation with a kept file or link and a second-tap retry. Pay particular attention to: every combination of `navigator.share`/`canShare` support, failed or slow downloads and user aborts; that no path shares twice per tap or leaves stale retry state; share pages viewed through a share token; and memory held by a kept file.

Please answer:

try-and-succeed: Thanks — the coverage gaps are closed in `e204021`. Each new test fails when the logic it covers is removed:

- a non-success response never reaches the saved file, and fails the share's preparation instead of sharing an error page;
- on `/share/:token`, media are fetched with the token for an anonymous visitor and without it for a logged-in user;
- closing the share sheet (`AbortError`) ends the share with no retry wording and no link fallback;
- no share button without `navigator.share`, or for a media without downloads.

kkovaletp:
@try-and-succeed, thanks for the contribution.
Can you please add screenshots for the sidebar with this functionality on a desktop and mobile device?
A media typically has several download options. Is this functionality added to all possible downloads?

ui/src/components/sidebar/SidebarDownloadMedia.tsx
```diff
@@ -246,6 +274,177 @@ const SidebarDownloadTable = ({ rows }: SidebarDownloadTableProps) => {
...
+      }
+    } finally {
+      setSharing(false)
+    }
+  }
+
+  return (
+    <div className="pl-4 py-2">
+      <button
+        className="text-green-500 font-bold uppercase text-xs disabled:opacity-50"
```
kkovaletp, line 440: @try-and-succeed, don't we need different classes for the white and dark themes?

ui/src/components/sidebar/SidebarDownloadMedia.tsx
```diff
@@ -246,6 +274,177 @@ const SidebarDownloadTable = ({ rows }: SidebarDownloadTableProps) => {
...
+        await shareLink()
+        clearRetryState()
+        return
+      }
+
+      let file = preparedFile.current?.file
+      if (preparedFile.current?.url !== row.url) {
+        const blob = await fetchMediaBlobQuiet(row.url)
+
+        const filename = row.url.match(/[^/]*$/)?.[0] ?? media.title ?? 'photo'
```
kkovaletp: @coderabbitai, as we had many issues in the past caused by non-optimal regex, can you please double-check and confirm that this regex here in the context of the function's business logic does exactly what is expected and in an optimal way?

kkovaletp: @try-and-succeed, would you like to replace the regex with the suggested URL parser?
If it does what we need with comparable efficiency, it makes the code more readable, at least.

try-and-succeed:
Thanks for the review! Screenshots below, and an answer to the download question.

### Screenshots

Desktop, light theme — the button sits under the download table:

desktop light

Mobile (390 × 844), light theme:

mobile light

Desktop, dark theme (relevant for your theme question on the button's classes):

desktop dark

One disclosure about how these were taken: Chromium on Linux exposes no Web Share API, so the button
would not render at all there. I stubbed `navigator.share`/`navigator.canShare` for the screenshots
only — the layout is real, the share sheet in them is not. What happens on a real tap is covered by
the tests in `SidebarShareMedia.test.tsx` (file share, link fallback, retry after a lost activation,
abort).

### "Is this functionality added to all possible downloads?"

Two different answers, depending on which half of the PR you mean:

**The `response.ok` check: yes, all of them.** Every row in the download table goes through the same
`downloadMedia`/`downloadMediaShowProgress` helper, and that is where the missing check was added.
So Original, Small, Large, Video thumbnail and Web optimized video are all covered.

**The share button: no, there is one per media, not one per row.** It picks a single variant, in this
order: `Web optimized video`, else `Original`, else `Large`, else the first row
(`pickShareRow`). The reasoning: sharing is "send this photo to someone", so the variant should be
the best one the server has, and for a video the web-optimised MP4 is the one other apps can play.
A share icon in every row would also compete with the row's own click, which starts the download.

If you would rather have it per row, I am happy to change it — say the word and I will put a share
action in each row instead, with the table's click behaviour kept intact.

One note on the test library while we are here: its videos have no `Web optimized video` entry (the
sources are already web-compatible, so the server keeps `Original`), which is why you see only
`Original` and `Small` in the screenshots above.

ui/src/components/sidebar/SidebarDownloadMedia.tsx
try-and-succeed: Good question, and I measured it rather than guessed. The project's `green-500` is `#4abf3c`:

| | contrast |
|---|---|
| on white (light theme) | **2.38:1** |
| on `dark.bg2` `#30363e` (sidebar in dark theme) | **5.11:1** |
| on `dark.bg` `#24292e` | 6.15:1 |

So the dark theme is the *better* of the two here, and a `dark:` variant would not make it more
readable — it is the light theme that falls short of WCAG AA (4.5:1 for this 12px bold text). The
dark screenshot in my other comment shows the button in that theme.

That weak light-theme value is not specific to this button: `text-green-500` without a `dark:`
variant is what the two neighbouring sidebar actions use as well — `SET AS ALBUM COVER PHOTO`
(`AlbumCovers.tsx`) and `ADD SHARES` (`Sharing.tsx`), both visible right below the share button in
the mobile screenshot. I kept this PR consistent with them rather than making one button look
different.

If you want it fixed, `text-green-700 dark:text-green-500` gives 4.85:1 on white and 5.11:1 on the
dark sidebar, so both themes pass AA. I would rather do that for all three buttons in its own PR
than change one of them here — happy to open it if you agree.

ui/src/components/sidebar/SidebarDownloadMedia.tsx
try-and-succeed: Yes — done in `38311f1`, and the bot's point about the query string is a real defect, not only a
readability question.

The filename now comes from the URL parser:

```ts
const filenameFromUrl = (url: string) => {
  try {
    return new URL(url, location.origin).pathname.split('/').pop() || undefined
  } catch {
    return undefined
  }
}
```

Two things came out of it:

- A url carrying a share token produced `holiday.jpg?token=abc` as the shared file's name, and the
  extension shown in the table's `Type` column had the same problem in principle. Both go through
  the one helper now, so the `Type` column no longer needs its own hand-written `?`/`#` stripping.
- The new test in `SidebarShareMedia.test.tsx` shares a row whose url is
  `photo/holiday.jpg?token=abc#preview` and asserts the file is named `holiday.jpg`. I checked that
  it fails with the old expression (`expected 'holiday.jpg?token=abc#preview' to be 'holiday.jpg'`),
  so it pins the fix down rather than just passing.

`new URL(..., location.origin)` also takes both the relative urls the API returns today and an
absolute one, should that ever change.

try-and-succeed:
@coderabbitai review

try-and-succeed:
@coderabbitai full review

kkovaletp:
@try-and-succeed 

> **The share button: no, there is one per media, not one per row.** It picks a single variant, in this order: `Web optimized video`, else `Original`, else `Large`, else the first row (`pickShareRow`). The reasoning: sharing is "send this photo to someone", so the variant should be the best one the server has, and for a video the web-optimised MP4 is the one other apps can play. A share icon in every row would also compete with the row's own click, which starts the download.
> If you would rather have it per row, I am happy to change it — say the word and I will put a share action in each row instead, with the table's click behaviour kept intact.

Yes, I meant this case. We cannot know the motivation of each user and the purpose of sharing something, so I'd not assume the best media for each case. Photoview provides download functionality for all available variants on purpose, letting users decide what they need to take this time. So, I'd prefer having the share button per row.
You've highlighted a good point that currently a whole row processes clicks and behaves as a button, initiating a download process. Putting a new share button inside this row would introduce a conflict and make the UX even more counter-intuitive. Do you have a suggestion for this UX issue?
I thought about removing the click handler from a row and putting explicit download and share buttons in it. This approach resolves the previously highlighted issue, but creates the following one:

- Not enough width to add 2 full-size buttons. 

Can be solved by putting buttons with only icons and keeping their names only in tooltips/aria-labels. These buttons would be small, so we shouldn't put them 1 after another to prevent misclicks/mistaps. Maybe place them on both sides of the Name column, for example: `[Dl] Original [Sh]`...

Also, we have the permanent direct links functionality, represented by the "Sharing options" table below. Similar names for different functions would create another UX issue, even if the new button will have only the icon, as it is still going to have a tooltip and aria-label. Please think about the naming solution.

Finally, the same approach needs to be applied to the Album download table as well: an album > gear icon > Download table on the sidebar.

ui/src/components/sidebar/SidebarDownloadMedia.tsx
kkovaletp: @try-and-succeed, sounds good to me to fix this issue globally for all styled elements in the sidebar as a new PR. However, the current implementation is good enough too, so we can go with it as a minimal solution as well.

try-and-succeed:
Understood — per row it is. Here is what I would build, before I build it, plus a naming proposal.
And the two findings from the latest bot review are fixed in `43ed79d` (details on their threads).

### The row-as-a-button problem

I would rather not add a second small target to a row that is itself a button — but I also think the
row being a button is the actual problem. It is a `<tr>` with `onClick` and `tabIndex`, so a screen
reader announces a table row, not an action, and there is no way to tell which part of it does what.

So: **the row stops being clickable, and the name becomes the download button.**

```
 NAME                     DIMENSIONS     SIZE     TYPE
 [ Original ]             1200 x 1600    96 KB    jpg      [↗]
 [ Small ]                768 x 1024     25 KB    jpg      [↗]
   ^ a real <button>, label = the variant name            ^ icon-only, "Send"
```

- The download target stays as big as it is today (the whole name cell), and it becomes a real
  button instead of a row pretending to be one.
- The send action is icon-only in a narrow trailing column, so no extra width is needed for a second
  label — which was your width concern.
- The two targets sit at opposite ends of the row, as you suggested, so a mistap hits neither.
- Both get a tooltip and an `aria-label`; the icon button also gets `title` for hover.

If you would rather have your symmetric `[Dl] Original [Sh]` version with two icons, that is fine
too — say so and I will do that instead.

### Naming

The clash you are pointing at is real: "Share" for the OS share sheet versus the "Sharing options"
table below, which is about permanent public links. I would keep "Sharing options" untouched and
avoid the word "share" for the new action entirely:

| | label (tooltip + `aria-label`) | i18n key |
|---|---|---|
| download button | `Download Original` | `sidebar.download.actions.download` |
| send button | `Send Original to another app` | `sidebar.download.actions.send` |

"Send … to another app" is what the action actually does, it never collides with the link sharing
below, and it also reads correctly on a desktop browser where the sheet offers mail or a messenger.
Alternatives if you prefer: `Share via…` (closer to the OS wording, but keeps the clash) or
`Open in another app` (weaker — it is a copy, not an open).

### The album download table

Same pattern for the name-as-button and the removed row click, yes. On a send button there, I would
argue **no**, and I want your opinion before I skip it:

- Those four rows are server-generated **zip archives** of a whole album
  (`/download/album/<id>/<purpose>`), fetched today by simply setting `location.href`. To hand a zip
  to the share sheet it has to be downloaded into memory as a blob first — for a large album that is
  hundreds of megabytes to gigabytes in the tab, and the Web Share API will usually refuse a file
  that size anyway.
- The album already has the "Sharing options" table below it for handing the album to someone.

So my proposal: album table gets the accessibility fix (name as a real button, row no longer a
button) and no send action. If you want one anyway, the honest version is a send that shares the
album's **link**, not the archive — happy to add that instead.

Once you have picked the layout and the naming, I will implement it, add the tests, and post fresh
desktop and mobile screenshots of both tables.

kkovaletp:
@try-and-succeed,

> So: **the row stops being clickable, and the name becomes the download button.**

Agree; this design looks even more solid and user-friendly.

> * The send action is icon-only in a narrow trailing column, so no extra width is needed for a second
>   label — which was your width concern

Well, technically speaking, even an icon-only button needs some space. The difference is that it's much less than a labeled button, and we already have it. Taking into account that there are cases when it is not rendered, and that the text in the table's cells can have different lengths, please make sure that the total row width and each column width are properly managed, and that we don't have a corrupted table layout or "jumping" content.

> "Send … to another app"

Agree; this naming looks consistent and self-explanatory.

> Same pattern for the name-as-button and the removed row click, yes. On a send button there, I would
> argue **no**

Agree. Let's not add the Send functionality there.

Once everything is implemented, please share a new set of screenshots.

try-and-succeed:
@kkovaletp, done in `3d37ea6` — per-row send, as agreed.

### What changed

- **Media table:** the row is no longer a button. The name cell is the download button (`Download Original`) and fills its cell, so the download target is as large as before. The send button (`Send Original to another app`) sits in a narrow trailing column. Both have an `aria-label` and a `title`.
- **Album table:** same pattern (`Download Originals`, …), no send button.
- **`pickShareRow` is gone.** Each row sends its own variant.
- **Tap again:** a slow download can use up the share sheet's user activation, so a second tap is sometimes needed. The send button of that row gets a ring, and a line below the table names the variant (`Tap again to send Original to another app`). It sits below the table so no column has to make room for it.
- The download now also takes its filename from the URL parser, like the share and the Type column already did. The old `/[^/]*$/` regex is gone from this file.

### Layout

`table-fixed`, with fixed widths for Dimensions (1/4), Size (1/6), Type (1/6) and the send column (3rem). The name column takes the rest. The send column is on every row or on none, header included, so a browser without `navigator.share` just gets a wider name column. Long names wrap inside their cell instead of pushing the other columns.

Measured in Chromium on every screenshot below: each row has the same column widths as the header (e.g. `115,98,65,65,48` px at 390 px width, `127,105,70,70,48` on desktop), no cell overflows, the table is exactly as wide as the sidebar, and `elementFromPoint` at the centre of every button returns that button. The only height change is the wrapped `Video thumbnail` row (64 px instead of 40).

### Screenshots

As before, `navigator.share` is stubbed, because Chromium on Linux has none. #5 removes it to show the table without the column. #6 stubs a first share that is refused with `NotAllowedError`.

| | |
|---|---|
| 1. Desktop, light |  |
| 2. Mobile 390×844, light |  |
| 3. Desktop, dark, hovering `Original` |  |
| 4. Mobile, video with a wrapping variant name |  |
| 5. Mobile, browser without share support |  |
| 6. Mobile, second tap needed |  |
| 7. Mobile, album table |  |
| 8. Desktop, album table |  |

### Tests

The existing share tests now drive the per-row button. New ones cover:

- each row sending its own variant;
- the name button downloading its variant;
- rows no longer being focusable;
- the column being absent, header included, without `navigator.share`;
- the tap-again line;
- the album table's buttons.

Verified locally: full UI suite (only the 2 known `MediaSidebarExif` failures, which fail on master too), eslint, prettier, `tsc --noEmit` clean for the touched files, `npm run build`. The two old translation keys (`share`, `share_again`) are replaced by `sidebar.download.actions.*` in all bundles.

try-and-succeed:
@coderabbitai review

PR #1460, opened 2026-08-02
fix: prevent browser back/forward navigation on horizontal swipe in gallery view

waterWang:
## Description

On touch devices (tablets, phones), horizontal swiping in the gallery view could trigger the browser's built-in back/forward navigation gesture, especially when pinch-to-zoom is used. This disrupts the user experience and makes gallery navigation unreliable.

## Changes

1. **`overscroll-behavior: none`** added to the gallery container CSS — prevents the browser from initiating back/forward navigation when the user swipes horizontally beyond the edge of the content.

2. **`preventScrollOnSwipe: true`** set on the `react-swipeable` handler — prevents page scrolling during swipe gestures, ensuring swipe navigation works smoothly.

## Testing

- All existing gallery tests pass (3/3 in PresentNavigationOverlay.test.tsx)
- The fix is minimal: only 2 lines changed in 2 files

## Related

Fixes #1380

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved photo gallery swiping to prevent unintended page scrolling.
  * Disabled overscroll behavior while viewing photos for smoother navigation.
  * Updated photo sharing to refresh automatically when switching between photos.

PR #1489, opened 2026-09-13
Reach the presentation controls on touch devices, and the photo's details from there

### LINKED ISSUE #546: Image reopening depending on method closed.
**Describe the bug**
If you close an image/video by pressing the 'X' in the upper left, then later close an image by pressing escape. Any image prior that was closed by pressing the 'X' will be reopened. 

**To Reproduce**
1. Open an image/video
2. Click on 'X' in upper left
3. Open another image/video
4. Press 'escape'
5. Previous closed images will be opened.

**Expected behavior**
Closing with X or Escape regardless of order should yield same behavior

**Your environment:**

 - Are you using Docker? Yes
 - Database Postgres
 - Browser Microsoft Edge
 - Version 2.3.6 (unable to test on 2.3.8 due to a different issue)

**Additional context**
If you close many objects by pressing the 'X' it basically creates a queue for when you press escape. e.g. I open and close 5 different images with the X icon. I then open a 6th object and press escape, I now have to cycle through those 5 previously closed objects by pressing the x.

Oddly enough this does not behave the other way around and is specific to close with x, then close with escape.

try-and-succeed:
Fixes #546 — closing the viewer with the X button left its history entry behind, so a later Escape (or the browser's back button) reopened previously closed photos.

The code is the version that came out of that PR's four review rounds, plus the focus-ring fix
CodeRabbit asked for in the last one; #1483 is closed and links to all five.

### Presentation mode is a dead end on touch devices

The overlay only reveals its controls on `mousemove`, which a touch device never fires — so on a
phone the fullscreen viewer has no visible way back out, no navigation and no indication of where
you are. A tap now reveals them on the same auto-hide timer, and so does any key press.

While hidden, the buttons are properly gone rather than merely transparent: `visibility: hidden`
means the first tap reveals the controls instead of firing whichever button sits under the finger,
and a keyboard user cannot tab to a control they cannot see. It also takes them out of the
accessibility tree. The delay lets the existing fade finish; revealing is instant, as no transition
applies in that direction.

### No way to see a photo's details from there

An info button opens a panel with the same EXIF data the sidebar shows, which on a phone is where
most viewing happens.

### Keyboard focus stays visible

`OverlayButton` sets `outline: none`, so once a key press could reveal the controls, a keyboard
user could focus a button with no indication of it. A `:focus-visible` rule restores a white ring
with a thin dark edge, which stays legible over a bright photo, and brightens the icon the same way
hover does.

### Testing

The touch and keyboard behaviour was exercised by hand while the work was done in #1483. The
`:focus-visible` rule added here is CSS only and has not been re-exercised in a browser since —
worth a glance when you try the branch.

was verified locally.

## Summary by CodeRabbit

* **New Features**
  * Present-mode navigation controls can be revealed with keyboard input or a touch tap.
  * Added touch-swipe navigation between media items.
  * Added an optional “Show media info” button for opening details in the shared sidebar.
* **Accessibility**
  * Hidden controls are excluded from keyboard and touch navigation when unavailable.
  * Added visible focus outlines for navigation buttons.
* **Bug Fixes**
  * Improved touch interaction and scrolling behavior in present mode.
  * Media information stays synchronized with the currently displayed item.
  * Improved media sizing within the presentation frame.
  * Exiting present mode now preserves browser history behavior.


ui/src/components/photoGallery/presentView/PresentNavigationOverlay.tsx
```diff
@@ -95,10 +132,26 @@ const PresentNavigationOverlay = ({
     }
   }, [])
 
+  useEffect(() => {
+    // Hidden controls are taken out of the tab order entirely, so a keyboard
+    // user has no way to reach them - and no pointer to move. Any key press
+    // reveals them, the same as a mouse movement or a tap, which then puts
+    // them back within reach of Tab.
+    const revealOnKey = () => {
+      onMouseMove.current && onMouseMove.current()
```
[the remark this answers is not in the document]
try-and-succeed: Not changing this one, because the premise does not hold for this `debounce`: it always schedules the trailing call.

```js
const debounced = (...args) => {
  if (timeout) { clearTimeout(timeout); timeout = undefined }
  else if (triggerRising) { func(...args) }
  timeout = window.setTimeout(() => { timeout = undefined; func(...args) }, wait)
}
```

So a single key press or tap invokes the toggle on the rising edge (reveal) **and** arms a 2s timer whose trailing edge invokes it again (hide). Rising and trailing edges therefore strictly alternate — every rising invocation is followed by exactly one trailing invocation, and a call inside the window only clears and re-arms the timer without toggling. The phase cannot drift, which is why the pre-existing mouse path has always worked this way; the tap and key handlers deliberately reuse it rather than introducing a second timer that would have to be kept in sync with it.

Rather than argue it from reading, I added tests in `9b7069f` (the overlay had none at all):

- a key press reveals; 2s later the controls hide themselves; the next press reveals again rather than toggling back the other way
- a tap reveals, and taps inside the window push the auto-hide back instead of hiding the controls under the finger reaching for them

Both pass on this branch, and they would catch the drift this finding describes if it were real. Happy to reconsider if you can point at a sequence where two rising edges land without a trailing one between them.

try-and-succeed:

ui/src/components/photoGallery/presentView/PresentNavigationOverlay.tsx
```diff
@@ -95,10 +132,26 @@ const PresentNavigationOverlay = ({
...
+    return () => {
+      document.removeEventListener('keydown', revealOnKey)
+    }
+  }, [])
+
   const handlers = useSwipeable({
     onSwipedLeft: () => dispatchMedia({ type: 'nextImage' }),
     onSwipedRight: () => dispatchMedia({ type: 'previousImage' }),
-    preventScrollOnSwipe: true,
+    preventScrollOnSwipe: false,
```
kkovaletp: @try-and-succeed, this property was recently modified as part of the #1460 PR. Please confirm that this is not a regression, that this change is consistent with all other changes in this PR, and made on purpose.

ui/src/components/photoGallery/presentView/PresentView.tsx
```diff
@@ -14,11 +16,13 @@ const StyledContainer = styled.div`
   top: 0;
   left: 0;
   z-index: 100;
-  overscroll-behavior: none;
```
kkovaletp, line 17: @try-and-succeed, this property was recently added as part of the #1460 PR. Please confirm that this is not a regression, that this removal is consistent with all other changes in this PR, and made on purpose.

ui/src/components/photoGallery/presentView/PresentNavigationOverlay.tsx
try-and-succeed: Good catch — it is a regression, and not intentional. Fixed in `c0f39ef`.

`preventScrollOnSwipe` was `false` before #1460 and `true` after it. The branch these changes were written on predates #1460, and when I split it into this PR I copied whole files across rather than the individual hunks, so it carried the older value and silently reverted your fix. Same story one file over, where `overscroll-behavior: none` had been dropped.

Both are back at their upstream values. Nothing in this PR depends on either: the swipe handlers and the tap/keyboard reveal are separate handlers, and preventing the scroll gesture does not affect a tap.

I checked the rest of the split for the same mistake — #1460 is the only upstream commit that landed between that branch's base and the ones I created, and its third file (`Sharing.tsx`) is not touched by any of my PRs. So this was the only place it could have happened, and it is now the only place it did.

ui/src/components/photoGallery/presentView/PresentView.tsx
try-and-succeed: Same cause as the other one, fixed in the same commit `c0f39ef`: `overscroll-behavior: none` came from #1460, and splitting this PR out of an older branch copied the pre-#1460 version of the file over it. Restored.

The styled block otherwise keeps this PR's own change — `PreventScroll` is scoped to `html, body` instead of `*`, so a panel rendered on top of the viewer (the media info sidebar this PR adds) can still scroll its own content.

kkovaletp:
@try-and-succeed, thanks for the UX improvement.
Please answer my review questions and check the code coverage report (just to make sure that the current coverage is optimal and the missed lines are not easily coverable or don't make sense to be covered).

try-and-succeed:
Both review questions answered on the threads, and the coverage report checked — thanks for
pointing at it, it was worth the look.

**The two properties were a regression**, and not a subtle one: `preventScrollOnSwipe` and
`overscroll-behavior: none` both come from #1460, which landed after the branch these changes were
written on. Splitting that branch into this PR copied whole files across instead of the individual
hunks, so it carried the pre-#1460 versions and silently undid your fix. Restored in `c0f39ef`.

I checked whether the same thing happened anywhere else in the split: #1460 is the only upstream
commit between that branch's base and the ones I created, and its third file (`Sharing.tsx`) is not
touched by any of my PRs — so this was the only place it could have happened.

**Coverage**: all eleven missed lines were easily coverable, and none of them were noise — they were
the parts a user actually exercises. Covered in `6dfa109`:

- a horizontal swipe dispatching next/previous, which also exercises the restored
  `preventScrollOnSwipe` setting;
- the exit button in both modes — stepping back in history, and leaving history alone when the
  caller manages it;
- the info panel's two effects: it follows the image you navigate to, and stays shut once the
  sidebar has been closed some other way. Those had no coverage at all, and they are the part of
  this PR most likely to break quietly.

`PresentView.tsx` and `PresentNavigationOverlay.tsx` are both at 100% statements now (locally
measured; the CI report will confirm).

Happy to keep going the same way on the other PRs if that is useful — say the word and I will check
their coverage reports too rather than waiting for you to ask.

ui/src/components/photoGallery/presentView/PresentView.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const updateSidebar = vi.fn()
+  let content: React.ReactNode = null
+
+  const view = render(
+    <SidebarContext.Provider
+      value={{
+        updateSidebar,
+        setPinned: vi.fn(),
+        content,
+      }}
```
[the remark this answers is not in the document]
try-and-succeed: Fixed in `e1b9516` — `pinned: false` added to every test context value, and the two-argument `expect()` calls (which this version of vitest does not have) turned into comments.

Worth recording why this got past me: neither `npm run lint` nor `npm run build` sees it. ESLint's type-aware rules do not check for missing properties, and vite strips types without checking them, so `tsc --noEmit` is the only thing that catches this class of mistake. I have added it to what I run before pushing, and it turned up three more in #1485 — an import of a type the generated file does not export, and a debounced callback that did not satisfy the helper's constraint. Fixed there too.

kkovaletp:
@try-and-succeed, thanks for addressing the findings.
Please remember to look at the bot's comments under collapsibles (like here https://github.com/photoview/photoview/pull/1489#pullrequestreview-5195943131) and put a reaction on those messages once reviewed - this is how I'll know that you saw them.

ui/src/components/photoGallery/presentView/PresentNavigationOverlay.tsx
kkovaletp, line 1: @coderabbitai, I'll write this comment here, but it is relevant for all changes in this PR.
Please perform the final review and carefully analyze this PR within the scope of the app, including the business logic, UX, and other relevant aspects. Do you see any regressions, inconsistencies, or other issues? Are there any risks in merging this PR? Anything else you'd like to highlight?

kkovaletp: @try-and-succeed, please take a look at these findings.

BTW, this is a good example of asking the bot to do the final code review after we think that a PR is ready. Feel free to use this approach in your other PRs. Just make sure to write such a comment for a main file in the PR as a file-level code review comment and explicitly describe what the bot has to do, in which context, and what answers you expect from it.

try-and-succeed: @kkovaletp — all three looked at and fixed in `90bdc57`, each reproduced in a real browser first. Thanks for the pointer to the final-review approach; I will use it on the other PRs.

**1. The info panel was hidden behind the viewer — confirmed, and it was the worst of the three.** The panel is the shared sidebar at `z-40`; the viewer is at `z-100`. "Show media info" changed the sidebar's state and showed nothing. I checked it in Chromium by asking which element sits on top at the panel's position: before the fix it was the viewer, at both 1440 and 390 px wide.

| | Before | After |
|---|---|---|
| Desktop | before | after |
| Phone | before | after |

The viewer's global style now lifts the sidebar above it — only while the viewer is mounted, so its stacking everywhere else is unchanged. The sidebar gets a `data-sidebar` attribute for that rule to target, rather than a utility or generated class name. One trade-off worth naming: on desktop the open panel covers the viewer's "next" arrow; keyboard arrows and swipe still work, and the panel has its own close button.

**2. The exit button never restored history — confirmed, and it predates this PR.** `PresentView` passed `disableSaveCloseInHistory` as a bare attribute, which is always `true`; upstream has the same line. Because the gallery and timeline open the viewer with `history.pushState`, leaving by the button left that entry behind: I verified that pressing the browser's back button afterwards **reopened the viewer**, and that with the fix it leaves the album as expected. Escape was always right. Since this PR adds buttons on exactly that path, I fixed it here by passing the caller's value through — the map view, which passes `true`, is unaffected.

**3. The panel stayed open after leaving the viewer** (the finding under the collapsible — reacted to). It is closed on unmount now, but only if the viewer opened it: a sidebar the user already had open before presenting stays open.

**Tests**: 2 and 3 are covered and fail without their fixes. The stacking is not unit-tested on purpose — jsdom does not apply `createGlobalStyle` rules to its CSSOM, so such a test would check the test environment rather than the page; the browser check above is the evidence instead. Also run: `tsc --noEmit` on the touched files, eslint, prettier, build.

kkovaletp: @try-and-succeed, thanks for fixing the issues and providing detailed descriptions for them!
Regarding issue 1:

> One trade-off worth naming: on desktop, the open panel covers the viewer's "next" arrow; keyboard arrows and swipe still work, and the panel has its own close button.

What if we show the sidebar on big-enough screens not on top of the main image frame with its controls, but by resizing the main image frame horizontally and taking the needed space on the right side for the sidebar - a similar approach to the pinned sidebar on the timeline or album page? I see the pin button of the sidebar in the disabled state in your screenshot, so I'd assume that rendering it with the pinned property by default should work, and this would be the easiest fix. It will allow us to keep all the controls and the full image on screen, while letting the user unpin the sidebar if he/she wants that for some reason.
What do you think?

I just remembered, while reading your description of issue 2, that we have it reported in the Issues for a long time (#546 - please take a look). Good that it is now going to be fixed)

try-and-succeed: Good idea — done in `476476f`, and it works as you expected.

The info panel now opens pinned. On screens from the `lg` breakpoint up, the viewer narrows to the space beside the panel, so the whole photo and all of its controls stay in view:

pinned info panel

Unpinning drops back to the panel overlaying the photo, and on a phone — where pinning has no layout of its own — it overlays as before.

One thing it needed beyond pinning: the photo and video in `PresentMedia` were sized to the *viewport* (`100vw` × `100vh`), not to the frame, so narrowing the frame alone would have left them running underneath the panel. They size to the frame now, which at full width is the viewport, so nothing changes there. Checked in Chromium at 1440 px: pinned, the photo ends exactly at the panel's left edge (x = 1020) and the next arrow is the topmost element at its position; unpinned, it overlays again.

**On #546** — thanks for the pointer, it is exactly this bug. I ran its reproduction steps in the browser: open a photo, close it with X, open another, press Escape. Without the fix, Escape reopens the *first* photo; with it, the viewer closes cleanly. I have added `Fixes #546` to the PR description so it closes on merge.

ui/src/components/photoGallery/presentView/PresentView.tsx
```diff
@@ -38,6 +59,52 @@ const PresentView = ({
...
+    infoOpenRef.current = infoOpen
+  }, [infoOpen])
+
+  useEffect(
+    () => () => {
+      // Leaving the viewer with the info panel open would otherwise leave it
+      // behind over the gallery, still describing the last presented photo. A
+      // sidebar the user opened before presenting is theirs and stays open.
+      if (infoOpenRef.current) {
+        updateSidebar(null)
```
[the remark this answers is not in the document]
kkovaletp: @coderabbitai, the current implementation is good enough. If a user opens the same media in the presentation view, the sidebar will show the same content. If a user opens different media, it means that the info about the previous one is no longer needed and can be replaced. The `unmount` operation on close also works consistently: if the user wants to see the info of the previous media, he/she can easily open it again.

kkovaletp (approved):
LGTM

PR #1490, opened 2026-09-13
Keep the visible part of an album breadcrumb instead of dropping all of it

try-and-succeed:
to all five.

### An album deep in a tree shows no breadcrumb at all

When the path could not be resolved in full — an ancestor the user cannot see — the whole
breadcrumb was dropped rather than the part that is not visible. The user is left without a trail
even for the albums they do have access to. The visible ancestors are kept now, and the rest is
simply absent.

### Ancestors come back closest first, and cycles still end

The truncation above needs the ancestors in order, closest first, and a recursive CTE guarantees no
order. The query now sorts by `LENGTH(path) DESC`: an album's path is its parent's path plus its own
directory name, so an ancestor's path is always a strict prefix of its descendant's, and the closest
ancestor has the longest one. Nothing is added to the recursion itself, so its rows stay identical
from lap to lap, and `UNION` still ends a cyclic parent chain by dropping the repeats. There is no
depth limit.

(An earlier version sorted by a `depth` column, which broke that cycle protection and needed a limit
of 64 instead. Replaced in `c4b96c1` after @googollee's question.)

### Testing

- New tests: truncation at the first inaccessible ancestor, a failing path query reported as an
  error, ancestors closest first, and a two-album cycle that has to return within a timeout (it fails
  when a `depth` column is added back without a bound). The walk is tested on its own
  (`albumAncestors`), because `AlbumPath` then checks ownership through `GetParents`, whose
  `UNION ALL` query still loops on a cycle. That is fixed separately in #1497.
- CI green on sqlite, MySQL and PostgreSQL.

## Summary by CodeRabbit

* **Bug Fixes**
  * Album breadcrumbs now stop at the first inaccessible ancestor while preserving accessible folders closer to the current album.
  * Album path generation now reliably handles deeply nested and cyclic album hierarchies.
  * Errors encountered while loading album paths are now reported instead of appearing as empty paths.
  * Ancestors are listed from closest to farthest for more accurate breadcrumb paths.

* **Tests**
  * Added coverage for restricted access, recursive path failures, cyclic hierarchies, and ancestor ordering.


try-and-succeed:

kkovaletp:
@googollee, please review this PR, as it is addressing an API/DB issue

try-and-succeed:
On the two collapsed findings in https://github.com/photoview/photoview/pull/1490#pullrequestreview-5195926825 — both read, both **declined**, with the reasoning here since they have no thread:

**"`depth < 64` truncates valid chains deeper than 64 levels."** True in principle, not changing it. The bound is what ends a cyclic parent chain now that the `depth` column stops `UNION` from deduplicating repeated rows, so some bound has to exist. Sixty-four levels of nested folders is far beyond any real photo library, and a breadcrumb that stops at the 64th ancestor is a much smaller failure than a query that never terminates. If you would prefer a larger number, it is a one-line change and I am happy to make it.

**"`OwnsAlbum`'s recursive CTE uses `UNION ALL` without a cycle guard."** Also true — but that query is upstream's `GetParentsFromAlbums`, unchanged by this PR and called from many places across the app. Guarding it belongs in its own change rather than in a breadcrumb fix. For what it is worth, the album tree in #1485 no longer goes through it for its batch authorization, and uses the same depth bound instead. Happy to open an issue for `GetParentsFromAlbums` itself if you want it tracked.

api/graphql/models/actions/album_actions.go
try-and-succeed, line 1: > **Updated 2026-09-22:** the context below originally described a `depth` column bounded at 64. That was replaced in `c4b96c1` by ordering on the path length, without any depth limit.

@coderabbitai, I'll write this comment here, but it is relevant for all changes in this PR, not only this file. As a final review step before asking the maintainers to merge, please carefully analyze this PR within the scope of the whole app, including business logic, UX, the testing framework and other relevant aspects.

Context: this PR keeps the visible part of an album's breadcrumb instead of dropping all of it when an ancestor is not accessible, and returns the ancestors closest first by ordering on `LENGTH(path)`, which keeps the recursion free of extra columns so that `UNION` still ends a cyclic parent chain (no depth limit). Pay particular attention to: whether the breadcrumb can now expose the title or id of an ancestor the user has no access to; ordering and correctness of the CTE on sqlite, MySQL and PostgreSQL; and callers of `AlbumPath` that may rely on the old all-or-nothing behaviour, including share-token access.

Please answer:

try-and-succeed: Thanks for the thorough pass. The two access-control conclusions match what I found, and I am leaving the remaining two points as they are — reasoning below rather than silently.

**"The cycle protection is incomplete" — not fixed here, deliberately.** You are right that `AlbumPath` calls `OwnsAlbum` for each candidate ancestor, and that `OwnsAlbum` goes through `GetParentsFromAlbums`, whose recursive CTE has no bound. But that query is upstream's, unchanged by this PR, and it is on the path of most album operations in the app — not only the breadcrumb. Bounding it is a change to shared ownership checking with its own blast radius, and hiding it inside a breadcrumb fix is exactly the kind of bundling the maintainers asked me to stop doing (#1483 was split into five PRs for that reason). What this PR does bound is the query it adds a `depth` column to, which is where the `UNION`'s old cycle protection was lost. Happy to open an issue for `GetParentsFromAlbums`, `GetChildren` and the share-token descendant walk, or a separate PR if that is preferred.

**Test coverage for the depth limit and cycles.** A test for a chain deeper than 64 is easy and I will add one if you would like it. A cycle test is the interesting case: it cannot be written safely *while* `GetParents` is unbounded. Creating a cyclic parent row and then calling `AlbumPath` would hang the test suite in `OwnsAlbum`, not in the query this PR bounded. That test belongs with the fix above, and would be a good way to pin it down.

For the record on what this PR is verified against: `go test ./graphql/models/actions/... -database -filesystem -p 1` on sqlite locally, and the upstream matrix (sqlite, MariaDB, PostgreSQL) has since run green on this PR.

googollee:
I don't understand this fix. The fix added a limitation of foler depth. Why do we need that limitation?

try-and-succeed:
@googollee, good question. The limit was never the goal, only a side effect, and it's gone now (`c4b96c1`).

**Why it was there.** The fix needs the ancestors in a fixed order, closest first: the breadcrumb is cut at the first ancestor the user cannot see, counting outward from the album. A recursive CTE guarantees no order, so I had added a `depth` column to `ORDER BY`. But carrying `depth` through the recursion makes every lap's rows distinct. Before, a cyclic `parent_album_id` chain ended on its own, because `UNION` dropped the repeated rows. With `depth` it would have run forever, so I bounded it at 64. Nothing creates such a cycle today, but the query shouldn't be what turns one bad row into a hang.

**What it does now.** It orders by `LENGTH(path) DESC` instead. An album's path is its parent's path plus its own directory name, so an ancestor's path is always a strict prefix of its descendant's, and the closest ancestor has the longest one. The recursion carries no extra column, so `UNION` ends a cycle by itself again, and there is no depth limit.

Two new tests: the order (closest first), and a two-album cycle that has to return within a 10 s timeout. The cycle test fails with `depth` added back without a bound, so it really pins down the reason. It tests the walk on its own (`albumAncestors`), because `AlbumPath` then checks ownership through `GetParents`, whose `UNION ALL` query still loops on a cycle. That one is fixed separately in draft #1497.

googollee:
@try-and-succeed OK. Now the PR looks reasonable. It fixes the order of Album Path when querying ancestors.

api/graphql/models/actions/album_actions.go
googollee: This comment is out of date. @try-and-succeed please update it.

try-and-succeed: @googollee, thanks. Updated: this comment and the PR description now describe the current version, ordering by path length with no depth limit. The old `depth` approach is only mentioned as history.

try-and-succeed: @coderabbitai, the context of the request above has changed since your answer: `c4b96c1` replaced the `depth` column and its limit of 64 with `ORDER BY LENGTH(path) DESC`, so the recursive CTE carries no extra column and `UNION` ends a cyclic parent chain on its own again. The walk is now `albumAncestors`, with its own tests for order and for a cycle. Please redo the final review for the current head, with the same questions as above. In particular: is the path-length ordering sound for every way albums get their `path` and `parent_album_id` (scanner, symlinks, moved directories), and does anything else rely on the order `AlbumPath` returns?

googollee (approved):

try-and-succeed:
@googollee, thanks for the review and the merge. A follow-up for the one open point from this thread: `OwnsAlbum` still goes through `GetParents`, whose `UNION ALL` query never ends on a parent cycle, and the same holds for `GetChildren`, `Album.Thumbnail` and the share-token check. That is fixed in #1497, which is ready for review now. It is the same `UNION` idea this PR relies on.

PR #1491, opened 2026-09-13
Rescan a single album instead of the whole library

try-and-succeed:
Split out of #1484, which bundled two independent things — after @kkovaletp asked for the same
treatment on #1483, it seemed better to do it up front than after another review round. The other
half (queue status and cancellation) is in the companion PR; #1484 is closed and links to both.

### Moving folders around leaves the old contents behind

Media that has moved out of an album stays in it until someone scans the whole library, which on a
large one is minutes of work for a change that touched one folder.

`scanAlbum(albumId)` walks a single album and its sub-albums and cleans up only what that walk
covers: sub-albums whose directories are gone, and media rows the walk did not see. Nothing outside
the subtree is touched.

Neither cleanup runs when discovery hit an error — a failed walk cannot tell "gone from disk" from
"could not be read", and deleting on that basis is not recoverable.

`DeleteOldUserAlbums` no longer bails out when a scan found no albums at all. Callers only reach it
after a complete discovery, so an empty result means exactly that everything the user had is gone
from disk — which was the one case the cleanup could never handle. The `NOT IN` is left off when
the list is empty, since an empty list is not valid SQL everywhere.

### Permissions

Admins may rescan any album, everyone else the albums they own — the same rule that guards the
existing scanner mutations. The album sidebar gets the button.

### Known limitation, same as the existing scanner

`AddAlbumToQueue` discovers the subtree synchronously on the request thread, exactly as `scanUser`
already does, so a very large album can hold the request open for a while. The real fix is a
"discover children" job type that the workers process themselves — a change to the job model that
belongs in its own PR, and I am happy to open an issue for it.

### Testing

- `go test ./graphql/resolvers/... ./scanner/... -database -filesystem -p 1` against sqlite —
  new tests for the authorization rules, for sub-albums vanishing from disk, for discovery errors
  leaving albums alone, and for the empty-scan cleanup
- `go build ./...`, `go generate ./...` clean

Fork PRs still get no CI (workflows need a maintainer's "Approve and run"), so this was verified
locally against sqlite only.

## Summary by CodeRabbit

- **New Features**
  - Added an album rescan action in the album sidebar.
  - Rescans the selected album and its sub-albums without scanning the entire library.
  - Access is limited to album owners and administrators.
  - The button shows scan progress and can be used again after completion.

- **Bug Fixes**
  - Removes sub-albums that no longer exist on disk.
  - Incomplete scans no longer remove existing album records.
  - Empty scan results correctly clean up stale albums.

- **Localization**
  - Added album rescan labels for supported languages.


try-and-succeed:
@coderabbitai review

api/scanner/scanner_user.go
```diff
@@ -89,6 +90,68 @@ func FindAlbumsForUser(db *gorm.DB, user *models.User, albumCache *scanner_cache
...
+	}
+
+	// DeleteOldUserAlbums treats every album missing from userAlbums as
+	// stale, so a partial walk would take valid albums down with it.
+	if discoveryComplete {
+		deleteErrors := cleanup_tasks.DeleteOldUserAlbums(db, userAlbums, user)
+		scanErrors = append(scanErrors, deleteErrors...)
+	} else {
+		scanErrors = append(scanErrors, errors.Errorf("Skipped cleanup of removed albums for user '%s': album discovery was incomplete\n", user.Username))
+	}
```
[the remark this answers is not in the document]
try-and-succeed: Fixed in `fdfdd13`. Confirmed, and this one deletes data: `directoryContainsPhotos` answered `false` for a directory it could not open — the same answer as for one without media — and reported the failure through `ScannerError`, where the walk never saw it. The sub-directory was then simply not queued, so discovery looked complete and the cleanup removed the album row for a directory still sitting on disk with photos in it.

It returns the error now and the walk appends it to `scanErrors`. Both entry points already skip their cleanup once discovery reported an error, so the album survives until a scan that can actually read it.

New test: a sub-album's directory is made unreadable, and it asserts both that an error is reported and that the row survives. It fails without the change, and skips where the process can read a `0o000` directory anyway, as root can.

api/scanner/scanner_user.go
try-and-succeed, line 1: @coderabbitai, I'll write this comment here, but it is relevant for all changes in this PR, not only this file. As a final review step before asking the maintainers to merge, please carefully analyze this PR within the scope of the whole app, including business logic, UX, the testing framework and other relevant aspects.

Context: this PR adds `scanAlbum` to rescan a single album and its sub-albums, with cleanup scoped to that subtree, skipped whenever discovery hit an error (including a sub-directory that could not be read), and makes `DeleteOldUserAlbums` handle a scan that found nothing. Pay particular attention to: any path by which a scan could delete albums or media that still exist on disk (unreadable or unmounted directories, ignore files, symlinks, concurrent scans); authorization of `scanAlbum` for non-admin owners; and the known limitation that discovery runs synchronously on the request.

Please answer:

try-and-succeed: Thanks — this one deserves a detailed answer, because I disagree with the severity of the first finding and want to show why rather than just declining it.

**"An ignored descendant can be deleted during `scanAlbum`" — this is upstream's existing behaviour, not something this PR introduces.** The full user scan does exactly the same thing today, without this PR:

- `walkAlbumScanQueue` skips a directory matched by an ignore rule and adds neither an album nor an error (upstream `api/scanner/scanner_user.go`, lines 110-114 — the `ignorePaths.MatchesPath(...)` / `continue` block);
- `FindAlbumsForUser` then calls `cleanup_tasks.DeleteOldUserAlbums` unconditionally (same file, line 222), which deletes every album row missing from the walk — including the ignored one.

So adding an ignore rule and rescanning has always removed the album from the library, which is a defensible reading of what "ignore" means: the directory is excluded from the library, and its rows go with it. `scanAlbum` applies the same rule to one subtree instead of all of them, which is the consistent behaviour. If the project wants ignored-but-previously-imported albums retained instead, that is a change in what `.photoviewignore` means and should apply to both scan paths at once — I would be glad to do it as its own PR, with the behaviour decided first.

**The other three are marked "predates this PR" in your own analysis, and I agree with that classification:**

- *Media cleanup after a per-file scan failure* — `ScanAlbum` logs and continues past a failed `scanMedia`, and `CleanupMedia` then deletes rows for media it did not see. Real, and worth fixing, but it is upstream's scan loop; this PR only makes it reachable on demand.
- *Discovery and cleanup are not serialized* — the queue de-duplicates media scan jobs, not discovery walks, and that is upstream's queue design. Worth a dedicated change (a scan generation or a per-root lock), not a side effect of this one.
- *The mutation blocks the request for the walk* — named as a known limitation in the PR description, with the same shape as the existing `scanUser`. The real fix is a "discover children" job type that the workers process, which changes the job model.

Happy to open issues for any of these so they are tracked rather than lost.

**On the smaller points:** the rescan button re-enabling as soon as discovery returns is fair — it reflects "the request finished", not "the scan finished", and the queue view in #1492 is where progress actually shows. I can make the button reflect the queue state once #1492 lands, since it is that PR's query that knows.

What this PR is verified against: `go build`, `go generate` clean, and `go test ./scanner/... ./graphql/resolvers/... -database -filesystem -p 1` on sqlite locally; the new tests cover the authorization rules, vanished sub-albums, a sub-directory that cannot be read, and the empty-scan cleanup.

kkovaletp:
@googollee, please review this PR, as it is related to the scanner workflow

try-and-succeed:
CI ran on this PR for the first time, and Codecov's patch check failed at 46.7%. I've closed the gaps in `6e41b75` (plus a typing fix in the next commit):

- **`SidebarAlbumScan.tsx`: 30% → 100%.** Tests for queueing a scan, the button becoming usable again afterwards (that needs the mutation's `reset()`: the test fails without it), and after a failed request.
- **`AddAlbumToQueue`: 0% → 100%.** The resolver tests replace it with a stub, so the real function never ran. It is now tested against a real directory tree: the album and its sub-album are queued, an album outside the subtree is not, and a directory that is gone is reported with nothing queued. To use `test_utils` there, the queue package gets a `TestMain` like the other scanner packages, and its own copies of the `-database`/`-filesystem` flags go (they would clash with `test_utils/flags`).
- **`ScanAlbum`:** an album that does not exist, and a queue that refuses the album, now have tests. Both have to come back as errors, not as "Scanner started".

Measured locally on the lines this PR adds. Still uncovered, on purpose:

- `resolvers/scanner.go` 72–73 and `cleanup_media.go` 108–109: pass-through returns for a database error in the middle of the operation (`OwnsAlbum`, `GetChildren`).
- `scanner_user.go` 80 and 121: `os.Stat` failing with something other than "does not exist" (a permission error). That can't be produced when the tests run as root, which they do in the CI container.
- `scanner_user.go` 241–245, 314, 373: upstream code that only moved into the new shared walk function, or changed its return signature. The logic is unchanged.

try-and-succeed:
Follow-up to the cleanup finding from this PR's final review, which I left out here because the bug predates it: #1496 is ready for review now. A photo whose database lookup fails during a scan is no longer deleted by the cleanup that follows.

PR #1492, opened 2026-09-13
Show what the scanner is doing, and let it be stopped

try-and-succeed:
Split out of #1484, which bundled two independent things — after @kkovaletp asked for the same
treatment on #1483, it seemed better to do it up front than after another review round. The other
half (rescanning a single album) is in the companion PR; #1484 is closed and links to both.

### A running scan is invisible and unstoppable

There is no way to tell whether the scanner is doing anything, on which album, or how much is still
queued — and no way to stop a scan started by accident on a large library.

- `scannerQueueStatus` returns what is running and what is queued
- `cancelScanJob(albumId)` stops one album's job
- `cancelAllScanJobs` stops the lot and returns how many it cancelled

Admins see and cancel everything; everyone else only the albums they own. The settings page gets
the list with the two buttons.

### Cancellation is cooperative, and the UI says so

A running job stops after the file it is on. Already-processed files are kept rather than rolled
back, and no half-written cache file is left behind. That is also why a job carries an id now: it
needs a `context.CancelFunc`, and `ScannerJob` was compared by value.

A cancelled job stays on the queue while it winds down, and a restart of the same album may run
beside it — so cancelled jobs are hidden from the status. Listing both would show one album twice
and leave it ambiguous which row a cancel refers to, and the dead one is on its way out regardless.
That is also what makes cancelling by album id unambiguous: at most one *live* job per album can
exist, which `TestScannerQueueOneLiveJobPerAlbum` pins down from both sides.

The after-scan cleanup is skipped when the context is done. A cancel landing after the last file
would otherwise run it against a scan that had not necessarily seen everything, taking media still
on disk down with it.

### Testing

- `go test ./graphql/resolvers/... ./scanner/scanner_queue/... -database -filesystem -p 1` against
  sqlite — new tests for the queue status, both cancel paths, the permission filtering, and the
  one-live-job-per-album invariant
- `go build ./...`, `go generate ./...` clean
- `npm test` for the queue list, `npm run build`, `eslint` and `prettier --check`

Fork PRs still get no CI (workflows need a maintainer's "Approve and run"), so this was verified
locally against sqlite only.

## Summary by CodeRabbit

- **New Features**
  - Added scanner queue visibility in Settings, showing running and queued scans.
  - Added controls to cancel individual scans or clear all eligible scans.
  - Queue access is available to all users, while visibility and cancellation are limited to owned albums for non-admins.
  - Added localized labels and status indicators for the queue.

- **Bug Fixes**
  - Improved scan cancellation so the active file completes safely before the job stops.
  - Prevented overlapping scans for the same album during cancellation and restart.


try-and-succeed:
@coderabbitai review

api/scanner/scanner_queue/queue.go
```diff
@@ -255,10 +279,125 @@ func (queue *ScannerQueue) jobOnQueue(job *ScannerJob) (bool, error) {
 	scannerJobs := append(queue.in_progress, queue.up_next...)
 
 	for _, scannerJob := range scannerJobs {
+		// A cancelled in_progress job stays in the slice until it finishes its
+		// current file and exits on its own (see CancelJob) - ignore it here
+		// so a restart isn't silently dropped while that exit is still
+		// pending, which could take as long as one more file to process.
+		if scannerJob.ctx.Err() != nil {
+			continue
+		}
```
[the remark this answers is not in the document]
try-and-succeed: Fixed in `86d29fb`. You are right that the two rules did not meet: `jobOnQueue` deliberately lets a restart onto the queue while the cancelled job winds down (dropping it silently would be worse), but `processQueue` then took the head of the queue regardless and could start it beside the job still finishing its file — two scans on one album, each with its own transaction.

`processQueue` now picks the first queued job whose album is not already in `in_progress`, cancelled or not, and leaves the rest queued. The wait is bounded by the one file the old job still has to finish, and `removeFinishedJob` already notifies the queue, so the restart starts as soon as the album is free. A job for a different album is unaffected.

The test covers all three states: a different album runs immediately, the restart stays queued while the cancelled job is there, and it becomes runnable the moment `removeFinishedJob` takes it out.

try-and-succeed:
On the collapsed finding in https://github.com/photoview/photoview/pull/1492#pullrequestreview-5191767096 (make after-scan cleanup observe cancellation) — read and **declined**, reasoning here since it has no thread.

The finding itself notes the important part: `findMediaForAlbum` completes before the scan loop, so a cleanup that starts always works from a complete list of the album's media and can never delete files that are still on disk. What remains is a narrow race — a cancel arriving after the check that skips cleanup for an already-cancelled job, but before the cleanup finishes — in which a correct cleanup runs to completion instead of stopping part-way. That matches the cancellation contract documented on the schema: a running job stops once it finishes its current step, and nothing already processed is rolled back. Interrupting the cleanup mid-way would leave the album half cleaned for the next scan to finish, which is not better, just less finished.

api/scanner/scanner_queue/queue.go
try-and-succeed, line 1: @coderabbitai, I'll write this comment here, but it is relevant for all changes in this PR, not only this file. As a final review step before asking the maintainers to merge, please carefully analyze this PR within the scope of the whole app, including business logic, UX, the testing framework and other relevant aspects.

Context: this PR adds `scannerQueueStatus`, `cancelScanJob` and `cancelAllScanJobs`. Cancellation is cooperative — a running job stops after its current file — cancelled jobs are hidden from the status while they wind down, at most one live job exists per album, and a restart of a cancelled album is held in the queue until the cancelled job has exited. Pay particular attention to: races between cancel, restart, job completion and the periodic scanner; whether the after-scan cleanup can run against an incomplete scan; locking and goroutine lifecycle in the queue; and what non-admin users can see or cancel.

Please answer:

try-and-succeed: Thanks — both issues are fixed in `0e1b797`.

**Cancellation now really lets the current file finish.** `TaskContext.WithoutCancel` returns a context with the same album, cache and database that the cancel does not reach — re-binding the database handle, which otherwise carried the job's cancellation into every query. `ScanAlbum` and the discovery loop still check the job's context between files and run each file under the detached one. Tests: the task pipeline, cancelled from inside a file's processing, still runs that file's remaining steps under the detached context, and stops between them under the job's own; a second test checks that the detached database handle still queries after the job is cancelled, and fails without the re-binding. `TestFullScan` passes apart from the subtests that need face detection or a JPEG XL delegate in this environment.

**The queue is shown to every user.** It renders outside the admin-only scanner section, still after the scanner controls for an admin and still nothing while a user has no jobs; library-wide scanning stays admin-only. Tests cover both roles, and moving it back into the admin section fails the non-admin one.

kkovaletp:
@googollee, please review another scanner-related PR

PR #1493, opened 2026-09-14
feat(api): configure the periodic scan interval from the environment

ndreno:
### Problem

The periodic scan interval only lives in the database, set through the web interface. A deployment cannot describe its own scan schedule, so the value is invisible to configuration management and any change made in the UI silently drifts away from what the deployment declares.

This matters for anything that adds folders to the library from outside Photoview: the new media only appears after someone remembers to press *Scan All*, or after guessing what the stored interval currently is.

### Solution

`PHOTOVIEW_PERIODIC_SCAN_INTERVAL` sets the interval in seconds, `0` disabling periodic scans. When set, it takes precedence over the stored value, consistent with the other environment variables the server reads. An unset, empty, or invalid value falls back to the database and logs a warning, mirroring `PHOTOVIEW_MEDIA_PROBE_TIMEOUT`.

The override is logged at startup, the same way `PHOTOVIEW_DISABLE_FACE_RECOGNITION` announces itself.

### Tests

Six cases on parsing the variable, in a new `api/utils/environment_variables_test.go`, and three added to the existing `TestGetPeriodicScanInterval`: precedence over the database, `0` disabling scans, and an invalid value falling back. `go vet` and `gofmt` are clean.

### Note

The interval field in the web interface still accepts input while the variable is set; it simply no longer takes effect after a restart. Making it read-only would need a frontend change, which I left out to keep this focused. Happy to follow up if you would like it.

## Summary by CodeRabbit

* **New Features**
  * Added an environment variable to configure the periodic media-library rescan interval in seconds.
  * Environment configuration takes precedence over the interval set in the web interface.
  * Set the interval to `0` to disable periodic scanning.
  * Invalid, negative, or excessively large values fall back to the web interface setting.

* **Documentation**
  * Added configuration guidance and an example value to the sample environment file.


ndreno:
Both findings were real, thanks. Fixed in the latest commit.

**Overflow.** `9223372037` seconds parsed fine and then wrapped to a negative `time.Duration`, which the scanner reads as "periodic scanning disabled" — silently, which is the worst shape for this bug. Parsing now uses `strconv.ParseInt` and rejects anything above the largest representable number of seconds, falling back to the database with the same warning as any other invalid value. Three cases cover it: the boundary value, boundary plus one, and a value beyond `int64` entirely.

**Test isolation.** Correct, an inherited `PHOTOVIEW_PERIODIC_SCAN_INTERVAL` would have turned the fallback cases into override cases. Every case now sets the variable explicitly through `t.Setenv`, and the cases whose contract is that it is absent unset it afterwards, so the restore still happens. `TestGetPeriodicScanInterval` sets it empty at the parent scope for the same reason.

One thing I did not touch: `MediaProbeTimeout` has the same overflow, since I mirrored it. It is less exposed, its value is a timeout rather than a scheduling interval, and fixing it here would widen a focused change. Happy to send it separately if you would like.

kkovaletp:
@ndreno, thanks for the new feature proposal.
The idea of having the default value for the scan interval configurable through an env var looks good to me.
But having the actual value for that interval set in a priority through an env var breaks the UX of the Admin settings, and I don't want that. If you need to control the actual scan interval value on a running instance from the shell environment, I'd recommend you write a simple script that would interact with the Photoview DB and read/modify the value for you.

The problem described in the description about the inability to trigger a full scan programmatically from outside of the GUI as part of the source media library content update is valid. But I'd solve it not by manipulating the scan interval value, but by implementing an easily callable API + a shell script (or something like that - this idea needs to be researched).

@googollee, what do you think about this case?

googollee:
I'm concerning that what the behavior is when scan interval in database is not same as this periodic ENV. Are they both triggering the scan?

For my usage, I'm fine there is a delay between photos are copied in the libraryn and showing in the UI.

If we want to reduce that delay, the better solution is [fsnotify](https://pkg.go.dev/github.com/fsnotify/fsnotify), but not a rapid periodical scan. It only works with local fs, but not remote fs.

PR #1495, opened 2026-09-15
fix: scope AddAlbumShare ownership check to the requested album

### LINKED ISSUE #1494: Missing ownership check in shareAlbum lets any user share another user's album
## Summary

Photoview lets a user generate a share link for one of their albums via the `shareAlbum` GraphQL mutation. The resolver's authorization check, `actions.AddAlbumShare`, verifies only that the calling user owns some album in the system, it never checks that the user owns the specific album id being shared. Any authenticated user who owns at least one album can call `shareAlbum` with an arbitrary `albumID` belonging to another user and receive a working share token for it, exposing that album's photos (and, if left unprotected, without a password) to anyone holding the link.

Affected versions: confirmed on commit `81affea` (current `main`).

## Details

`graphql/resolvers/share_token.go`, the `ShareAlbum` resolver:

```go
func (r *mutationResolver) ShareAlbum(ctx context.Context, albumID int, expire *time.Time, password *string, label *string) (*models.ShareToken, error) {
	user := auth.UserFromContext(ctx)
	if user == nil {
		return nil, auth.ErrUnauthorized
	}

	return actions.AddAlbumShare(r.DB(ctx), user, albumID, expire, password, label)
}
```

`graphql/models/actions/share_token_actions.go`, `AddAlbumShare`'s ownership check as shipped:

```go
var count int64
err := db.
	Model(&models.Album{}).
	Where("EXISTS (SELECT * FROM user_albums WHERE user_albums.album_id = albums.id AND user_albums.user_id = ?)",
		user.ID).
	Count(&count).Error

if err != nil {
	return nil, errors.Wrap(err, "failed to validate album owner with database")
}

if count == 0 {
	return nil, auth.ErrUnauthorized
}
```

The `Where` clause never references `albumID`, the mutation's own argument. It counts every album row for which a `user_albums` entry exists for the calling user, i.e. it answers "does this user own at least one album anywhere", not "does this user own this album". As long as that count is non-zero, execution falls through to create a share token for the caller-supplied `albumID`, regardless of whose album it is.

The sibling `AddMediaShare`, in the same file, gets this right by scoping the ownership join to the specific requested id:

```go
func AddMediaShare(db *gorm.DB, user *models.User, mediaID int, ...) (*models.ShareToken, error) {
	var media models.Media
	...
	err := db.Joins("Album").
		Where(query, user.ID).
		First(&media, mediaID).
		Error

	if err != nil {
		if errors.Is(err, gorm.ErrRecordNotFound) {
			return nil, auth.ErrUnauthorized
		}
		...
	}
	...
}
```

`First(&media, mediaID)` filters by the target media's primary key while the join requires that its album belong to the caller, so a media id the caller doesn't own correctly 404s into `ErrUnauthorized`. `AddAlbumShare` has no equivalent filter on `albumID`.

## POC

(available upon request)

## Impact

Any authenticated, non-admin user who owns at least one album can call the `shareAlbum` mutation with any other album's numeric id and receive a valid share token for it:

- Read access to another user's private album (all of its photos, and recursively its own sub-albums, since child albums inherit the parent share) via the resulting share link, optionally with no password if none is supplied.
- Album ids are small sequential integers, not secrets, so enumeration is practical.
- The share token is a persisted row owned by the attacker (`OwnerID: user.ID`), so the attacker can also later modify its password, expiry, or label, extending their control over exposure of another user's album indefinitely.

carfeii:
Fixes #1494.

`AddAlbumShare`'s ownership query checked only that the caller owns some album, not the specific `albumID` argument being shared, so any user with at least one album could create a share token for any other album. This adds the missing `albums.id` filter, matching the pattern `AddMediaShare` already uses for its own ownership check, plus a regression test.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected album sharing authorization to verify ownership of the specific album being shared.
  - Prevented users from sharing albums they do not own, even when they own other albums.


api/graphql/models/actions/share_token_actions.go
```diff
@@ -67,6 +67,7 @@ func AddAlbumShare(db *gorm.DB, user *models.User, albumID int, expire *time.Tim
 	var count int64
 	err := db.
 		Model(&models.Album{}).
+		Where("albums.id = ?", albumID).
```
kkovaletp, line 70: @coderabbitai, this PR contains a very focused, specific change, but it has significant security impact. Please carefully analyze this PR and pay additional attention to the security aspects. Try to find and analyze all possible corner cases and non-standard usage scenarios.
Do you see any regressions, security or performance degradations, and any other types of bugs? Anything else you'd like to highlight before the merge?

kkovaletp: @carfeii, thanks for the security fix contribution!
Would you like to contribute the fixes for the 2 bot's findings above as well (in separate PRs)?

@googollee, please review this security fix.

googollee (approved):

api/graphql/models/actions/share_token_actions.go
googollee: Approved.

googollee: We can create issues with the 2 issues that AI found.

kkovaletp: @coderabbitai, please create new issues for the following previously found concerns in this thread:

- albumResolver.Shares in api/graphql/resolvers/album.go returns all tokens for an album without checking the authenticated token owner. A caller with a valid album share token can access the shared album and query shares { token }. This can disclose sibling token values. It can also weaken token revocation, because the caller can retain a different disclosed token for the same album.
- queryResolver.Media in api/graphql/resolvers/media.go dereferences shareToken.MediaID without checking for nil. Supplying album-token credentials to media can cause a nil-pointer panic. The GraphQL recovery behavior determines whether this is request-local or process-wide.

For each issue, please add a detailed description and an implementation plan.

PR #1496, opened 2026-09-17
Keep a photo whose scan failed instead of deleting it in the cleanup

try-and-succeed:
Opened as a draft so it is tracked. There are already several PRs of mine waiting for review, so I
will mark this ready once some of those have moved.

CodeRabbit raised this in its final review of #1491, and I left it out of that PR because the bug
already exists on master. This PR fixes it on its own.

### A failed lookup deletes a photo that is still on disk

`findMediaForAlbum` runs one database transaction per file. When that transaction fails, the error
is logged and the file is skipped. It is then missing from the list of media the scan found. After
the scan, `CleanupMedia` deletes every media row that is not in that list. So a transient database
error during the lookup of an **existing** photo deletes its row, along with its favourites, face
assignments and cache, although the file has not gone anywhere. The next successful scan adds the
photo back as a new media row, but the user data attached to the old row is gone.

For an existing file, the only step that can fail in that transaction is the database lookup
itself. The later hooks only do work for new media.

(The review pointed at the `scanMedia` loop in `ScanAlbum`. A failure there does not remove
anything from the list. The deletion happens through the discovery step described above.)

### The fix

`findMediaForAlbum` now also returns the paths of the files that failed. `ScanAlbum` stores them in
the task context (`WithUnscannedMediaPaths`), and `CleanupMedia` keeps the rows with those path
hashes.

- A file that really was removed is still cleaned up, in the same scan.
- A new file that keeps failing has no row yet, so it does not block the cleanup for the rest of
  the album. That is why this keeps individual rows instead of skipping the whole cleanup.

`AfterScanAlbum` keeps its signature, so the other scanner tasks are unchanged.

### Testing

- New `TestCleanupMediaKeepsMediaThatFailedToScan`. It scans an album with two photos, then scans
  again with one file removed from disk and the database lookup of the other one failing. A GORM
  query callback injects that failure, so there is no test seam in production code. The removed
  photo's row is deleted and the failed one's row is kept. **With the fix disabled, the test fails**
  and both rows are gone.
- `go build ./...`, and `go test ./... -database -filesystem -p 1` on sqlite locally. The only
  failures are the ones that also fail on unmodified master in my environment: face detection
  (built with `no_face_detection`) and the JPEG XL blurhash in `TestFullScan`.

### Overlap with #1492

The two merge without a textual conflict, but they do not compile together. #1492 adds a
cancellation check to `findMediaForAlbum` that uses `return nil, err`, and after this PR that line
has to be `return nil, nil, err`. Whichever PR lands second needs this one-line change. I will do
it on whichever PR that is.


try-and-succeed:
Marked ready now that #1490 has landed. Merged locally onto current `master` (`4d25eb6`): builds, and all tests pass apart from the known local face-detection and JPEG XL failures.

**Codecov (82% patch):** the 3 missed lines are existing error returns in `findMediaForAlbum` (`os.ReadDir`, `item.Info()`, `MediaFound`) that only gained an extra `nil` for the new return value. They were uncovered on `master` before. `item.Info()` only fails in a race with a file being deleted, and no task's `MediaFound` returns an error today, so I left them uncovered rather than add a test seam for them.

try-and-succeed:
@coderabbitai full review

api/scanner/scanner_tasks/cleanup_tasks/cleanup_media.go
try-and-succeed, line 1: @coderabbitai, I'll write this comment here, but it is relevant for all changes in this PR, not only this file. As a final review step before asking the maintainers to merge, please carefully analyze this PR within the scope of the whole app, including business logic, UX, the testing framework and other relevant aspects.

Context: a per-file transaction failing in `findMediaForAlbum` used to leave that file out of the scanned list, so the cleanup afterwards deleted its media row along with favourites, face assignments and cache, although the file was still on disk. The failed paths now travel through the task context (`WithUnscannedMediaPaths`) and the cleanup keeps their rows.

Please pay particular attention to: whether a file that really was deleted can now survive the cleanup; whether the new return value or the context key changes behaviour for the other scanner tasks; and the interaction with a cancelled scan.

Please answer:

try-and-succeed: Thanks — that matches my own reading, including the trade-off you name.

**The deletion race: intended, and the cheaper of the two mistakes.** A file removed *after* `os.ReadDir` saw it, whose scan then fails, keeps its row for one cycle. As you say, that case cannot be told apart from the one this PR is about. Of the two possible mistakes — keeping a row one scan too long, or deleting favourites, face assignments and cache for a photo that never left the disk — only the first one is reversible, and the next successful scan does reverse it. I made that choice knowingly and the PR description says so.

**The cancellation test: not adding one here.** The skip it would cover is upstream's, not this PR's: `AfterScanAlbum` runs through `simpleCombinedTasks`, which checks `ctx.Done()` before each task (`scanner_tasks.go:36-42`), and `MediaCleanupTask` is the last registered task. This PR changes neither that loop nor the task order, and a test for it would pass with the PR reverted, so it belongs with the cancellation work in #1492 rather than here. Happy to add it there or in a follow-up if a maintainer wants it.

PR #1497, opened 2026-09-17
Stop recursive album queries from running forever on a parent cycle

try-and-succeed:
Opened as a draft so it is tracked. There are already several PRs of mine waiting for review, so I
will mark this ready once some of those have moved.

CodeRabbit raised this in its final review of #1490. I left it out of that PR because it changes
queries shared across the whole app, not just the breadcrumb. This PR handles it on its own.

### Recursive album queries never finish on a parent cycle

Four queries walk the album tree with a recursive CTE joined by `UNION ALL`:

- `GetChildrenFromAlbums` (and `GetChildren`)
- `GetParentsFromAlbums` (and `GetParents`)
- `Album.Thumbnail`, for albums without a cover
- the album share-token check in `routes/authenticate_routes.go`

If `parent_album_id` links ever form a cycle, `UNION ALL` keeps producing the same rows and the
query does not end. `OwnsAlbum` goes through `GetParents`, so this affects most album operations,
not just one page. The share-token check hangs as well, because its `COUNT` has to finish the whole
walk.

The scanner builds parents from the directory tree and does not create cycles. A hand-edited,
restored or otherwise corrupted database can still contain one.

### The fix

`UNION` instead of `UNION ALL`. It drops rows that were already found, so the recursion ends once
every album in the cycle has been visited. `AlbumPath` already relies on this on master. On a valid
tree every album is reached exactly once, so the results do not change. The one exception is
`GetChildrenFromAlbums` called with overlapping roots, which now returns each album once instead of
twice. No caller on master passes overlapping roots.

Each query has a short comment so that nobody switches it back to `UNION ALL` as an optimisation.
`AlbumPath` is not touched here, because #1490 changes it.

**Cost**, measured on sqlite on a Raspberry Pi with a 2,071-album tree, averaged over 50 runs:

| Query | `UNION ALL` | `UNION` |
|---|---|---|
| all children of the root (2,071 rows) | 61.4 ms | 65.9 ms |
| all parents of the deepest album (33 rows) | 1.08 ms | 1.31 ms |

### Testing

- New `TestAlbumQueriesFinishOnCyclicParents` builds two albums that are each other's parent and
  calls `GetParents`, `GetChildren`, `GetChildrenFromAlbums`, `OwnsAlbum` and `Thumbnail`.
- New `TestAuthenticateAlbumShareTokenFinishesOnCyclicParents` checks that an album inside the
  cycle is accepted and an unrelated album is refused with 403.
- Each call runs under a 10 s context timeout, so a regression fails the test instead of hanging the
  suite. **On unmodified master, all seven calls hit that timeout.** With the fix, they finish in
  well under a millisecond.
- `go build ./...`, and `go test ./... -database -filesystem -p 1` on sqlite locally. The only
  failures are the ones that also fail on unmodified master in my environment: face detection
  (built with `no_face_detection`) and the JPEG XL blurhash in `TestFullScan`.

I could only run this against sqlite locally. `UNION` in a recursive CTE is supported by MySQL 8,
MariaDB and PostgreSQL, but the CI matrix is what actually verifies that.


try-and-succeed:
Marked ready now that #1490 has landed. Merged locally onto current `master` (`4d25eb6`): builds, and all tests pass apart from the known local face-detection and JPEG XL failures. CI was 20/20 on the draft (MySQL and PostgreSQL included), and Codecov patch coverage is 100%.

try-and-succeed:
@coderabbitai full review

api/graphql/models/album.go
try-and-succeed, line 1: @coderabbitai, I'll write this comment here, but it is relevant for all changes in this PR, not only this file. As a final review step before asking the maintainers to merge, please carefully analyze this PR within the scope of the whole app, including business logic, UX, the testing framework and other relevant aspects.

Context: four recursive album queries used `UNION ALL`, so a cycle in `parent_album_id` made them run forever — `GetChildrenFromAlbums`, `GetParentsFromAlbums`, `Album.Thumbnail` and the album share-token check. They use `UNION` now, which ends the walk. #1490 has since landed and made `AlbumPath` cycle-safe the same way.

Please pay particular attention to: whether any caller depends on duplicate rows from `UNION ALL` (overlapping roots in `GetChildrenFromAlbums`); the share-token check's `COUNT` semantics with deduplicated rows; and performance on large trees across sqlite, MySQL and PostgreSQL.

Please answer:

try-and-succeed: Thanks. Two notes on your points:

**The CI matrix you ask for is green**, on sqlite, MariaDB/MySQL and PostgreSQL, and Codecov reports 100% patch coverage. I have also merged the branch onto current `master` (`4d25eb6`, which contains #1490) locally and run the API suite there: no failures beyond the two that fail on unmodified `master` in my environment (face detection with `no_face_detection`, and the JPEG XL blurhash in `TestFullScan`).

**The de-duplication contract for overlapping roots:** agreed that a plain non-cyclic test would state it more directly than the cycle test does. Of the in-repo callers, none depends on multiplicity, as you found. I would rather not add a test whose only purpose is to pin down a contract for a hypothetical external consumer of an exported method — but if a maintainer would like `GetChildrenFromAlbums` to document "each album once, regardless of overlapping roots" with a test of its own, it is a few lines and I will add it.

PR #1502, opened 2026-09-21
fix: reject album share tokens in the media query instead of panicking

### LINKED ISSUE #1501: Reject album share tokens in media queries without a panic
## Summary

`queryResolver.Media` dereferences `shareToken.MediaID` without first checking whether it is nil.

Album share tokens have no media ID. If a caller supplies album-token credentials to the `media` query, the resolver can dereference a nil pointer.

## Security and reliability impact

The failure can cause a panic. The GraphQL recovery configuration determines whether the effect is limited to one request or affects the process. The resolver must reject credentials that do not apply to the requested resource type before it dereferences token-specific fields.

## Affected area

- `api/graphql/resolvers/media.go`
- `queryResolver.Media`
- Share-token credential validation for media queries

## Required changes

1. Validate the share-token type before accessing `shareToken.MediaID`.
2. Reject album-token credentials for `media` with a controlled authorization or validation error.
3. Preserve valid media-token access behavior.
4. Add regression tests for the token-type mismatch.

## Implementation plan

1. Trace how `queryResolver.Media` parses and validates share-token credentials.
2. Add a nil check or explicit token-type check before dereferencing `shareToken.MediaID`.
3. Return the established GraphQL authorization or validation error for an album token supplied to `media`.
4. Confirm that valid media-token credentials still grant access only to the token's media item.
5. Add resolver tests for album-token credentials, media-token credentials, missing credentials, and mismatched media IDs.
6. Confirm that the mismatch path returns an error and does not panic.

## Acceptance criteria

- Supplying album-token credentials to `queryResolver.Media` returns a controlled error.
- The resolver does not dereference a nil `shareToken.MediaID`.
- Valid media-token credentials continue to work for their linked media item.
- Automated tests cover the album-token and media-query mismatch.
- The test suite verifies that this path does not panic.

## Backlinks

- Reported during security review of https://github.com/photoview/photoview/pull/1495.
- Original review comment: https://github.com/photoview/photoview/pull/1495#discussion_r4035242542.
- Requested by @kkovaletp.

Retsumdk:
## Summary

`queryResolver.Media` read `*shareToken.MediaID` without checking that the supplied credential is a media token at all. Album share tokens have no media ID, so passing one to the `media` query dereferenced a nil pointer and panicked — a request-level crash whose blast radius depends on how GraphQL recovery is configured in a given deployment.

The resolver now rejects a credential that does not apply to the requested resource type before reading any token-specific field, returning the same `auth.ErrUnauthorized` this file already returns for access it does not grant.

Closes #1501.

## Behaviour

| Credential passed to `media` | Before | After |
| --- | --- | --- |
| album share token | panic (nil dereference) | `unauthorized` |
| media share token, matching ID | the media | the media (unchanged) |
| media share token, different ID | falls through to session check | unchanged |
| unknown token value | `share not found` | unchanged |
| no token, no session | `unauthorized` | unchanged |
| no token, owner session | the media | unchanged |

One deliberate choice worth flagging: an album token is rejected even when the caller also carries a valid session, rather than being ignored in favour of the session. Rejecting a credential that does not match the resource type is the behaviour #1501 asks for, and it keeps the answer independent of which credential the client meant. The alternative — mirroring the guard style in `queryResolver.Album` (`if shareToken.Album != nil`) and falling through to the session check — is a two-line change if you prefer it; say the word and I will switch it.

## How it was tested

New `api/graphql/resolvers/media_test.go` (9 cases): album token against an unrelated media item, album token against a media item in the token's own album, matching media token, mismatched media token, unknown token, token whose media row has been deleted, no credentials, owner session, and album token alongside an owner session.

The album-token case fails with `panic: runtime error: invalid memory address or nil pointer dereference` on `master` and passes with the change, so the regression is pinned:

```
--- FAIL: TestMediaQueryRejectsAlbumShareToken (0.59s)
panic: runtime error: invalid memory address or nil pointer dereference
```

Run locally against SQLite, the way CI runs the matrix:

```
go test ./graphql/... -database -count=1 -p 1     # all pass
```

Full `go test ./... -database -filesystem -count=1 -p 1` passes except `api/scanner`'s `TestFullScan/CheckMedia` and `TestFullScan/CheckMediaURL`, which fail identically at `master` in my environment (local ImageMagick 7.1.2-31 and ffmpeg instead of the image's pinned builds), so they are unaffected by this change.

## Notes

- Changes are additive: 7 lines in `media.go` (no deletions), plus the new test file.
- Only #1501 is addressed; `#1500` (sibling token disclosure in `Album.shares`) is a separate resolver and is untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed media access through share links so album share links are rejected when accessing individual media.
  * Prevented errors when using invalid, missing, or outdated share credentials.
  * Ensured deleted media links return a clear “share not found” error instead of causing a failure.
  * Preserved valid access for matching media share links and authorized owner sessions.


Retsumdk:
Amended: this branch now uses the `shareToken.Media != nil` guard (same shape as #1509 and as `queryResolver.Album`) and keeps the nine-case test file. `go test ./graphql/... -database -p 1` green. See #1501 for the comparison.

kkovaletp:
@googollee, we have 2 PRs fixing the same issue #1501 
Please check the comparison of these PRs in https://github.com/photoview/photoview/issues/1501#issuecomment-5798419748
Review both PRs (#1502 and #1509), choose 1 and close the opposite one.

googollee:
Since this one is earlier than #1509 and the test covers more cases, I'd like to process this one.

googollee (approved):

